PR #25, opened 2014-03-11
Add RFB protocol version negotiation

easybe:
Before, if the client did not support the by the server
    advertised RFB protocol version, the program would hang. The
    correct behavior is to select the next lower supported version.
    Like before, we really only support version 3.3.

### DIFF
--- a/vncdotool/rfb.py
+++ b/vncdotool/rfb.py
@@ -125,15 +125,22 @@ class RFBClient(Protocol):
     def _handleInitial(self):
         buffer = ''.join(self._packet)
         if '\n' in buffer:
+            version = 3.3
             if buffer[:3] == 'RFB':
-                #~ print "rfb"
-                maj, min = [int(x) for x in buffer[3:-1].split('.')]
-                #~ print maj, min
-                if (maj, min) not in [(3,3), (3,7), (3,8), (3,889), (4,0)]:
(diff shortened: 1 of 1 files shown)


vncdotool/rfb.py
```diff
@@ -125,15 +125,22 @@ def __init__(self):
...
+            version = 3.3
             if buffer[:3] == 'RFB':
-                #~ print "rfb"
-                maj, min = [int(x) for x in buffer[3:-1].split('.')]
-                #~ print maj, min
-                if (maj, min) not in [(3,3), (3,7), (3,8), (3,889), (4,0)]:
-                    log.msg("wrong protocol version, %s.%s\n", maj, min)
-                    self.transport.loseConnection()
+                version_server = float(buffer[3:-1].replace('0', ''))
+                SUPPORTED_VESRSIONS = (3.3,)
```
sibson: spelling error

sibson:
sorry its took me so long to look at this, if you could fix the spelling error and merge master so we can get the tests passing I'd be happy to merge it.

easybe:
Hi, completely forgot about this change. I fixed the typo and rebased. Thanx.

sibson:
thanks from rainy Vancouver and be careful the plug doesn't get wet :)

Issue #26, opened 2014-04-26
Non-letter characters which require "shift-" not handled by --force-caps


TylerOderkirk:
With vncdotool v0.8.0 and the VNC server provided by QEMU v1.5.0 (Debian 1.5.0+dfsg-3ubuntu5.3) I had trouble sending an URL with...

```shell
vncdo -v -v --force-caps --delay=1000  -s localhost:3 type "http://foo.com/my_file"
```
It became...

```
http;//foo.com/my-file
```

My (non-locale-portable?) workaround was to change the following in client.py from...
```python
if self.factory.force_caps and key.isupper():
```
to...
```python
if self.factory.force_caps and (key.isupper() or key in "~!@#$%^&*()_+{}|:\"<>?"):
```
Apologies if I've misunderstood something about locales or keymaps.

This might be related to these closed bugs: https://github.com/sibson/vncdotool/issues/3 https://github.com/sibson/vncdotool/issues/9

PR #35, opened 2015-01-08
Fix decoding uppercase letters, special keys. US key board.

aragats:
I would like to suggest this little fix.

Bug: decodeKey function worked in wrong way for special keys (~!@# and so on). and for lower case letters.

This little change is based on the follow principle:
1. CAPS_LOCK affects only letters. 
2. if CAPS_LOCK is true, then it affects letters as follows: lower case letters become uppercase and uppercase letters become lowercase
3. key.isupper() returns true only for uppercase letters (AQWEDF). For special keys, it returns false.
4. There are special keys, which we can type without holding SHIFT, and there are which require holding SHIFT. As I mentioned,  key.isupper()  does not return true for special keys typed with holding key. That's why we should have a list of special keys which must be typed with holding SHIFT.  Of course, it depends on keyboard layout, but I imply the US keyboard layout. The adjustment for other keyboards could be done later, if it will be required.

### DIFF
--- a/vncdotool/client.py
+++ b/vncdotool/client.py
@@ -131,7 +131,9 @@ class VNCDoToolClient(rfb.RFBClient):
     cmask = None
 
     def _decodeKey(self, key):
-        if self.factory.force_caps and key.isupper():
+        if (self.factory.force_caps and key.isalpha() and key.islower()) \
+                or (key.isupper()) \
+                or (key in "~!@#$%^&*()_+{}|:\"<>?"):
             key = 'shift-%c' % key
 
         if len(key) == 1:
--- a/vncdotool/client.py
+++ b/vncdotool/client.py
@@ -102,6 +102,8 @@ KEYMAP = {
     'kpenter': rfb.KEY_KP_Enter,
 }
 
+SPECIAL_KEYS_US = "~!@#$%^&*()_+{}|:\"<>?"
+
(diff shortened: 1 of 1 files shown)


vncdotool/client.py
```diff
@@ -131,7 +133,9 @@ class VNCDoToolClient(rfb.RFBClient):
     cmask = None
 
     def _decodeKey(self, key):
-        if self.factory.force_caps and key.isupper():
+        if (self.factory.force_caps and key.isalpha() and key.islower()) \
```
sibson: is it possible to send a lower case a when force_caps is true?

aragats: No, when self.factory.force_caps is true and you send lower case letter then it becomes upper case letter. It is like when you are typing with switched on CAPS LOCK.

sibson: ok, I like points 3 and 4, with the aim of fixing special keys #26 but I don't understand why the CAPS LOCK case inversion behaviour is desirable.  --force-caps is intended to work around vncservers that aren't properly handling upper case letters.

sibson:
--force-caps is not intended to simulate caps-lock.  Rather, it forces [A-Z] to be prepended with shift. As some servers treat [A-Z] the same as [a-z].   I'm not convinced making it behave like caps-lock is desirable.  Are there use cases where you need cap-lock like behaviour?

PR #37, opened 2015-01-15
Measure error tolerance in MSE and support negative amounts for negation

gw0:
In some use cases measuring error tolerance in RMSE (root mean square error) may produce values that are almost 0. Therefore MSE (mean square error) seems like a simpler measure (square root does not need to be computed) and could also be computed by hand.

Additionally I added support for negated matching (matches if difference is greater than MSE) in case the user inputs a negative MSE value.

### DIFF
--- a/vncdotool/client.py
+++ b/vncdotool/client.py
@@ -240,50 +240,52 @@ class VNCDoToolClient(rfb.RFBClient):
 
         return self
 
-    def expectScreen(self, filename, maxrms=0):
+    def expectScreen(self, filename, maxmse=0):
         """ Wait until the display matches a target image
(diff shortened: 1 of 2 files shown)


sibson:
I understand that MSE can be more useful that RMSE in some cases.  Unfortunately, I can't accept this PR as is.  It would be too large change in behaviour, for other users.  I'd be open to a discussion about supporting multiple similarity measures.  I'm not sure if a --mse switch or new expect_mse command would be the cleanest option.

gw0:
Than leave RMSE, but add negated matching support. This was the main reason for my modifications.

PR #43, opened 2015-06-09
Update client.py

bobturbos:
With this modification works fine with WVWare ESXi VNC server

### DIFF
--- a/vncdotool/client.py
+++ b/vncdotool/client.py
@@ -366,7 +366,7 @@ class VNCDoToolClient(rfb.RFBClient):
             return
 
         size = (width, height)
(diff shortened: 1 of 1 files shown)


sibson:
thanks, I love to include a fix for ESXi servers, though it looks like the tests are failing and I suspect it will break when used with other VNC servers.  There is probably something during initial protocol negotiation that we can use to identify if we need to use BGRX or RGBX.  redshift/blueshift seem promising, https://github.com/sibson/vncdotool/blob/master/vncdotool/rfb.py#L196.

sibson:
As this PR stands it will break non-ESXi servers and is backwards incompatible.  We need to find a way to control which encoding is selected.  If you are able to do that then I'd be happy to merge the PR.

PR #50, opened 2015-09-28
Fix mouse press

sibson:
@pevogam re https://github.com/sibson/vncdotool/commit/e9526e90f048db957820937d5de5fbf01cd4d222#commitcomment-13366266, could you let me know if this helps?

### DIFF
--- a/docs/library.rst
+++ b/docs/library.rst
@@ -20,7 +20,7 @@ For example::
 This can be used to automate the starting of an Virtual Machine or other application::
 
     vmtool.start('myvirtualmachine.img')
-    client.connect('vmaddress:123')
+    client.connect('vmaddress::5950')
     client.expectScreen('booted.png')
     for k in 'username':
         client.keyPress(k)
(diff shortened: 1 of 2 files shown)


pevogam:
Unfortunately the problem is still not fixed - the vnc server I access with the tool is qemu (QEMU emulator version 1.5.3) in case you need some more debugging information. I will go with the workaround of manually pressing mouseUp for now.

sibson:
hmm, I wonder if the mouseUp is occurring too quickly.  Would you mind testing if a sleep between the mouseDown and mouseUp, [here](https://github.com/sibson/vncdotool/pull/50/files#diff-067798b2bc9042b38552c3f994de8626R196), makes a difference?

pevogam:
Unfortunately not. Here is a screenshot of the mouse performing single left click at position (300,400) and then simply moving to position (400,400). As you can see from the screenshot, the mouse remains clicked.

click_lock

PR #60, opened 2016-01-25
run functests on travis

sibson:
https://docs.travis-ci.com/user/migrating-from-legacy/#How-Do-I-Install-Custom-Software%3F


sibson:
just about ready to give up

PR #67, opened 2016-05-31
Add a 'typefile' command

RoganDawes:
This enables a commandline like:

vncdo -s server -p password typefile abc.txt

to type the contents of abc.txt into the target server

Limitations:

This is only intended to support typeable characters, so trying to type a binary file
will fail.

This required adding of mappings for \n and \t to the relevant keys, which was enough
for my requirements, but there may be others that need to be added for full coverage.

### DIFF
--- a/vncdotool/command.py
+++ b/vncdotool/command.py
@@ -87,6 +87,7 @@ class VNCDoToolOptionParser(optparse.OptionParser):
             'Common Commands (CMD):',
             '  key KEY\t\tsend KEY to server, alphanumeric or keysym: ctrl-c, del',
             '  type TEXT\t\tsend alphanumeric string of TEXT',
+            '  typefile FILENAME\t\ttype out the contents of FILENAME',
             '  move X Y\t\tmove the mouse cursor to position X,Y',
             '  click BUTTON\t\tsend a mouse BUTTON click',
             '  capture FILE\t\tsave current screen as FILE',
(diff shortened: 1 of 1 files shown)


sibson:
I'm +0.5 on this.  My hesitation is around adding the escaped chars to the KEYMAP, I'd deliberately avoided doing that in the past.  It might be better to parse the \n into `enter` during the file open/parse stage rather than adding to the keymap.

Could you include some tests, along the lines of https://github.com/sibson/vncdotool/blob/master/tests/unit/test_command.py#L67 and https://github.com/sibson/vncdotool/blob/master/tests/functional/test_send_events.py#L57

RoganDawes:
Hi Marc,

Can you explain your reluctance to add the escaped chars to the keymap? I get that it is not a "name" for the key, but surely it still fits as the key side of the map?

I will look at the tests that you indicated, and try to include tests for this new functionality.

sibson:
I can't recall the reason I had previously avoided it :( For now I suspect adding "escaped" characters to a CLI is going to be poor UX.  I'm thinking about trying to figure out how to type ```vncdo type \\\\n``` into bash.  Having a single clear way of use ```enter``` feels like a better option. 

If we handle it in the file parsing code we'll be better able to reason about different line endings if that becomes an issue.  Doing the conversion "magically" via the KEYMAP means we won't have that context if needed.

RoganDawes:
Hi,

I understand your reasoning around line ending confusion, but I think that the reality is that when you "press the enter key", the receiving OS makes the decision around what line ending to use. As far as I can tell, when using a regular vnc client, there is never a CR/LF sent when you press enter. As such, it doesn't make sense to send it when using a scripted tool either. So, "handling it in the file parsing code" sounds pretty much like "skip LF anyway" to me. 

From a CLI perspective, it was never the intention for the user to supplied escaped line endings in that way. The "advertised" way to do it would remain as is.

sibson:
I don't understand why putting ```\t``` and ```\r``` in  KEYMAP is preferable to doing an explicit conversion in https://github.com/sibson/vncdotool/pull/67/files#diff-d87a1110d3f48a286b362afd5a2db147R150.  Perhaps you could explain your reasoning?

RoganDawes:
I'm happy to do it either way. Let me make the changes in the "typefile" command as you suggest, and see what is involved.

RoganDawes:
Updated changes are at https://github.com/RoganDawes/vncdotool/commit/6ccd117447a6e4f97a32328a32d0aba9d44a3ab7

Not sure of the best way to update this pull request. Should I just do another one?

sibson:
Hi, the PR is automatically updated.  From a quick skim it looks good ✨ .  Thanks for updating it.  I'll take a proper look at it shortly before merging.

PR #71, opened 2016-10-08
support python 27 and 3

jamtwister:
To support python 27 and python 3 I adapted following parts:
- use print only as function
- change to relative imports
- use new style to except an error
- change from string to bytes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: python
 python:
   - "2.6"
   - "2.7"
+  - "3.5"
 
 # command to install dependencies
 install:
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,9 +1,10 @@
 0.10.1 (unreleased)
 --------------------
+  - add api.client.disconnect()
 
 0.10.0 (2016-03-03)
 ------------------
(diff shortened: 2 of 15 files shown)


sibson:
✨ 🍰 ✨ , thanks for putting this together.  It looks good.  To ensure we don't relapse and break py3 support I'd like to get CI setup to test both 2.x and 3.x before merging.  I'm not sure what the best configuration is these days or when I'll have a chance to investigate further.

sibson:
If you merge the latest master and grab the changes from [toxify-35](https://github.com/sibson/vncdotool/compare/toxify-35?expand=1) that should be enough to get tests running on CI.

jamtwister:
I am not sure, if I did it the right way.

jamtwister:
Sorry. Seems like I did wrong.
I wanted to update my pull request and not generating a new one.

jamtwister:
It seems to work ;)

jamtwister:
Hi Marc, 
should I delete my fork, when you merged this pull request?
I do not know the common behavior in that case.

sibson:
thanks for updating your PR. 

What you do with your fork is totally up to you. I'm not aware of any etiquette that should be followed. If you keep it around its one less step, if ever you want to make another PR ;)

PR #76, opened 2016-10-19
interger division due to py3

jamtwister:
Sorry I miss one interger division at rfb.
Looked at all files twice again to not overlook stuff like this again.


sibson:
great thanks for following up and fixing ✨ 🍰 ✨ .

PR #80, opened 2016-12-08
Allow to customize the client class in ThreadedVNCClientProxy

alexklaeser:
I would like to suggest to access factory.protocol instead of VNCDoToolClient class directly in api.py. This change allows to alter the factory class in order to included one's own client class via:

`VNCDoToolFactory.protocol = MyVNCClient`


sibson:
Thanks for the PR.  You're right it should have been that way from the start.

PR #81, opened 2016-12-20
Fix debian packaging configuration

alexklaeser:
Hi, I needed to apply these changes in order to succeed building and installing vncdotool as Debian package.
best regards
Alex

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,11 @@ Source: vncdotool
 Maintainer: Marc Sibson <sibson+vncdotool@gmail.com>
 Section: python
 Priority: optional
-Build-Depends: python-setuptools (>= 0.6b3), debhelper (>= 7), python-support (>= 0.8.4)
+Build-Depends:
+ python-setuptools (>= 0.6b3),
+ debhelper (>= 7),
+ python-support (>= 0.8.4),
+ python-sphinx
 Standards-Version: 3.8.4
 
 Package: python-vncdotool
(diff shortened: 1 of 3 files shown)


debian/control
```diff
@@ -2,7 +2,11 @@ Source: vncdotool
 Maintainer: Marc Sibson <sibson+vncdotool@gmail.com>
 Section: python
 Priority: optional
-Build-Depends: python-setuptools (>= 0.6b3), debhelper (>= 7), python-support (>= 0.8.4)
+Build-Depends:
+ python-setuptools (>= 0.6b3),
+ debhelper (>= 7),
+ python-support (>= 0.8.4),
+ python-sphinx
```
sibson: somewhat surprised we need python-sphinx, as its only used to build docs.

sibson:
thanks, I'd be happy to merge once the need for python-sphinx is confirmed.

debian/control
sibson: that's for checking that.  Sorry, I'm out of touch with packaging.  Does the package use the built docs? Or is debhelper running the make command simply because it detected a makefile present?  I've created a new [issue](https://github.com/sibson/vncdotool/issues/82) to track if we should change the packaging to exclude the docs and allow this PR to be merged now.

alexklaeser: yes, debian/rules will call all the dh-helper tools. For the build target, dh will also call dh_auto_build which will call the Makefile. It would be possible to include the docs in a separate vncdotool-doc package, for instance.
best regards
Alex

PR #83, opened 2016-12-21
Provide HTML doc pages via package python-vncdotool-doc

alexklaeser:
You could provide the HTML documentation via an additional package python-vncdotool-doc. I added the package in this commit here.


sibson:
awesome thanks, ✨ 🍰 ✨

PR #85, opened 2017-01-11
Add a 'pastefile' command

RoganDawes:
This sends the contents of the file to the VNC server, using the ClientCutText
message.


sibson:
I'd be happy to include this PR if you are able to add some tests for the new command.  Otherwise, its becomes difficult for me to ensure it continues to work in future releases.

PR #87, opened 2017-02-02
Implemented DesktopSize Pseudo-Encoding.

rebasegod:
Hi sibson,
I've added the DesktopSize Pseudo-Encoding [specified in the RFB protocol](https://tools.ietf.org/html/rfc6143#section-7.8.2). As [suggested](http://vncdotool.readthedocs.io/en/latest/rfbproto.html?highlight=size#client-semantics) the data of the previous frambuffer is retained. Feel free to merge :)

### DIFF
--- a/vncdotool/client.py
+++ b/vncdotool/client.py
@@ -412,6 +412,11 @@ class VNCDoToolClient(rfb.RFBClient):
         y = self.y - self.cfocus[1]
         self.screen.paste(self.cursor, (x, y), self.cmask)
 
+    def updateDesktopSize(self, width, height):
+        new_screen = Image.new("RGB", (width, height), "black")
+        new_screen.paste(self.screen, (0, 0))
+        self.screen = new_screen
+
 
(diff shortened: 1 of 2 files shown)


sibson:
🍰 ✨ 🍰  Thanks for submitting this.  It seems reasonable.  Some tests would be nice to avoid regressions, as I've no idea how to test it in future releases.

rebasegod:
Thanks for merging. If I can find some time I'll take a look at how your tests work, maybe I can come up with some test. The way I tested it was with a VM, that booted and thus changed it's resolution.

rebasegod:
I didn't do this yet, because I wanted to avoid regression, but maybe the DesktopSIze Pseudo-Encoding should be enabled by default, to give a better out-of-the-box experience to users. I think enabling it shouldn't break anything, except when someone used the "cropping-effect".

sibson:
I agree, a better experience is worth the regression.  We can bump the minor version on the next release to signal the change.

rebasegod:
Alright, I've created #88.

PR #88, opened 2017-02-13
Added the DesktopSize Pseudo-Encoding as default.

rebasegod:
As discussed in #87 vncdotool will now signal that it supports the DesktopSize Pseudo-Encoding to
the server by default.

### DIFF
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -40,7 +40,7 @@ class TestVNCDoToolClient(TestCase):
         cli.vncConnectionMade()
         factory = cli.factory
         factory.clientConnectionMade.assert_called_once_with(cli)
-        self.client.setEncodings.assert_called_once_with([client.rfb.RAW_ENCODING, client.rfb.PSEUDO_CURSOR_ENCODING])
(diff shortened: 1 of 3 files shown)


sibson:
Thanks, it looks like tests also need updating.  It might also be a good idea to expose a --disable-desktop-resizing option in the CLI for backwards compatibility  Though that could easily be a separate PR.

rebasegod:
I'm afraid I didn't understand how the tests work, since they still fail after c063ca6 .

Also I am unable to test if the --disable-desktop-resizing flag works fully, since I don't know how to test this. I just ran my application and did `vncdo --disable-desktop-resizing -s myserver type "hello world"` and  `vncdo -s myserver type "hello world"` without problems.

sibson:
You were on the right track. I'd started a similar PR and had to do some local debugging to figure out the missing [mock magic](https://github.com/sibson/vncdotool/compare/pseudodesktop?expand=1#diff-87020ebeb7f433bac8fc936494fe2c75R14).  I think that odd syntax rather than using a class decorator is a relic of supporting an old python version.  Rather than ask you to make a trivial update, I've cherry picked your changes into another PR #91 and will merge that in favour of this one.

PR #89, opened 2017-02-17
refactored due to PEP8

jamtwister:
refactored many thing with respect to PEP8

- indentation
- whitespaces
- method names -> lower case
- variable names -> lower case
- constant names -> upper case


sibson:
Thanks for putting this together, I appreciate your diligence and effort.  It poses an interesting problem for me.  I've been a bad/lazy/pragmatic OSS contributor.  The pydes and rfb files are included from the upstream versions of https://github.com/techtonik/python-vnc-viewer.  I've made some minor modifications that I intended to push back upstream but never got around too.  Merging this PR will likely make the status of these files as a local fork permanent.

sibson:
I'd like to keep the option of merging changes back to RFB and this will prevent us being able to do that.  The path forward here is, to work on merging vncdotool rfb fixes into upstream, convert vncdotool to use the fixed upstream and then pep8'ify upstream if its agreeable with the maintainer there.

jamtwister:
Ok. I try to apply the adaptations here: [https://github.com/techtonik/python-vnc-viewer](url)

PR #91, opened 2017-03-02
Pseudodesktop

sibson:


sibson:
@rebasegod thanks for putting the bulk of the work for this together in #88.

PR #105, opened 2017-11-01
Fix password check with XTightVNC server and updating the first screen

askervin:
I could not get connection and screenshot working without these fixes with a XTightVNC server that requires password.

First issue: server closed connection right after client sent the password. This got fixed by changing the password code back to what is used to be commit 659a9e3.

Second issue: getting the first screenshot failed to a PIL error. This got fixed by not trying to update non-existing screenshot.


sibson:
thanks, ✨ 🍰 ✨

Issue #108, opened 2017-11-16
Feature Request: Timeout functionality in API


gottsman:
As described in the title, looking to start a thread on add a timeout feature to the API. Its available on the CLI, but the API would be nice as well.

sibson:
some ideas are in https://github.com/sibson/vncdotool/issues/70#issuecomment-344780105

gottsman:
Confirming that the script has worked wonders in my script. Thank you @noxdafox for the patch.

noxdafox:
Np, glad it helped!

PR #111, opened 2017-12-14
issue #108: add timeout to blocking API client

noxdafox:
The timeout is a property of the returned client.

VNCDoException is raised on timeout.

### DIFF
--- a/docs/library.rst
+++ b/docs/library.rst
@@ -9,7 +9,7 @@ To use the syncronous API you can do the following::
     from vncdotool import api
     client = api.connect('vnchost:display', password=None)
 
-You can then call any of the methods available on 
+You can then call any of the methods available on
 :class:`vncdotool.client.VNCDoToolClient` and they will block until completion.
 For example::
 
@@ -17,7 +17,19 @@ For example::
     client.keyPress('enter')
     client.expectScreen('login_success.png', maxrms=10)
 
-This can be used to automate the starting of an Virtual Machine or other application::
+It is possible to set a per-client timeout in seconds to prevent calls from blocking indefinitely.
+
+::
+
+    client.timeout = 10
+    try:
+        client.captureScreen('screenshot.png')
+    except VNCDoException:
+        print('Timeout when capturing screen')
+
+In case of too many timeout errors, it is recommended to reset the client connection via the `disconnect` and `connect` methods.
+
+The syncronous API can be used to automate the starting of a Virtual Machine or other application::
 
     vmtool.start('myvirtualmachine.img')
     client.connect('vmaddress::5950')
--- a/vncdotool/api.py
+++ b/vncdotool/api.py
@@ -80,6 +80,17 @@ class ThreadedVNCClientProxy(object):
     def __init__(self, factory):
         self.factory = factory
         self.queue = queue.Queue()
+        self._timeout = 60 * 60
+
+    @property
+    def timeout(self):
+        """Timeout in seconds for API requests."""
+        return self._timeout
+
+    @timeout.setter
+    def timeout(self, timeout):
+        """Timeout in seconds for API requests."""
+        self._timeout = timeout
 
     def connect(self, host, port=5900):
         reactor.callWhenRunning(reactor.connectTCP, host, port, self.factory)
@@ -106,7 +117,11 @@ class ThreadedVNCClientProxy(object):
         def proxy_call(*args, **kwargs):
             reactor.callFromThread(self.factory.deferred.addCallbacks,
                                    callback, errback, args, kwargs)
-            result = self.queue.get(timeout=60 * 60)
+            try:
+                result = self.queue.get(timeout=self._timeout)
+            except queue.Empty:
+                raise VNCDoException("Timeout of waiting for client response")
+
             if isinstance(result, Failure):
                 raise VNCDoException(result)
 


noxdafox:
I opted for a property on the Client object as it's a good compromise between a global configuration value set once during initialization and an optional parameter available on all methods.

I tested on my setup and worked flawlessly. I am not sure how it will behave in case the async Client loop will be stuck while serving the previous request.

A better approach would consist in forwarding the timeout to the async Client loop and ensure is honoured on that side as well.

vncdotool/api.py
```diff
@@ -106,7 +117,11 @@ def result_callback(result):
         def proxy_call(*args, **kwargs):
             reactor.callFromThread(self.factory.deferred.addCallbacks,
                                    callback, errback, args, kwargs)
-            result = self.queue.get(timeout=60 * 60)
+            try:
+                result = self.queue.get(timeout=self._timeout)
+            except queue.Empty:
+                raise VNCDoException("Timeout of waiting for client response")
```
sibson: What if we disconnected the client connection before raising the exception?  Then we would have a deterministic state for the next request. As you pointed out, the current behaviour is going to be confusing to users if they try to make another request.

noxdafox: I was originally thinking about something similar.

Yet I am not very familiar with VNC protocols. How much burden would such approach put on the server? What if the server is just overloaded and hence slow? Will this approach make it even worst?

sibson: I'm -0.5 on re-connecting but willing to listen to other points of view.  The server load is a valid concern but not the one I'm most concerned about.  We don't know what state end-user applications are maintaining or relying on and how it might be impacted by a reconnect.  Specifically, where is the mouse pointer, what is the current framebuffer? If we raise then we give users an explicit opportunity to reset and re-connect.  Silent reconnection would be hard to debug.

noxdafox: Very valid point indeed. 

Nevertheless, rather than disconnecting I would simply raise the timeout error and let the user decide whether to re-set the connection or try again.

The issue with disconnecting on error is that is not a commonly expected behaviour. Sockets are not automatically closed on send/recv error as well as for files with read/write errors.

What we could do is offering a `contextmanager` type of interface so users could build a solid logic as with regular IO.

```
with connect('localhost') as client:
    while True:
        client.captureScreen('foo.png')
        time.sleep(1)
```

The above logic is quite pythonic and makes clear that the `client` would not be functional outside the `with` statement.

I can easily add the `contextmanager` protocol to the class in this PR.

sibson: As much as I dislike the socket() api I see your point about expectations.  The contextmanager is a good idea too but solves a different issue and probably good for another PR.  

If you raise the Timeout, how would a user recover?  The Twisted side of the Queue is still stuck waiting for a response from the Server.  Perhaps, the answer for now is ```raise Timeout``` and document the expectation that users disconnect. That leaves the door open for smarter options later.  How does that sound?

noxdafox: Yes it sounds pretty reasonable. Shall I document it? Where?

Right now, we are not sure that all the Timeout errors will require a recovery. It might very much be that most of the time you can just re-try. 

Moreover, I kinda dislike implicit side-effects during error handling. Feels more natural to try a couple of times and re-set rather than getting two different error types on subsequent calls.

BTW: unrelated question, have you considered switching to asyncio in favour of Twisted at least on Python > 3.5?

sibson: 100% agree, on implicit side-effects being annoying.  For documentation, I'd go with the docstring or https://github.com/sibson/vncdotool/blob/master/docs/library.rst, I haven't thought too much about where people would look first.

Bumping the asyncio question out of this thread, https://github.com/sibson/vncdotool/issues/114

noxdafox:
Documentation updated.

noxdafox:
Ping @sibson. Any chance to get this in the next vncdotool release?

vncdotool/api.py
```diff
@@ -80,6 +80,17 @@ class ThreadedVNCClientProxy(object):
     def __init__(self, factory):
         self.factory = factory
         self.queue = queue.Queue()
+        self._timeout = 60 * 60
+
+    @property
+    def timeout(self):
```
sibson: What is the motivation behind using a property here rather than a simple data member on the class?

docs/library.rst
```diff
@@ -9,15 +9,27 @@ To use the syncronous API you can do the following::
...
 
-This can be used to automate the starting of an Virtual Machine or other application::
+It is possible to set a per-client timeout in seconds to prevent calls from blocking indefinitely.
+
+::
+
+    client.timeout = 10
+    try:
+        client.captureScreen('screenshot.png')
+    except VNCDoException:
```
sibson: Seeing this here, I can't help but wonder if we should have a finer grained exceptions, e.g.```class VNCDOTimeout(VNCDoException)```

vncdotool/api.py
noxdafox: 1. A bit more future proof. If the timeout needs to be propagated to the client class, for example, this will be handled transparently.
2.  Properties allow to document and introspect the arguments.
    ```
    In [3]: client.timeout?
    Type:        property
    String form: <property object at 0x7f7ed48db548>
    Docstring:   Timeout in seconds for API requests.
    ```

docs/library.rst
noxdafox: Python 3 provides a `TimeoutError` builtin I'd rather use that than introducing a redundant VNCDoTool based exception.

For Python 2 we could define `TimeoutError` within the module.

sibson: ok, that sounds reasonable.  I'm not planning a release of vncdotool, but you'd need to make that change before I do one, as Timeout not being part of the VNCDoException hierarchy would be a breaking change.

vncdotool/api.py
sibson: thanks for the explanation, I tend to use members and convert them to properties when needed but hadn't considered the documentation aspect before.

PR #112, opened 2017-12-17
Add support for UNIX sockets

noxdafox:
Few VNC server implementations support UNIX socket transport.

UNIX sockets are simpler and faster than TCP sockets but they work
only on the local host.

### DIFF
--- a/vncdotool/api.py
+++ b/vncdotool/api.py
@@ -4,6 +4,7 @@ This feature is under developemental, you're help testing and
 debugging is appreciated.
 """
 
+import socket
 import threading
 try:
     import queue
@@ -60,9 +61,8 @@ def connect(server, password=None):
     if password is not None:
         factory.password = password
 
-    host, port = command.parse_host(server)
     client = ThreadedVNCClientProxy(factory)
-    client.connect(host, port)
+    client.connect(server)
 
     return client
 
@@ -81,8 +81,14 @@ class ThreadedVNCClientProxy(object):
         self.factory = factory
         self.queue = queue.Queue()
 
-    def connect(self, host, port=5900):
-        reactor.callWhenRunning(reactor.connectTCP, host, port, self.factory)
+    def connect(self, server):
+        address_family, host, port = command.parse_server(server)
+
+        if address_family == socket.AF_INET:
+            reactor.callWhenRunning(
+                reactor.connectTCP, host, port, self.factory)
+        elif address_family == socket.AF_UNIX:
+            reactor.callWhenRunning(reactor.connectUNIX, host, self.factory)
 
     def disconnect(self):
         def disconnector(protocol):
--- a/vncdotool/api.py
+++ b/vncdotool/api.py
@@ -18,7 +18,7 @@ from twisted.python.log import PythonLoggingObserver
 from twisted.python.failure import Failure
 
 from . import command
-from .client import VNCDoToolFactory
+from .client import VNCDoToolFactory, factory_connect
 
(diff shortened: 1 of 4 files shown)


vncdotool/api.py
```diff
@@ -81,8 +81,14 @@ def __init__(self, factory):
         self.factory = factory
         self.queue = queue.Queue()
 
-    def connect(self, host, port=5900):
-        reactor.callWhenRunning(reactor.connectTCP, host, port, self.factory)
+    def connect(self, server):
```
sibson: This is a backwards incompatible change to the API.  I'm not sure if the right answer is a new method, new kwarg or something else.

vncdotool/command.py
```diff
@@ -222,7 +223,10 @@ def build_tool(options, args):
     build_command_list(factory, args, options.delay, options.warp)
 
     # connect
-    reactor.connectTCP(options.host, int(options.port), factory)
+    if options.address_family == socket.AF_INET:
```
sibson: This conditional feels like a code smell.  Perhaps there is a better way to abstract the logic here?

vncdotool/command.py
```diff
@@ -277,22 +281,27 @@ def setup_logging(options):
...
-def parse_host(server):
+def parse_server(server):
     split = server.split(':')
 
     if not split[0]:
         host = '127.0.0.1'
     else:
         host = split[0]
 
+    if os.path.exists(host):
```
sibson: What happens if its a unix socket style path but the socket doesn't exist?

vncdotool/command.py
noxdafox: Not sure I understand, the `socket.AF_*` constants are pretty standard when it comes to indicating the [Address Family](https://docs.python.org/3/library/socket.html#socket-families) of a socket.

sibson: Both command and [api.py](https://github.com/sibson/vncdotool/pull/112/files#diff-e45ab1b1ae5f9a01d36a4a4913f9c101R87) have a similar if/then block to create the connection.  Can this be refactored into the client.py class somehow? It seems like it might be a leaky abstraction.  What happens if we add a third type of connection?  It's probably fine for now and not a blocker.

noxdafox: Now I got it, good point. I'll fix it asap.

vncdotool/command.py
noxdafox: You get the following error which is the same error you would get with the current logic if instead of passing a valid hostname you pass a path.

```
VNCDoException: [Failure instance: Traceback (failure with no frames): <class 'twisted.internet.error.DNSLookupError'>: DNS lookup failed: address '/foo/bar/baz' not found: [Errno -2] Name or service not known.
]
```

There is not a proper way to distinguish between a hostname and a relative path at a string level. Someone could for example input the following relative path: `example.com/unix.skt`.

Only working approach would be calling [getaddrinfo](https://docs.python.org/3/library/socket.html#socket.getaddrinfo) if `os.path.exists` returns `False` to verify it's a legitimate hostname. Unfortunately, `getaddrinfo` will trigger a DNS resolution which might lead to few unexpected behaviours.

sibson: good point, yuck. We could bake in a bunch of logic to guess the type but as you say ultimately it will fail.  The best we could do it provide a nicer error message or document the behaviour but that shouldn't hold up this PR.

sibson:
thanks, ✨ 🍰 ✨

PR #113, opened 2017-12-17
Add introspection support for Client API

noxdafox:
The `__getattr__` protocol adds a level of indirection to the API calls
effectively obfuscating the wrapped API.

The `__dir__` protocol instructs the interpreter on how to correctly
introspect the wrapper class.

This enables the traditional `<.>` + `<TAB>` completion in Python REPL
consoles such as IPython.

### DIFF
--- a/vncdotool/api.py
+++ b/vncdotool/api.py
@@ -114,6 +114,9 @@ class ThreadedVNCClientProxy(object):
 
         return proxy_call
 
+    def __dir__(self):
+        return dir(self.__class__) + dir(self.factory.protocol)
+
 
 if __name__ == '__main__':
     import sys


noxdafox:
I personally am a fan of the `__getattr__` protocol as it enables an elegant implementation of the compositor pattern.

A drawback though, is the loss of introspection capabilities which is quite neat when exploring APIs on consoles such as `IPython`.

The `__dir__` protocol comes in handy to fix this type of issue.

https://ipython.org/ipython-doc/3/config/integrating.html#tab-completion

sibson:
great idea thanks, ✨ 🍰 ✨, I'm assuming this just works(tm).

PR #117, opened 2018-01-06
 Fix password feature on Python 3 (broken in 97816c3) 

agfor:
Neither bytes nor string work as a password on Python 3 without this change. String blows up in this method, while bytes blows up in rfb.py.

### DIFF
--- a/vncdotool/pyDes.py
+++ b/vncdotool/pyDes.py
@@ -415,7 +415,7 @@ class des(_baseDes):
 
 	def __String_to_BitList(self, data):
 		"""Turn the string data, into a list of bits (1, 0)'s"""
-		if _pythonMajorVersion < 3:
+		if isinstance(data[0], str):
 			# Turn the strings into integers. Python 3 uses a bytes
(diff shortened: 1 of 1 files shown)


sibson:
I'm +1 on merging this as is but I'd like to understand why the ```_pythonMajorVersion``` guard doesn't appear to work, as it seems intended to address this case.

agfor:
On https://github.com/sibson/vncdotool/blob/master/vncdotool/rfb.py#L180 `\0` is a unicode literal on Python 3, so you can't add it to a bytes object, which is what the `_pythonMajorVersion` check is trying to require you use. Changing it to a bytes literal doesn't work either.

On the opposite side, the only thing stopping a string / unicode password from working on Python 3 (assuming all codepoints < 255) is not running the codepoint-by-codepoint conversion this method already implements.

If you want to support non-ASCII characters in passwords on Python 3, you'll need to rewrite the `rfb.py` side of this instead, and use bytes instead of string / unicode at that level.

sibson:
thanks for the explanation

Issue #90, opened 2017-02-23
always got black picture


yunpengb:
Hi, I use python script and try to capture a picture of one host 
the host is installed tightvnc and it's win7 system.
my code is like this:

    from vncdotool import api
    client = api.connect('192.168.1.199', 'passwd')
    client.captureScreen('pic.png')

it always leave a black screenshot, what's wrong with my code?

julianweise:
Facing exactly the same problem using the cl-tool using:
`vncdo -s 127.0.0.1 capture screen5.png`
Using Win10 and have installed TightVNCServer

MagicFab:
Same here, using TightVNCserver 2.8.8 on Windows 7 Pro.

sibson:
Has this been resolved with recent changes?

pvtmert:
okay i tried with both command line invocation and api-call via python

i also got black screen when using tightvnc 2.8.11 on windows 10 (1511)

but i used following code to get screenshot, which worked without problems:

```
with api.connect("1.2.3.4", password="mypass", timeout=10.0) as cli:
	cli.refreshScreen()
	cli.captureScreen("test.jpg")
```

Chenshennan:
I got black screen via vncdotool api:

from vncdotool import api
import time
class vnctest(object):
    def takescreenshot(self,in_host,in_password):
        vnc_host=in_host
        vnc_password=in_password
        vnc_session=api.connect(vnc_host,password=vnc_password)
        time.sleep(5)
        try:
            vnc_session.captureScreen('screenshot.png')
        except:
            print('Timeout when attempt to capuring screen')
        vnc_session.disconnect() 

test=vnctest()
test.takescreenshot('<my vnc server IP>','<my vnc server password>')

But if I use tightvnc-2.8.11(Clinet is WIN7, and server is WIN10), I can observe all veiwer.
At the same time, I run above test script. It can got normal screenshot.
Once I disconnect tighvnc veiwer and wait a few mins, the screenshot can't be use again.

pmhahn:
`vncdotool -s $HOST:$PORT  capture screen.png` works fine for me.
Maybe this 5 year old bug can be closed?

mikhail-ict:
I experience exactly the same issue with TightVNCServer...

pmhahn:
> I experience exactly the same issue with TightVNCServer...

What does "exactly" mean: You use that 5 year old version and got a black screen?

Please(!) provide **actionable** data, e.g.
- which Python version
[...]

PR #118, opened 2018-01-06
Don't take a screenshot until the framebuffer has been updated

agfor:
This probably isn't the correct way to fix this, but it should at least illustrate the idea.

Without this change, screenshots are almost always taken before the response to the framebuffer update request has been applied -- defeating the purpose of calling `framebufferUpdateRequest` in `_capture` in the first place.

With this change, code using the synchronous API looks like this:

```
screen = "screen.png"
start = os.stat(screen)        
client.captureScreen(screen)   
while os.stat(screen) == start:
    sleep(0.05)         
sleep(0.2)                     
```

The final sleep accounts for the time between when the file is opened and writing completes.

### DIFF
--- a/vncdotool/client.py
+++ b/vncdotool/client.py
@@ -129,6 +129,7 @@ class VNCDoToolClient(rfb.RFBClient):
     y = 0
     buttons = 0
     screen = None
+    screen_updated = None
     deferred = None
 
     cursor = None
@@ -243,7 +244,15 @@ class VNCDoToolClient(rfb.RFBClient):
         self.deferred.addCallback(self._captureSave, filename, *args)
         return self.deferred
 
+    def framebufferUpdateRequest(self, *args):
+        self.screen_updated = False
+        return super(VNCDoToolClient, self).framebufferUpdateRequest(*args)
+
     def _captureSave(self, data, filename, *args):
+        if not self.screen_updated:
+            reactor.callLater(0.05, self._captureSave, data, filename, *args)
+            return self
+
         log.debug('captureSave %s', filename)
         if args:
             capture = self.screen.crop(args)
@@ -394,6 +403,7 @@ class VNCDoToolClient(rfb.RFBClient):
             d = self.deferred
             self.deferred = None
             d.callback(self.screen)
+        self.screen_updated = True
 
     def updateCursor(self, x, y, width, height, image, mask):
         if self.factory.nocursor:


vncdotool/client.py
```diff
@@ -394,6 +403,7 @@ def commitUpdate(self, rectangles):
             d = self.deferred
```
sibson: perhaps we need an additional ```and self.screen``` here?

vncdotool/client.py
```diff
@@ -243,7 +244,15 @@ def _capture(self, filename, *args):
         self.deferred.addCallback(self._captureSave, filename, *args)
         return self.deferred
 
+    def framebufferUpdateRequest(self, *args):
+        self.screen_updated = False
+        return super(VNCDoToolClient, self).framebufferUpdateRequest(*args)
+
     def _captureSave(self, data, filename, *args):
+        if not self.screen_updated:
```
sibson: Rather than introduce a new data member could we branch of `if not self.screen`?

sibson: If we make the change suggested in commitUpdate, I think this change may be unnecessary.

agfor: No, this is not just a problem when the screen hasn't been updated at all. It's a problem every time a screenshot is created, so you can't tell just by looking at the screen attribute whether or not it's been updated recently.

The goal of this is to always update the screen before taking a screenshot, so it's always up to date -- not just ensure we have something in the buffer, which could be arbitrarily old.

agfor:
It should definitely solve #90. Let me clarify the two parts of what I think would be an ideal solution here, one of which this PR solves, and one of which it doesn't.

- Always update the framebuffer before taking a screenshot. This is not just to make sure the buffer is initialized, but to make sure it's current when the client is long-lived. This is solved by this PR.
- Allow a callback for the async API, or blocking for the sync API, so user code can know when the screenshot is done being written without watching the filesystem. This is not solved by this PR, because I'm extremely rusty working with Twisted and not familiar with your sync API shim.

vncdotool/client.py
sibson: Thanks for the explanation. I think [_capture()](https://github.com/sibson/vncdotool/blob/3cadfcbd35c0dd85a55ce2cde0045c112f1c4b71/vncdotool/client.py#L241-L244) and [commitUpdate()](https://github.com/sibson/vncdotool/blob/3cadfcbd35c0dd85a55ce2cde0045c112f1c4b71/vncdotool/client.py#L393-L397) are trying to achieve the same as what you are doing with ```screen_updated```.  I wonder if there is an order of ops or other unexpected call to ```commitUpdate()``` that is causing it to prematurely fire the callback. In particular a pseduo-cursor or empty rect update.

vncdotool/client.py
```diff
@@ -243,7 +244,15 @@ def _capture(self, filename, *args):
         self.deferred.addCallback(self._captureSave, filename, *args)
         return self.deferred
 
+    def framebufferUpdateRequest(self, *args):
+        self.screen_updated = False
```
sibson: why do this in framebufferUpdateRequest() rather than _capture()?

agfor: Basically, because in the future I might want to use the PIL Image object directly without writing it out to disk, so I didn't want to couple it to actually saving the screenshot.

vncdotool/client.py
agfor: Yeah, at some point I changed `_capture` to return `self` instead of `self.deferred` thinking that would stop the callback getting called prematurely, but it didn't, so I'm not sure.

When I added some extra debug logging, I didn't see extra calls to any of the methods, the callback was just getting called before `commitUpdate`.

I didn't spend a lot of time trying to figure it out what exactly the cause of the problem was once I figured out how to work around it. I may have time to take another look this weekend.

sibson:
For API usage that makes sense, and makes me wonder if its a distinct enough use case that we need to refactor ```_capture()``` and make it public, rather than overloading the captureSave() to both write a file and update the screen.  I’m thinking something along the lines of

```
def refreshScreen():  # _capture() without .addCallbacks()
    self.deferred = Deferred()
    self.framebufferUpdateRequest()
    return self.deferred

def captureScreen():
    d = self.refreshScreenI()	
  d.addCallback(self._captureSave, filename, *args) # moved from _capture()
  return d
```

That assumes there is a solvable bug somewhere causing the deferred to be fired prematurely.

agfor:
I'm going to close this for now. I did some better debugging today and I think this is an artifact of my server using protocol version 3.8 while rfb.py only supports version 3.3.

There are in fact spurious `commitUpdate`s being triggered, not sure how I missed it before. The client thinks it's getting a single pixel update. I'll have to hack around the behavior (or invest the time to use a different RFB client, or update rfb.py).

agfor:
After some more investigation, what I'm getting is a real framebuffer update, but only for the upper left pixel of the screen, rather than the requested region.

I tried:
- changing the region requested to not include the `(0, 0)` pixel;
- moving the cursor first so it's not at `(0, 0)`;
- requesting an incremental update;
- and waiting several minutes between update requests;

but nothing changed the response.

If there is something in protocol version 3.8 causing this to happen, I can't figure out what -- so, like ignoring the pixel format request, it may just be weird behavior of the VMWare VNC server.

I'll be opening a couple of new PRs shortly.

PR #121, opened 2018-01-13
Allow access non-callable client attributes through the API shim

agfor:
I want to be able to access attributes of the client instance through
the API shim. Specifically, my need is for the `screen` attribute. This
is somewhat non-trivial since you can't just add getter methods to the
client -- any methods you call on the protocol need to return a deferred
or `self`.

I see two ways to approach this. One is the method in this PR, to detect
whether or not the desired attribute is callable. The other is to add a
`getSelf` method to the client, allowing you to pierce the proxy by
calling that method and accessing the attributes on the result.

There may be other, better ways I don't see because of my inexperience
with Twisted.

### DIFF
--- a/vncdotool/api.py
+++ b/vncdotool/api.py
@@ -83,14 +83,6 @@ class ThreadedVNCClientProxy(object):
         def errback(reason, *args, **kwargs):
             self.queue.put(Failure(reason))
 
-        def callback(protocol, *args, **kwargs):
-            def result_callback(result):
-                self.queue.put(result)
-                return result
-            d = maybeDeferred(method, protocol, *args, **kwargs)
-            d.addBoth(result_callback)
-            return d
-
         def proxy_call(*args, **kwargs):
             reactor.callFromThread(self.factory.deferred.addCallbacks,
                                    callback, errback, args, kwargs)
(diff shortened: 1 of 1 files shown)


sibson:
I think the 3rd alternative is to replace [connect()](https://github.com/sibson/vncdotool/blob/master/vncdotool/api.py#L71-L73) with something like 

```
def connect(...):
  def capture_protocol(protocol):
      self.protocol = protocol
   self.factory.addCallback(capture_protocol)
   reactor.callWhenRunning(
            factory_connect, self.factory, host, port, family)
```

Then the [else block](https://github.com/sibson/vncdotool/pull/121/files#diff-e45ab1b1ae5f9a01d36a4a4913f9c101R133) becomes simply

```
return getattr(self.protocol, attr)
```

Or you could access the properties directly with ```client.protocol.screen```.  I'm not sure what the cleanest solution is.  If you want to merge as is I can do so.

agfor:
I implemented your suggestion. I like that it allows the user to choose whether to access the object via the protocol instance (for example, if the attribute needed is not defined on the protocol class) or consistently with how you access callables through the proxy. It also makes the diff nicer.

sibson:
sounds good, thanks for being open to the idea.

PR #122, opened 2018-01-13
Attempt to determine the correct image mode from the pixel format

agfor:
VMWare's built-in VNC server seems to ignore `SetPixelFormat` messages.
However, it does correctly indicate the pixel format it uses in the
`ServerInit` message.

This commit attempts to use the pixel format info from that initial
message to derive the correct image mode to use for PIL. If that fails,
it falls back to the previous behavior of setting the pixel format to
the `rfb.py` default of `RGBX`.

This was the least-hacky way I could think of to fix VMWare screenshots,
while potentially giving some benefit to other implementations (skipping
the often unnecessary, if harmless, `SetPixelFormat` message).

### DIFF
--- a/vncdotool/client.py
+++ b/vncdotool/client.py
@@ -129,6 +129,7 @@ class VNCDoToolClient(rfb.RFBClient):
     y = 0
     buttons = 0
     screen = None
+    image_mode = "RGBX"
     deferred = None
 
     cursor = None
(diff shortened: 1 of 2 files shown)


vncdotool/client.py
```diff
@@ -336,6 +337,22 @@ def mouseDrag(self, x, y, step=1):
 
         return self
 
+    def setImageMode(self):
+        """ Extracts color ordering and 24 vs. 32 bpp info out of the pixel format information
+        """
+        if (self.truecolor and (not self.bigendian) and self.depth == 24
+                and self.redmax == 255 and self.greenmax == 255 and self.bluemax == 255):
```
sibson: To check I'm understanding this correctly, this is testing we're using a 3 channel 8bit colour map?

vncdotool/client.py
```diff
@@ -347,7 +364,9 @@ def vncRequestPassword(self):
         self.sendPassword(self.factory.password)
 
     def vncConnectionMade(self):
-        self.setPixelFormat()
+        if not self.setImageMode():
```
sibson: I'm finding this hard to parse. Would it make sense push setPixelFormat() inside setImageMode() to avoid the negative conditional?

vncdotool/client.py
agfor: Yeah. Basically, it can handle different color orderings, and with or without a padding / alpha byte, but that's it. So any attribute of the pixel format that I don't explicitly use to determine the image mode, I explicitly assert on here.

vncdotool/client.py
agfor: Yeah, that's better. Updated.

vncdotool/client.py
sibson: cool, I like it also opens the door to other formats being implemented, as needed. There are failing tests, seems like truecolor isn't being handled consistently.  Once those are fixed I'll merge this.

PR #123, opened 2018-01-13
Separate refreshing the screen from saving a screenshot

agfor:
This change was suggested by @sibson in https://github.com/sibson/vncdotool/pull/118#issuecomment-357261582

I want to be able to use the in-memory PIL image without writing it out
to disk, through the API shim. This commit separates out the two pieces
of functionality.

What makes this a little more complicated is that you can't return a
deferred without a callback from a method on the client, or, just as if
you hadn't returned a deferred at all, the client instance gets replaced
by another object. I haven't spent the time to learn why this is the
case. To get around that, I add a no-op callback to the deferred in
`refreshScreen`.

When actually taking a screenshot, the callback to save the image is
simply added to the same deferred.

### DIFF
--- a/vncdotool/client.py
+++ b/vncdotool/client.py
@@ -237,11 +237,16 @@ class VNCDoToolClient(rfb.RFBClient):
         log.debug('captureRegion %s', filename)
         return self._capture(filename, x, y, x+w, y+h)
 
+    def refreshScreen(self, incremental=0):
+        d = self.deferred = Deferred()
+        self.framebufferUpdateRequest(incremental=incremental)
+        d.addCallback(lambda *args: self)
+        return d
+
     def _capture(self, filename, *args):
-        self.framebufferUpdateRequest()
-        self.deferred = Deferred()
-        self.deferred.addCallback(self._captureSave, filename, *args)
-        return self.deferred
+        d = self.refreshScreen()
+        d.addCallback(self._captureSave, filename, *args)
+        return d
 
     def _captureSave(self, data, filename, *args):
         log.debug('captureSave %s', filename)
--- a/vncdotool/client.py
+++ b/vncdotool/client.py
@@ -240,7 +240,6 @@ class VNCDoToolClient(rfb.RFBClient):
     def refreshScreen(self, incremental=0):
         d = self.deferred = Deferred()
         self.framebufferUpdateRequest(incremental=incremental)
-        d.addCallback(lambda *args: self)
         return d
 
     def _capture(self, filename, *args):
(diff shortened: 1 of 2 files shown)


vncdotool/client.py
```diff
@@ -237,11 +237,16 @@ def captureRegion(self, filename, x, y, w, h):
         log.debug('captureRegion %s', filename)
         return self._capture(filename, x, y, x+w, y+h)
 
+    def refreshScreen(self, incremental=0):
+        d = self.deferred = Deferred()
+        self.framebufferUpdateRequest(incremental=incremental)
+        d.addCallback(lambda *args: self)
```
sibson: I think I've lost track of the changes we've been discussion.  What was the purpose of this lambda callback?

agfor:
Returning a deferred with no callbacks set doesn't work, and I don't know
why (I mentioned this in the pr description). This is one of the things
that I ran into when originally debugging my screenshot issues that made it
so difficult.

You should be able to reproduce by removing the callback, calling the
method, then calling another method that refers to 'self' in the body
(through the api shim, I don't know for sure how it'd work if done directly
but I believe it'll be the same). I would give an exact repro snippet but
I'm on my phone.

vncdotool/client.py
agfor: If you don't set the callback, then call `refreshScreen` twice, you get:

```
  File "./foo.py", line 63, in main
    client.refreshScreen()
  File "/Users/agf/projects/eq/vncdotool/api.py", line 108, in proxy_call
    raise VNCDoException(result)
vncdotool.api.VNCDoException: [Failure instance: Traceback: <class 'AttributeError'>: 'Image' object has no attribute 'framebufferUpdateRequest'
```

on the second attempt because the result of the callback is used in place of `self` the second time `refreshScreen` is called. Having the callback return `self` solves the problem.

I was trying to avoid changing `commitUpdate` so this change wouldn't affect the `expectScreen` code. I've just pushed a commit that updates all of that instead (including tests). Let me know what you think.

vncdotool/client.py
```diff
@@ -393,7 +397,7 @@ def commitUpdate(self, rectangles):
         if self.deferred:
             d = self.deferred
             self.deferred = None
-            d.callback(self.screen)
+            d.callback(self)
```
sibson: Sorry, I know you explained this before but I can't seem to find the comment or recall the reasoning for this change.  Was there was some downstream interaction with api.py that broke?

agfor: https://github.com/sibson/vncdotool/pull/123#discussion_r162175093

agfor: Not the code already does return `self` explicitly in a number of places, so I assume this is a Twisted thing.

sibson: Thanks for the added context there.  I was concerned about changing the return type being a breaking change But I think your right, it needs to return either self or a deferred to allow proper chaining of commands/deferreds.

PR #124, opened 2018-01-13
Support subclassing of VNCDoTool types in API shim

agfor:
I want to be able to modify the behavior of the underlying client
without monkey patching. Specifically, I need to work around VMWare VNC
server bugs.

I also want to be able to add additional behavior to the proxy object,
in my case to add additional sync methods which wrap multiple pointer
or key events.

This also includes a small change to allow setting the proxy's timeout
when calling `connect`.

With these changes, my VNC client module looks like this:

```
from vncdotool.api import ThreadedVNCClientProxy, connect as api_connect
from vncdotool.client import VNCDoToolFactory, VNCDoToolClient

from events import Events # pointer and key event wrappers
from screen import Screen # screenshot wrappers

class VNCClient(VNCDoToolClient):
     pass # VMWare VNC server bug workarounds here

class VNCFactory(VNCDoToolFactory):
    protocol = VNCClient

class VNCProxy(Events, Screen, ThreadedVNCClientProxy):
    pass

def connect(host, password = None, timeout = 2):
    return api_connect(host, password, factory_class=VNCFactory, proxy=VNCProxy, timeout=timeout)
```

### DIFF
--- a/vncdotool/api.py
+++ b/vncdotool/api.py
@@ -37,7 +37,7 @@ if sys.version_info.major == 2:
         pass
 
 
-def connect(server, password=None):
+def connect(server, password=None, factory_class=None, proxy=None, timeout=None):
     """ Connect to a VNCServer and return a Client instance that is usable
     in the main thread of non-Twisted Python Applications, EXPERIMENTAL.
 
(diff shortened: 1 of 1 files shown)


vncdotool/api.py
```diff
@@ -37,7 +37,7 @@ class TimeoutError(OSError):
         pass
 
 
-def connect(server, password=None):
+def connect(server, password=None, factory_class=None, proxy=None, timeout=None):
```
sibson: could we do use default values for the kwargs rather than conditionally set inline in the function body?

```
def connect(server, password=None, factory_class=VNCDoToolFactory, proxy=ThreadedVNCClientProxy, timeout=None):
```

agfor: Updated. It required I reorder the code so that the proxy class is defined.

sibson:
❤️ , thanks for the cleanup.

PR #125, opened 2018-01-15
Add support for RFB versions 3.7 and 3.8

agfor:
I was able to successfully connect with all three protocol versions,
with and without a password. Not providing a password when one is
required also works as expected.

However, the server I'm working with seems to closes the connection
immediately upon receiving an invalid challenge response, so I
didn't test all of the code around incorrect password handling. The
behavior is consistent with and without my patch, so isn't caused by
this change.

### DIFF
--- a/vncdotool/rfb.py
+++ b/vncdotool/rfb.py
@@ -118,6 +118,7 @@ class RFBClient(Protocol):
         self._packet_len = 0
         self._handler = self._handleInitial
         self._already_expecting = 0
+        self._version = None
 
     #------------------------------------------------------
     # states used on connection startup
@@ -129,7 +130,7 @@ class RFBClient(Protocol):
             version = 3.3
             if buffer[:3] == b'RFB':
                 version_server = float(buffer[3:-1].replace(b'0', b''))
-                SUPPORTED_VERSIONS = (3.3,)
+                SUPPORTED_VERSIONS = (3.3, 3.7, 3.8)
                 if version_server in SUPPORTED_VERSIONS:
                     version = version_server
                 else:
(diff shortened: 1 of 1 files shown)


vncdotool/rfb.py
```diff
@@ -189,14 +218,28 @@ def _handleVNCAuthResult(self, block):
...
+            else:
+                self.expect(self._handleAuthFailed, 4)
         else:
             log.msg("unknown auth response (%d)" % result)
 
+    def _handleAuthFailed(self, block):
+        (waitfor,) = unpack("!I", block)
+        self.expect(self._handleAuthMessage, waitfor)
+
+    def _handleAuthMessage(self, block):
```
sibson: On the surface I find this name confusing, as it only seems to run in the failure case.  Do you have a link to the protocol spec that outlines this flow, as that context might help my understanding.  Otherwise, this PR seems good to me.

sibson:
lgtm, if there is no need to change the name, then I'll merge as is.  Also, note it may take me some time to work through your other PRs.

vncdotool/rfb.py
agfor: Yeah, I agree. I was trying to stay consistent with the existing code as much as possible here, specifically `_handleConnMessage` in this case. I think a lot of the methods should be renamed to match the message types defined in the spec, and there are a lot of other small cleanups possible, but I didn't want to mix cosmetic and functional changes.

I'll rename this to `_handleAuthFailedMessage` which is both pretty consistent with the other code and still indicates it's only called for failure.

agfor:
Method name update pushed. No rush on getting stuff merged (though of course I appreciate it when you can), I don't have any short-term plans to distribute the code I'm building on top of this so it's easy for me to just maintain a fork.

sibson:
Thanks for the cleanup and understanding.  

I was lazy and included rfb.py from https://github.com/techtonik/python-vnc-viewer/blob/master/rfb.py.  I've tried to avoid making too many cosmetic changes in the hope of eventually pushing changes back upstream.

PR #126, opened 2018-01-24
Make the client work with badly behaved servers

agfor:
When responding to framebuffer update requests, the VMWare VNC server
will send a single-pixel update on alternate responses, even when a
full buffer update has been requested. To cancel this out, you can
resend any framebuffer update request that receives such a response.
Since this means essentially rejecting what are valid responses, I do
this in a subclass so as to avoid doing it when it's not needed.

The MacOS built-in VNC server is unusual in several ways.

- It advertises itself as supporting protocol version 3.889
- If no password is used it attemts a nonstandard authentication
  sequence
- 8-, 24-, and 32-bit pixel formats cause the server to disconnect as
  soon as the first post-handshake client request as made

The first two are already handled without changes to the client (it
falls back to operating in protocol version 3.8, which works fine, and
you just have to set a password), but the third requires special
handling. Since this is a nonstandard protocol version, and this only
has to happen once per session, it seems safe to build this directly
into `setImageMode`.

I currently just have my own subclass that supports these two
behaviors, but they're common enough servers (and both referenced in
open issues on this repo) that I think including the fixes here is
valuable. If you disagree, at least storing the server-sent version
number is helpful for detecting the MacOS server in a subclass.

### DIFF
--- a/vncdotool/client.py
+++ b/vncdotool/client.py
@@ -344,7 +344,14 @@ class VNCDoToolClient(rfb.RFBClient):
     def setImageMode(self):
         """ Extracts color ordering and 24 vs. 32 bpp info out of the pixel format information
         """
-        if (self.truecolor and (not self.bigendian) and self.depth == 24
+        if self._version_server == 3.889:
+            self.setPixelFormat(
+                    bpp = 16, depth = 16, bigendian = 0, truecolor = 1,
+                    redmax = 31, greenmax = 63, bluemax = 31,
+                    redshift = 11, greenshift = 5, blueshift = 0
+                    )
+            self.image_mode = "BGR;16"
+        elif (self.truecolor and (not self.bigendian) and self.depth == 24
                 and self.redmax == 255 and self.greenmax == 255 and self.bluemax == 255):
 
             pixel = ["X"] * self.bypp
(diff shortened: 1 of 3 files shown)


sibson:
I've yet to review the code, but wanted to mention that I'm OK with encoding server specific hacks into the vncdotool client, provided they don't negatively impact other usage.  The VNC protocol is weakly defined and lacks any type of shared validation system.  Thus, we can't take a principled stance here.  Even if we file bugs with server vendor's there is no way to ensure they are fixed or updated servers are in use.

I'm open to idea's around what the most developer friendly way to implement these exceptions is.

agfor:
OK Cool. The VMWare hack feels weird enough to me -- rejecting valid data, and inspecting every packet -- that I don't think it should be "on by default". How, specifically, we make it off by default I don't really have strong thoughts on.

vncdotool/client.py
```diff
@@ -344,7 +344,14 @@ def mouseDrag(self, x, y, step=1):
     def setImageMode(self):
         """ Extracts color ordering and 24 vs. 32 bpp info out of the pixel format information
         """
-        if (self.truecolor and (not self.bigendian) and self.depth == 24
+        if self._version_server == 3.889:
```
sibson: It would be useful to have a comment why this quirk exists, rather than requiring a reader to view the history, it could be as simple as linking back to this PR.

vncdotool/client.py
```diff
@@ -444,6 +451,16 @@ def updateDesktopSize(self, width, height):
         self.screen = new_screen
 
 
+class VMWareClient(VNCDoToolClient):
+    def dataReceived(self, data):
+        single_pixel_update = b'\x00\x01\x00\x00\x00\x00\x00\x01\x00\x01\x00\x00\x00\x00'
```
sibson: Another project with similar issues, https://glyptodon.org/jira/browse/GUAC-1247

PR #130, opened 2018-08-05
Add RPM packaging capability

pevogam:
The RPM packaging system is used on systems like Fedora, CentOS,
RHEL, etc. in comparison to Debian packaging systems.


sibson:
thanks, ✨ 🍰 ✨

PR #132, opened 2018-08-12
ignore error on super().updateDesktopSize()

mihailstoynov:


sibson:
I can't merge this without a better understanding of what problem this is resolving.  Blindly catching and discarding exceptions leads to very difficult to debug problems in the future.  Can you provide more information about the problem?

mihailstoynov:
Hi,
I understand your reasoning. I don't remember what problem this change solved - if there is time, I'll revert it and see what the issue was.

pvtmert:
hi there, probably when you connect kvm/qemu/xen hypervisor console, resolution can change because of boot sequence/card initialization.

maybe problems with that?

mihailstoynov:
I really don't remember.

hilbix:
AFAICS this is not the correct fix.  IMHO the bug stems from

https://github.com/sibson/vncdotool/blob/e133a8916efaa0f5ed421e0aa737196624635b0c/vncdotool/client.py#L447-L451

where `Image.new('RGB', (-1, -1))` raises a ValueError.  Some RFB-server might sends some very irregular sizes, perhaps when the Screen's VM hardware is (re)initialized.

Another way would be to fix it there:

https://github.com/sibson/vncdotool/blob/e133a8916efaa0f5ed421e0aa737196624635b0c/vncdotool/rfb.py#L505-L507

The idea is to sanitize what is sent to `updateDesktopSize()` such that the size is kept in 0 to some sane (configurable) upper limit which - perhaps - defaults to 1M (if ever can construct screens bigger than 1Mx1M pixels, we are hopefully wise enough to not try to transfer this using RFB).

Fun fact:

`Image.new('RGB', (1000000, 1000000))` gives `Killed process 22663 (python) total-vm:2054946160kB, anon-rss:563604kB, file-rss:4kB, shmem-rss:0kB`, while
`Image.new('RGB', (10000000, 10000000))` just raises a `MemoryError`.

Probably `ulimit -m unlimited` is not the wisest setting ;)

sibson:
Some type of sanitize and discard when receiving invalid/crazy sizes seems like a better approach than blindly ignoring exceptions.

sibson (changes requested):
sanitize input rather than discard exceptions, per convo

pmhahn:
I implemented this in f7f9138 in my PR #232 
FB width and height are `u16` so maximum `65535`

PR #135, opened 2018-09-28
Use cropped image for compare

mfuerni:
The histogram should be computed from the cropped image not from the full screen. Otherwise, expectRegion does not work.


sibson:
thanks!

PR #137, opened 2018-12-21
added port argument to api.connect

pvtmert:
hi there,

i noticed server argument to `api.connect` function parses it and gives port number automatically. In order to use standard tcp/ip port double colon (`::`) needed as seperator

i mostly use this library through ssh-forward, thus port numbers generally in 10k space.
specifying port number as an argument simpler than `host + "::" + port` imho, also looks better

only downside is that when specified, port argument overrides port if server was given with display number or port

### DIFF
--- a/vncdotool/api.py
+++ b/vncdotool/api.py
@@ -118,7 +118,7 @@ class ThreadedVNCClientProxy(object):
         return dir(self.__class__) + dir(self.factory.protocol)
 
 
-def connect(server, password=None,
+def connect(server, password=None, port=None,
         factory_class=VNCDoToolFactory, proxy=ThreadedVNCClientProxy, timeout=None):
     """ Connect to a VNCServer and return a Client instance that is usable
     in the main thread of non-Twisted Python Applications, EXPERIMENTAL.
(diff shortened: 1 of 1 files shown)


sibson:
Unless I've misunderstood the problem, that this is mostly a stylistic question of API design and I'm not convinced the solution is much better than the problem it is trying to solve. It trades one type of API design wart for another, specifically, the slightly cumbersome passing of ```port``` embedded in the ```server``` argument with an invisible overriding ```port```.  

The change needs to be a clear improvement and be backwards compatible. As it stands it's a small improvement and a small step back, which doesn't seem worth possible costs of debugging why isn't my ::port being honoured.  If there was a way to improve the API without compromising simplify I'd be willing to accept the PR.   Allowing ```server``` to be a structured string was probably a mistake in the first place but I don't know its egregious enough to warrant breaking working code to fix.

sibson:
@pvtmert I see you 👍 my comment.  Should we close this PR or do you want to refine it to address the issue a different way?

pvtmert:
yeah closing now, (totally forgot last time :) )

PR #142, opened 2019-01-25
Allow expect to request non-incremental frame buffer update

### LINKED ISSUE #141: (r)expect does not check the current screen
`expect` and `rexpect` commands fail to match if the current screen matches the provided image.

 - VNC Server: VNC server of QEMU emulator version 3.1.0
 - vncdotool 0.12.1

1. Start the QEMU instance, let it boot until you get the screen that you will "expect".
2. Do an "expect" operation, VNC hangs forever.
3. Switch to another terminal (possibly with different image) and switch back,
4. "expect" now matches.

If I switch [incremental to 0 on client.py](https://github.com/sibson/vncdotool/blob/d308133ee66a95c849f2d9b05faeeb87fc120561/vncdotool/client.py#L281), then "expect" matches the current screen as expected.

tghosgor:
Closes #141

Signed-off-by: Hosgor, Tolga (CT RDA DS EU TR MTS) <tolga.hosgor@siemens.com>

### DIFF
--- a/vncdotool/command.py
+++ b/vncdotool/command.py
@@ -393,6 +393,9 @@ def vncdo():
         metavar='FACTOR', default=1.0,
         help='pause time is accelerated by FACTOR [x%default]')
 
+    op.add_option('--incremental-expect', action='store', type='int', metavar='INCREMENTAL_EXPECT', default='1',
+        help='try setting this to 0 if expect has problems matching the current screen [%default]')
+
     options, args = op.parse_args()
     if not len(args):
         op.error('no command provided')
@@ -417,6 +420,8 @@ def vncdo():
 
     if options.force_caps:
         factory.force_caps = True
+
+    factory.incremental_expect = options.incremental_expect
 
     if options.timeout:
         message = 'TIMEOUT Exceeded (%ss)' % options.timeout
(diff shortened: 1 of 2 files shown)


vncdotool/command.py
```diff
@@ -393,6 +393,9 @@ def vncdo():
         metavar='FACTOR', default=1.0,
         help='pause time is accelerated by FACTOR [x%default]')
 
+    op.add_option('--incremental-expect', action='store', type='int', metavar='INCREMENTAL_EXPECT', default='1',
+        help='try setting this to 0 if expect has problems matching the current screen [%default]')
```
sibson: Originally, I made incremental=1 the default because it seemed more compatible, https://github.com/sibson/vncdotool/commit/28c15b65305ecc010a2cd0b708667cb489e83568. This feels like its a leaking abstraction and exposing protocol level features in the UX.  Is there a better name we can use here to help users understand what the feature is and why they might use it?  ```--disable-incremental-refreshes``` is the first thing that comes to mind but I think it suffers the same problem and perhaps implies a wider scope of impact.

tghosgor: > This feels like its a leaking abstraction and exposing protocol level features in the UX. Is there a better name we can use here to help users understand what the feature is and why they might use it?

I guess you are right but this is *vnc*dotool and exposing vnc features should be fine for some amount?

> `--disable-incremental-refreshes` is the first thing that comes to mind but I think it suffers the same problem and perhaps implies a wider scope of impact.

I don't know, maybe `--expect-full-frames` or `--request-full-frames`? It seems a little bit hard to understand though?

sibson:
https://github.com/sibson/vncdotool/pull/142 is a partial duplicate of this pr.

Issue #147, opened 2019-03-20
Process not returning error on bad arguments


arossert:
When using the CLI it should return an error code on bad arguments but 0 is returned.

```
Amirs-MacBook-Pro-2:~ arossert$ vncdo -s 127.0.0.1::5900 bad cmd
unknown cmd "bad"
unknown cmd "cmd"
Amirs-MacBook-Pro-2:~ arossert$ echo $?
0
```

vncdotool version - 0.12.1

PR #148, opened 2019-03-20
write to stdout and and exit with error code on bad arguments

arossert:
Following #147, this fix will return an error code and print the output to STDERR on bad arguments.


sibson:
✨ 🍰 ✨ thanks

Issue #145, opened 2019-03-10
Capture is not working MacOS on 0.12.1


arossert:
I'm using vncdotool to control a Linux machine using Vmware Fusion, everything is working as expected except for the capture feature. 
When I submit a capture command I get a file with one pixel in it sized 69b, the command is:

`vncdo -s 127.0.0.1::5900 capture pic.png`

The file generated has one small pixel in it, if the screen is black then it is black and if the screen is blue than it is blue so it seems that the capture is kinda working.

Is there anything that needs to be configured to make it work?

**System information:**
vncdotool version - 0.12.1
Pillow version - 5.4.1
Vmware Fusion - 10.1.5
MacOS - 10.14.3 (Mojave)
Python version -  2.7.15 (tried with 3.6 with the same results)
VNC server - Built in Vmware Fusion

**Update:**
Tested on vncdotool version 0.11.2 and it seems to fix the issue (0.12.0 didn't work as well)

**Update2:**
I have looked at the code and it seems that the issue is with the screen size in this function:
```
def _captureSave(self, data, filename, *args):
  log.debug('captureSave %s', filename)
  if args:
      capture = self.screen.crop(args)
  else:
      capture = self.screen
  capture.save(filename)

  return self
```
When I debug it I see that `self.screen.size` is a tuple of `(1,1)` when it should be `(800,600)` I'm willing to provide a fix but I have no idea where to start.
On version 0.11.2 it works but it does not seem that the changes are related, maybe something with the server version.

arossert:
@sibson can you please address this issue? I need the capture/expect features to work so currently I'm reverting to version 0.11.2 but I want to use the latest version.

sibson:
There is little I can personally do, I don't have VMWare.  It sounds like the server is sending back an incremental update or a pseudo-update vncdo doesn't expect/recognize.  Doing something similar to this [PR](https://github.com/sibson/vncdotool/pull/142) forcing a full update might help.  Otherwise, running with ```--vvv``` and adding some logging to track what updateReactangle calls are being seen is the best way to debug the issue.

arossert:
Thanks for the fast replay, this is the output with `-vvv` on both versions:

V0.12.1(not working):
```
INFO:root:connecting to 127.0.0.1:5901
INFO:twisted:Starting factory <vncdotool.command.VNCDoCLIFactory instance at 0x108fcda28>
INFO:twisted:Using protocol version 3.800
INFO:root:connected to Windows 10 x64
[...]

PR #149, opened 2019-03-23
added the ability to control the "incremental" flag from command and api

arossert:
Following #145, this PR allows the user to control the "incremental" flag both in the command and API.
When setting the incremental flag to 1 it fixes the capture on the latest `0.12.1` version.

### DIFF
--- a/vncdotool/command.py
+++ b/vncdotool/command.py
@@ -113,7 +113,7 @@ class VNCDoToolOptionParser(optparse.OptionParser):
         return result
 
 
-def build_command_list(factory, args, delay=None, warp=1.0):
+def build_command_list(factory, args, delay=None, warp=1.0, incremental_refreshes=False):
     client = VNCDoCLIClient
 
     if delay:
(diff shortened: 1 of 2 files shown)


vncdotool/command.py
```diff
@@ -172,7 +172,7 @@ def build_command_list(factory, args, delay=None, warp=1.0):
                         imgformat, SUPPORTED_FORMATS), file=sys.stderr)
                 sys.exit(1)
             else:
-                factory.deferred.addCallback(client.captureScreen, filename)
+                factory.deferred.addCallback(client.captureScreen, filename, int(incremental_refreshes))
```
sibson: The type conversion here feels a bit strange.  Can we avoid this somehow, do it earlier, later or trust duck typing to do the right thing?

vncdotool/command.py
```diff
@@ -398,6 +398,9 @@ def vncdo():
         metavar='FACTOR', default=1.0,
         help='pause time is accelerated by FACTOR [x%default]')
 
+    op.add_option('-i', '--incremental-refreshes', action='store_true', default=False,
+        help='set the "incremental" flag')
```
sibson: I always struggle with these flag names.  I'd like to make them more approachable to users and hide VNC protocol details but that can also make it harder to understand.  I guess my real question is what is the correct default, what is the most compatible setting we can use that works for most people out of the box?

arossert: The current default value is 0 (False) so I kept it the same.
Regarding the flag name, I think that it needs to give enough information on what it is used for, I can change it if you have a better idea.

vncdotool/command.py
arossert: The function expecting an int value and the flag is boolean, do you think that we need to convert it before?

sibson:
thanks for putting this together.

Issue #140, opened 2019-01-14
vnclog output is 0 bytes


sibeshkar:
Have you included the following information?

 - [x] VNC server and version

Xvnc TigerVNC version 1.8.0

 - [x] vncdotool version

vncdotool 0.12.1

 - [x] steps to reproduce
1. Run this docker command to start vnc server in ubuntu docker container with unity environment.
```
sudo docker run -itd -p 84:6080 -p 92:5901 -p 95:95 -e PASSWORD=password -e SUDO=yes chenjr0719/ubuntu-unity-novnc
```
2. Exec into docker container, `sudo apt-get update`, `sudo apt-get install python-pip`
3. Install vncdotool in container: `sudo pip install vncdotool`
4. Run this in container : 
```
vnclog -s 127.0.0.1::5901 -p boxware --password-required --listen 95 -v keylog.vdo
```
5. Outside the container, open VNC viewer and connect to `127.0.0.1::95` with password : 'password'

Output of running vnlog loop: 
```
INFO:twisted:VNCLoggingServerFactory starting on 95
INFO:twisted:Starting factory <vncdotool.loggingproxy.VNCLoggingServerFactory instance at 0x7fa45413de60>
INFO:proxy:new connection from 172.17.0.1
INFO:twisted:Starting factory <vncdotool.loggingproxy.VNCLoggingClientFactory instance at 0x7fa454143e18>
```
6. [do things with keyboard and mouse]
7. Go inside container, do `cat keylog.vdo`
8. `keylog.vdo` is zero bytes, nothing is logged.

sibson:
Have you terminated the connection or vnclog process? If not then my guess is the OS output buffer isn't being flushed before you run cat, in which case running [sync](https://linux.die.net/man/8/sync) before ```cat``` should solve the issue.

sibeshkar:
Still no luck. Have tried terminating both connection and vnclog process before running `cat keylog.vdo`. Running `sync` during a connection doesn't help either. Was it possible for you to reproduce this error? Is there some specific way to debug this?

Paulynz:
Good day! Any update of this? vnclog not logging at all.

PR #153, opened 2019-06-24
Added support for QEMU Client Messages

DarkMaguz:
I'm currently in need of a non-proprietary tool to help me build and run CI smoke tests for a project.
This tool seem perfect for the task, but it is missing support for QEMU Client Messages as described in the [RFB protocol](https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#client-to-server-messages).

This should also solve issue [#140](https://github.com/sibson/vncdotool/issues/140).

### DIFF
--- a/vncdotool/loggingproxy.py
+++ b/vncdotool/loggingproxy.py
@@ -38,7 +38,7 @@ class RFBServer(Protocol):
         self._handler = self._handle_version, 12
 
     def dataReceived(self, data):
-        self.buffer = data + self.buffer
+        self.buffer = self.buffer + data
         while len(self.buffer) >= self._handler[1]:
             self._handler[0]()
 
@@ -60,13 +60,9 @@ class RFBServer(Protocol):
             self._handler = self._handle_security, 2
 
     def _handle_security(self):
-        sectype = unpack('!H', self.buffer[:2])[0]
+        sectype, shared = unpack('!BB', self.buffer[:2])
         self.buffer = self.buffer[2:]
-        if sectype != 1:
-            self.transport.loseConnection()
-            print("sectype: ", sectype)
-            raise "Security type not implanted!"
-        else:
+        if sectype == 1:
             self._handler = self._handle_protocol, 1
 
     def _handle_VNCAuthResponse(self):
--- a/vncdotool/loggingproxy.py
+++ b/vncdotool/loggingproxy.py
@@ -19,6 +19,7 @@ TYPE_LEN = {
     3: 10,
     4: 8,
     5: 6,
+    255: 12
 }
 
 REVERSE_MAP = dict((v, n) for (n, v) in KEYMAP.items())
@@ -31,19 +32,20 @@ class RFBServer(Protocol):
         Protocol.connectionMade(self)
         self.transport.setTcpNoDelay(True)
 
-        self.buffer = ''
+        self.buffer = b''
         self.nbytes = 0
         # XXX send version message
         self._handler = self._handle_version, 12
 
     def dataReceived(self, data):
-        self.buffer += data
+        self.buffer = data + self.buffer
         while len(self.buffer) >= self._handler[1]:
             self._handler[0]()
 
     def _handle_version(self):
-        msg = self.buffer[:12]
+        msg = self.buffer[:12].decode('utf-8')
         self.buffer = self.buffer[12:]
+
         if not msg.startswith('RFB 003.') and msg.endswith('\n'):
             self.transport.loseConnection()
 
@@ -55,11 +57,17 @@ class RFBServer(Protocol):
                 self._handler = self._handle_clientInit, 1
         elif version in ('007', '008'):
             # XXX send security v3.7+
-            self._handler = self._handle_security, 1
+            self._handler = self._handle_security, 2
 
     def _handle_security(self):
-        sectype = self.buffer[0]
-        self.buffer = self.buffer[1:]
+        sectype = unpack('!H', self.buffer[:2])[0]
+        self.buffer = self.buffer[2:]
+        if sectype != 1:
+            self.transport.loseConnection()
+            print("sectype: ", sectype)
+            raise "Security type not implanted!"
+        else:
+            self._handler = self._handle_protocol, 1
 
     def _handle_VNCAuthResponse(self):
         self.buffer = self.buffer[16:]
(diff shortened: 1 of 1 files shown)


vncdotool/loggingproxy.py
```diff
@@ -31,19 +32,20 @@ def connectionMade(self):
         Protocol.connectionMade(self)
         self.transport.setTcpNoDelay(True)
 
-        self.buffer = ''
+        self.buffer = b''
```
agfor: Did you test this on both Python 2 and 3?

vncdotool/loggingproxy.py
```diff
@@ -31,19 +32,20 @@ def connectionMade(self):
...
         self.nbytes = 0
         # XXX send version message
         self._handler = self._handle_version, 12
 
     def dataReceived(self, data):
-        self.buffer += data
+        self.buffer = data + self.buffer
```
agfor: These aren't equivalent? I'm not sure how this worked at all unless the buffer was always empty when you received new data, as they'll put it in the wrong order?

vncdotool/loggingproxy.py
```diff
@@ -55,11 +57,17 @@ def _handle_version(self):
...
 
     def _handle_security(self):
-        sectype = self.buffer[0]
-        self.buffer = self.buffer[1:]
+        sectype = unpack('!H', self.buffer[:2])[0]
+        self.buffer = self.buffer[2:]
+        if sectype != 1:
+            self.transport.loseConnection()
+            print("sectype: ", sectype)
+            raise "Security type not implanted!"
```
agfor: IIRC there was at least one case where just ignoring the security type worked, so this may break some previously working setups?

vncdotool/loggingproxy.py
```diff
@@ -55,11 +57,17 @@ def _handle_version(self):
                 self._handler = self._handle_clientInit, 1
         elif version in ('007', '008'):
             # XXX send security v3.7+
-            self._handler = self._handle_security, 1
+            self._handler = self._handle_security, 2
```
agfor: How did this work before if it was chopping off the wrong number of bytes? Wouldn't it break on the next message? Is 2 definitely correct?

vncdotool/loggingproxy.py
DarkMaguz: I have tested with version 2.7.16 and 3.7.3

vncdotool/loggingproxy.py
DarkMaguz: That was a mistake, i'm glad you spotted it.
Must have been an empty buffer as you point out.

vncdotool/loggingproxy.py
DarkMaguz: Yeah, that looks like Monday coding for me!
That if statement should look like this:
```suggestion
            if sectype == 0:
                self._handler = self._handle_protocol, 1
```
Maybe a warning would be appropriate though?

vncdotool/loggingproxy.py
DarkMaguz: The first two questions you asked here i can answer with ease...
> How did this work before if it was chopping off the wrong number of bytes? Wouldn't it break on the next message?
`_handle_security` did not do any thing with the data before. If `vnclog` connected to any server with RFB version 3.7 or 3.8, nothing would ever happen because they havn't been implemented in the logging proxy.

The 3rd question however is a bit more complicated.
The proxy server does not receive all the data packages and thus it does not follow the RFB protocol completely, nor should it. It's just there to capture data and should not respond to it.

The RFB protocol requires the client and server to negotiate for a [SecurityType](https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#security). This is done (in version 3.7+) by sending 1 byte informing the client that X number of SecurityTypes is supported by the server followed by X bytes describing the supported SecurityTypes. But all of this has all ready been handled by the client.

The first byte is the SecurityType that the client has chosen.
The second byte is the shared-flag and should be handled by [ClientInit](https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#clientinit), but I have not been able to find an easy solution to pass it to `_handle_clientInit` without getting an error: `builtins.AttributeError: 'NullTransport' object has no attribute 'addressFamily'`.

vncdotool/loggingproxy.py
DarkMaguz: Do note that compound assignment of bytes objects is not possible in python 3.

vncdotool/loggingproxy.py
```diff
@@ -101,6 +105,10 @@ def _handle_protocol(self):
             self.handle_pointerEvent(x, y, buttonmask)
         elif ptype == 6:
             self.handle_clientCutText(block)
+        elif ptype == 255:
+            # https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#qemu-client-message
+            downFlag, keySym = unpack('!xHIxxxx', block)
```
sibson: should we have a new handle_QEMU() function here rather than coding the behaviour inline? What about the audio message sub-type?

sibson (changes requested):
https://github.com/sibson/vncdotool/pull/153#discussion_r315660248

vncdotool/loggingproxy.py
pmhahn: See 6c7374c from my #232 for an alternativ implementation

vncdotool/loggingproxy.py
pmhahn: Fixed by fdb1bf5d

vncdotool/loggingproxy.py
pmhahn: Changed by 756c5f89

vncdotool/loggingproxy.py
```diff
@@ -31,19 +32,20 @@ def connectionMade(self):
...
 
     def dataReceived(self, data):
-        self.buffer += data
+        self.buffer = self.buffer + data
         while len(self.buffer) >= self._handler[1]:
             self._handler[0]()
 
     def _handle_version(self):
-        msg = self.buffer[:12]
+        msg = self.buffer[:12].decode('utf-8')
```
pmhahn, line 46: A hostile server may send a non-UTF-8 sequence and your client will crash than. Instead of converting `bytes` to `str` compare as `bytes` directly as already implemented in c1b89106

pmhahn (changes requested):
Probably obsoleted by my PR #232 
This PR alone also is not sufficient as
- `PSEUDO_QEMU_EXTENDED_KEY_EVENT` must be sent from client to server via `setEncodings()`
- `PSEUDO_QEMU_EXTENDED_KEY_EVENT` must be handled in `_handleRectangle()` for the response from the server to the client
Only then the client is allowed to use the extension.

PR #156, opened 2019-09-24
in my env,can not type &,*,^.....etc

cqb98:
In my computer,can not type '&','*','^'.....etc.
So i modify the code.It can run on my comupter


sibson:
Did you try the force-caps option? I believe it should give the similar behaviour.

Issue #54, opened 2015-12-03
MouseDrag - Pointer not drawn on screen


pevogam:
Hi Marc,

Do you know if the cursor is supposed to be drawn during a mouse drag operation? Currently, using Fedora virtual machine and a Qemu hypervisor (for instance but also noticed on other OS vms) the mouse is not drawn during the drag calculations with any step size and extra timeouts until it reaches the last position when it is updated. Is there a way to force the cursor to be drawn on screen?

Thanks a lot,

sibson:
Increasing the [sleep here](https://github.com/sibson/vncdotool/blob/master/vncdotool/client.py#L324) might help and if so it could be exposed as a command line arg rather than hardcoded.

pevogam:
This is what I did - I increased this timeout to the daring 1.2 seconds but it didn't help. However, the cursor is drawn/restored in the end of the MouseDrag when we arrive at the destination position. I also tried reducing the step size to 1 pixel and increasing the timeout to 1.2 seconds but the mouse remains invisible until it reaches the end destination (which also lasts a lot for this step size and timeout among steps).

NethajiDuvarakanathan:
Hi Pevogam,

I am glad to be here in this conversation.
I am trying to drag the mouse using vncdo -s serveripaddress drag xy
But nothing is happening. 
Please could anyone help me in this.

sibson:
I would expect the cursor to be drawn on a physical host display but perhaps not on any vnc or remote displays.  Depending on how the vnc server and clients are implemented they might be busy processing the setCursor commands rather than sending screen or cursor refreshes.

Making the cursor visible was outside the design scope of the drag command.  I'm curious what use case you have that requires it to be visible thoughout the operation?

NethajiDuvarakanathan:
Thanks for the reply.
 Here I would like to elaborate the info from your reply statements:
"I would expect the cursor to be drawn on a physical host display but perhaps not on any vnc or remote displays "
detailed info : 
1. I hve installed VNC server in my mobile.
2. via VNCDOTOOL commands am trying to initiate commands  & trying to control mobile icons / apps [ eg: vncdo -s serveripaddress move x y click 1
vncdo -s serveripaddress key a

"Making the cursor visible was outside the design scope of the drag command. I'm curious what use case you have that requires it to be visible thoughout the operation? "

1. Here my intention is not to make cursor visible.
[...]

PR #163, opened 2020-01-30
Force reactor to poll socket when dragging

samiraguiar:
The `mouseDrag()` method is called during the reactor loop and all
of its calls to `mouseMove()` will be buffered as the reactor
won't check the underlying socket while that method is running.
That means that all the move operations will be sent at once to the
VNC server *after* `mouseDrag()` has executed, causing the cursor
to jump straight to the final location. The `sleep()` calls in the
drag method should prevent this, but since they are not yielded to
the reactor they have no effect.

A simple solution is to call `reactor.doPoll()` after each
`mouseMove()` to make sure that the socket is polled by the reactor
and that the previous move command is sent to the VNC Server.

This fixes issue #54.

### DIFF
--- a/vncdotool/client.py
+++ b/vncdotool/client.py
@@ -330,12 +330,14 @@ class VNCDoToolClient(rfb.RFBClient):
             ysteps = range(self.y, y, step)
 
         for ypos in ysteps:
-            time.sleep(.2)
             self.mouseMove(self.x, ypos)
+            reactor.doPoll(timeout=5)
+            time.sleep(.2)
 
         for xpos in xsteps:
-            time.sleep(.2)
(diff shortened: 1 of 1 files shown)


samiraguiar:
@sibson the sleep() calls are still needed, otherwise the cursor will still jump to the final location... we need to buffer the move operation, poll the socket for it to be executed, wait a bit and repeat.

sibson:
ok, I wonder if a chain of [reactor.callLater()](https://twistedmatrix.com/documents/10.1.0/core/howto/time.html) calls would avoid the need for doPoll() and sleep().  I'm happy to accept this as is, to get us into a better place.  Thanks for figuring this out, its been a long running bug.

PR #164, opened 2020-02-17
Fix HEXTILE encoding under Python3

c001c0de:
In Python 3, data is represented as `bytes`, not as `str`. So, things like `ord(block[pos])` won't work in Python3. This is used extensively in decoding Hextile.

I have added a replacement for `ord()` which is activated only on Python3 (i.e. when `b' ' ` is not a `str`) which returns the argument unchanged.

### DIFF
--- a/vncdotool/rfb.py
+++ b/vncdotool/rfb.py
@@ -24,6 +24,9 @@ from twisted.application import internet, service
 
 #~ from twisted.internet import reactor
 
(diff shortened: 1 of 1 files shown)


sibson:
thanks, ✨ 🍰 ✨

guicho271828:
This is a wrong fix causing #170 and #172

PR #165, opened 2020-02-17
Fix missing self argument in updateDesktopSize

c001c0de:
Minor fix: `updateDesktopSize()` is missing the `self` argument


sibson:
thanks, ✨ 🍰 ✨

PR #166, opened 2020-02-17
ZRLE implementation

c001c0de:
I have implemented ZRLE encoding. Since this is the most efficient and bandwith saving encoding of all available encodings so far, I would suggest making this the default.

This was originally written under Python 3. I have packported to Python2 using the `ord()` compatibility hack previously used to make hextile encoding work under Python3.

### DIFF
--- a/vncdotool/client.py
+++ b/vncdotool/client.py
@@ -124,7 +124,7 @@ class AuthenticationError(Exception):
 
 
 class VNCDoToolClient(rfb.RFBClient):
-    encoding = rfb.ZRLE_ENCODING
+    encoding = rfb.RAW_ENCODING
     x = 0
     y = 0
     buttons = 0
--- a/vncdotool/client.py
+++ b/vncdotool/client.py
@@ -124,7 +124,7 @@ class AuthenticationError(Exception):
 
 
 class VNCDoToolClient(rfb.RFBClient):
-    encoding = rfb.RAW_ENCODING
+    encoding = rfb.ZRLE_ENCODING
     x = 0
     y = 0
     buttons = 0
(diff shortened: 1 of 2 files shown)


vncdotool/rfb.py
```diff
@@ -112,6 +116,49 @@
 KEY_SpaceBar=   0x0020
 
 
+# ZRLE helpers
+def _next_bit(it, pixels_in_tile):
```
sibson: minor nit, should these have shared ```_zrle``` prefix to more easily identity them?

vncdotool/client.py
```diff
@@ -124,7 +124,7 @@ class AuthenticationError(Exception):
 
 
 class VNCDoToolClient(rfb.RFBClient):
-    encoding = rfb.RAW_ENCODING
+    encoding = rfb.ZRLE_ENCODING
```
sibson: I'm on the fence about changing the default.  I tend to agree with your logic but I'm not sure what general compatibility of servers is with ZRLE and worry about breaking existing users.  I see a couple of options I outline below and am open to others. 

The quickest path to merging this PR would be to keep the default as RAW.  You could then expose a command line option to control the selection of encoding in a follow up PR.  Finally, in a major version bump we can change the default from RAW to ZRLE.  

Another option is I hold this PR up until a major version bump is approaching and then add the CLI switch defaulting to ZRLE.

c001c0de: I have reverted the change of the default encoding

vncdotool/rfb.py
c001c0de: done

sibson:
Thank you, this is great.

PR #169, opened 2020-03-25
Add missing ZRLE case "raw pixel data"

c001c0de:
The case of subencoding = 0 (raw pixel data) described in https://tools.ietf.org/html/rfc6143#section-7.7.6 at the end of page 26  escaped my attention.

I have added it now.


sibson:
thanks, ✨ 🍰 ✨

Issue #172, opened 2020-04-20
CRITICAL:root:unsupported operand type(s) for &: 'str' and 'int'


nandish017:
Have you included the following information?

 -  Tight VNC 2.8.27
 -  Latest vncdotool
 - Installed vncdotool by following the steps mentioned in https://vncdotool.readthedocs.io/en/latest/install.html#windows
I am able to connect to the mentioned IP, when i give for the password for vncpassword I get the following error:

CRITICAL:twisted:Unhandled Error
Traceback (most recent call last):
  File "c:\users\appdata\local\programs\python\python3.8\lib\site-packages\twisted\python\log.py", line 103, in callWithLogger
    return callWithContext({"system": lp}, func, *args, **kw)
  File "c:\users\appdata\local\programs\python\python3.8\lib\site-packages\twisted\python\log.py", line 86, in callWithContext
    return context.call({ILogContext: newCtx}, func, *args, **kw)
  File "c:\users\appdata\local\programs\python\python3.8\lib\site-packages\twisted\python\context.py", line 122, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "c:\users\appdata\local\programs\python\python3.8\lib\site-packages\twisted\python\context.py", line 85, in callWithContext
    return func(*args,**kw)
--- <exception caught here> ---
  File "c:\users\appdata\local\programs\python\python3.8\lib\site-packages\twisted\internet\selectreactor.py", line 149, in _doReadOrWrite
    why = getattr(selectable, method)()
  File "c:\users\appdata\local\programs\python\python3.8\lib\site-packages\twisted\internet\tcp.py", line 243, in doRead
    return self._dataReceived(data)
  File "c:\users\appdata\local\programs\python\python3.8\lib\site-packages\twisted\internet\tcp.py", line 249, in _dataReceived
    rval = self.protocol.dataReceived(data)
  File "c:\users\appdata\local\programs\python\python3.8\lib\site-packages\vncdotool\rfb.py", line 685, in dataReceived
    self._handler()
  File "c:\users\appdata\local\programs\python\python3.8\lib\site-packages\vncdotool\rfb.py", line 694, in _handleExpected
    self._expected_handler(block, *self._expected_args, **self._expected_kwargs)
  File "c:\users\appdata\local\programs\python\python3.8\lib\site-packages\vncdotool\rfb.py", line 255, in _handleVNCAuth
    self.vncRequestPassword()
[...]

PR #177, opened 2020-05-27
test with xvnc

ponty:
This test can catch bug #172 with travis or tox.

Error message:
`builtins.TypeError: unsupported operand type(s) for &: 'str' and 'int'`

https://travis-ci.org/github/ponty/vncdotool/jobs/691896577


sibson (approved):
This looks great.  If I'm understanding correctly this is more of a functional test than a unit test.  If so can we move to tests/functional to ensure we keep the unit tests running quickly without external dependencies.

Issue #170, opened 2020-04-16
unsupported operand type(s) line 835 in rfb.py with Python 3.6


AlainB612:
Hi,
I've just installed the new version of vncdotool and get this error when trying to capture screen :
```
  File "/usr/local/lib/python3.6/site-packages/vncdotool/rfb.py", line 835, in setKey
    if bsrc & (1 << i):
builtins.TypeError: unsupported operand type(s) for &: 'str' and 'int'

```
The complete listing is at the end of this message.

_04/17/2020 EDIT : I remove vncdotool 1.0.0 and rebuild 0.12 version : it works fine. I retry with 1.0.0 (from pip), I get the error._

 Thanks for your help.
Alain

 - **VNC server and version** : UltraVNC 1.2.4.0 on Windows Server 2016 64bits
 - **vncdotool version** : vncdotool 1.0.0 (installed by PIP)
 - **Python version** : 3.6
 - **OS** : Linux CentOS 7 up to date
 - **steps to reproduce** : command = vncdotool -p pwd -s 1.2.3.4::5900 -v capture capture.png 2>&1

```
# vncdotool -p pwd -s 1.2.3.4::5900 -v capture capture.png 2>&1
INFO:root:connecting to 1.2.3.4:5900
INFO:twisted:Starting factory <vncdotool.command.VNCDoCLIFactory object at 0x7f769c669898>
INFO:twisted:Using protocol version 3.800
Unhandled Error
Traceback (most recent call last):
  File "/usr/local/lib64/python3.6/site-packages/twisted/python/log.py", line 103, in callWithLogger
    return callWithContext({"system": lp}, func, *args, **kw)
  File "/usr/local/lib64/python3.6/site-packages/twisted/python/log.py", line 86, in callWithContext
    return context.call({ILogContext: newCtx}, func, *args, **kw)
  File "/usr/local/lib64/python3.6/site-packages/twisted/python/context.py", line 122, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/usr/local/lib64/python3.6/site-packages/twisted/python/context.py", line 85, in callWithContext
    return func(*args,**kw)
--- <exception caught here> ---
  File "/usr/local/lib64/python3.6/site-packages/twisted/internet/posixbase.py", line 614, in _doReadOrWrite
    why = selectable.doRead()
  File "/usr/local/lib64/python3.6/site-packages/twisted/internet/tcp.py", line 243, in doRead
[...]

PR #178, opened 2020-07-07
Fix a wrong fix to `ord`

guicho271828:
Due to python3's incompatible change, data buffer is represented as a `byte` rather than `str` .
The RFB code depends on the assumption that they are str, which no longer works.
The code uses `ord` to convert a single-char string into an `int`.
There are two typical usages of `ord` in the code:

* buffer string of length > 1 --(index access)--> string of length 1 --(ord)--> int
* string of length 1 --(ord)--> int

However in python3, this usage morphs into

* byte of length > 1, --(index access)--> int --(ord)--> Error
* byte of length 1 --(ord)--> int

The previous wrong fix #164 only considers the first case, and replace `ord` with an identity that returns an int.
However this fix causes an error (#170 #172).
I reimplemented ord simply as a cast : When the input is an int, returns it, otherwise apply the normal `ord`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@ vncdotool.egg-info
 .noseids
 .tox
 docs/_build
+
+*~
+#*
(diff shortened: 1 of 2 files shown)


sibson:
@guicho271828 thanks for debugging and putting a fix together

Issue #150, opened 2019-04-08
CRITICAL:root:Connection was refused by other side: 111: Connection refused


r4venlord:
I have a problem with vncdotool both from python script and from linux CLI:

vncdotool -v -s 127.0.0.1:5902 capture test.png
INFO:root:connecting to 127.0.0.1:11802
INFO:twisted:Starting factory <vncdotool.command.VNCDoCLIFactory object at 0x7fd05916c160>
CRITICAL:root:Connection was refused by other side: 111: Connection refused.
INFO:twisted:Stopping factory <vncdotool.command.VNCDoCLIFactory object at 0x7fd05916c160>
INFO:twisted:Main loop terminated.

-----------

And also from python script:

#!/usr/bin/python3
from vncdotool import api

client = api.connect('127.0.0.1:5902', password="xxx")
client.captureScreen('screenshot.png')
client.disconnect()

./vnctool.py
Traceback (most recent call last):
  File "./vnctool.py", line 6, in <module>
    client.captureScreen('screenshot.png')
  File "/home/user/.local/lib/python3.5/site-packages/vncdotool/api.py", line 108, in proxy_call
    raise VNCDoException(result)
vncdotool.api.VNCDoException: [Failure instance: Traceback (failure with no frames): <class 'twisted.internet.error.ConnectionRefusedError'>: Connection was refused by other side: 111: Connection refused.
]

------------

Have  tried to find someone else with the same issue but haven't found any yet.

I have tried to tcpdump and use vncdotool from another server and when looking into tcpdump i do not recieve any connection on specified port.

It works to telnet from internal & external:

telnet localhost 5902
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
RFB 003.008

 Have you included the following information?

 - Xvnc version TightVNC-1.3.9
 - vncdotool 0.12.1

Hope that someone will find out a solution.

alex2304:
Faced almost the same issue.

I "solved" it in the following way:

1) Connect to the server via VNC Viewer (or may be some other client also can suit, but I used this one)
2) In the connection properties, choose <b>Medium or High quality of image - it's important!</b> Connect with new settings.
3) Close connection of VNC Viewer, and try to connect using VNCDoTool. In my case, after that strange manipulation it works.

I suggest that some variable related to image quality is needed to be sent to the server. 
Started to read RFB specification.

Bnaya:
I'm having the same issue using the CLI
My vnc server is the one on selenium docker
[...]

PR #180, opened 2020-10-02
Exemplify the many ways of specifying the server

samiraguiar:
The format for the server address is not clearly specified in the
documentation, and this has led to users having problems such as
issue #150.


sibson:
thanks, ✨ 🍰 ✨

Issue #183, opened 2020-12-22
builtins.ValueError: Invalid initialization option


marcobazzani:
Have you included the following information?

 - [x] VNC server and version
Xvnc TigerVNC 1.11.0 - built Sep  8 2020 12:27:03
Copyright (C) 1999-2020 TigerVNC Team and many others (see README.rst)
See https://www.tigervnc.org for information on TigerVNC.
Underlying X server release 12001000, The X.Org Foundation

 - [x] vncdotool version
1.0.0
 - [x] steps to reproduce
vncdo -v -s localhost::39737 -p xxxx type "ls"

output:

```
INFO:root:connecting to localhost:39737
INFO:twisted:Starting factory <vncdotool.command.VNCDoCLIFactory object at 0x7f2bbc2ac668>
Unhandled Error
Traceback (most recent call last):
  File "/home/mbazzan/.pyenv/versions/3.6.8/lib/python3.6/site-packages/twisted/python/log.py", line 103, in callWithLogger
    return callWithContext({"system": lp}, func, *args, **kw)
  File "/home/mbazzan/.pyenv/versions/3.6.8/lib/python3.6/site-packages/twisted/python/log.py", line 86, in callWithContext
    return context.call({ILogContext: newCtx}, func, *args, **kw)
  File "/home/mbazzan/.pyenv/versions/3.6.8/lib/python3.6/site-packages/twisted/python/context.py", line 122, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/home/mbazzan/.pyenv/versions/3.6.8/lib/python3.6/site-packages/twisted/python/context.py", line 85, in callWithContext
    return func(*args,**kw)
--- <exception caught here> ---
  File "/home/mbazzan/.pyenv/versions/3.6.8/lib/python3.6/site-packages/twisted/internet/posixbase.py", line 619, in _doReadOrWrite
    why = selectable.doWrite()
  File "/home/mbazzan/.pyenv/versions/3.6.8/lib/python3.6/site-packages/twisted/internet/tcp.py", line 627, in doConnect
    self._connectDone()
  File "/home/mbazzan/.pyenv/versions/3.6.8/lib/python3.6/site-packages/twisted/internet/tcp.py", line 641, in _connectDone
    self.protocol = self.connector.buildProtocol(self.getPeer())
[...]

PR #185, opened 2020-12-23
Fix for zlib < 1.2.3.5 

marcobazzani:
as per  https://docs.python.org/3/library/zlib.html#zlib.decompress
zlib.decompressobj(0) is only supported with zlib < 1.2.3.5


marcobazzani (commented):
https://github.com/sibson/vncdotool/issues/183

marcobazzani:
see https://github.com/sibson/vncdotool/pull/186

PR #186, opened 2020-12-23
https://github.com/sibson/vncdotool/issues/183

marcobazzani:
fix for https://github.com/sibson/vncdotool/issues/183

### DIFF
--- a/vncdotool/rfb.py
+++ b/vncdotool/rfb.py
@@ -13,6 +13,7 @@ MIT License
 """
 # flake8: noqa
 
+import builtins
 import sys
 import math
 import zlib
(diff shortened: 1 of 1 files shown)


vncdotool/rfb.py
```diff
@@ -181,7 +182,10 @@ def __init__(self):
         self._already_expecting = 0
         self._version = None
         self._version_server = None
-        self._zlib_stream = zlib.decompressobj(0)
+        try: # Automatically determine the window size from the zlib header. Only supported since zlib 1.2.3.5.
+            self._zlib_stream = zlib.decompressobj(0)
+        except builtins.ValueError:
```
sibson, line 187: Can we simple catch ValueError rather than needing to import the builtin's namespace?

sibson (approved):
I'll give a few days to address the import question then merge.

vncdotool/rfb.py
marcobazzani: I'll be unable to fix it until Jan the 11th if it's ok.

sibson: no problem, appreciate the help

sibson:
circling back on this, I read this as we require zlib >= 1.2.3.5 to work.  1.2.3.5 was released 2010-01-08, and is over 12 years old.  Most LTS OS versions top out at 10 years of support, so I don't expect to see zlib < 1.2.3.5 in the wild.  Unless, my reasoning is incorrect or there is another reason to support an outdated version of zlib I'm inclined to close this without merging because I don't see enough benefit to warrant the extra code complexity for future maintenance.

sibson (commented):
Per my other comment, I'd like to ensure we really need to support older zlib versions

PR #189, opened 2021-06-03
add errback for clientConnectionLost

dozysun:
Fix client operation timeout when vnc password is wrong,  twisted will call clientConnectionLost and it don't put any result in queue will lead self.queue.get timeout, add errback for clientConnectionLost.


sibson:
It would be nice to have documentation about how to catch the password error, it feels like a common case or perhaps even have special handling somewhere.

sibson:
_
thanks, ✨ 🍰 ✨_

Issue #198, opened 2021-09-11
vnclog Unhandled Error: builtins.TypeError: can only concatenate str (not "bytes") to str


joachimmetz:
Thx for the making this interesting tool available. I was just playing around with it, so nothing urgent.

### Description:

vnclog raises `Unhandled Error: builtins.TypeError: can only concatenate str (not "bytes") to str` see traceback below

### Requested information

 Have you included the following information?

 - [x] VNC server and version
 - [x] vncdotool version
 - [x] steps to reproduce
 - [x] traceback

### VNC server and version

QEMU built-in VNC server, listening on 127.0.0.1:5901
note that a SPICE sever is listening on 127.0.0.1:5900

### vncdotool version

1.0.0

### steps to reproduce

Run `vnclog --forever --listen 6000 /tmp`

### Traceback

```
Unhandled Error
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/twisted/python/log.py", line 101, in callWithLogger
    return callWithContext({"system": lp}, func, *args, **kw)
  File "/usr/lib/python3.9/site-packages/twisted/python/log.py", line 85, in callWithContext
    return context.call({ILogContext: newCtx}, func, *args, **kw)
  File "/usr/lib/python3.9/site-packages/twisted/python/context.py", line 118, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/usr/lib/python3.9/site-packages/twisted/python/context.py", line 83, in callWithContext
    return func(*args, **kw)
--- <exception caught here> ---
  File "/usr/lib/python3.9/site-packages/twisted/internet/posixbase.py", line 687, in _doReadOrWrite
    why = selectable.doRead()
  File "/usr/lib/python3.9/site-packages/twisted/internet/tcp.py", line 246, in doRead
    return self._dataReceived(data)
  File "/usr/lib/python3.9/site-packages/twisted/internet/tcp.py", line 251, in _dataReceived
    rval = self.protocol.dataReceived(data)
  File "vncdotool-1.0.0/vncdotool/loggingproxy.py", line 202, in dataReceived
    RFBServer.dataReceived(self, data)
  File "vncdotool-1.0.0/vncdotool/loggingproxy.py", line 40, in dataReceived
    self.buffer += data
builtins.TypeError: can only concatenate str (not "bytes") to str

CRITICAL:twisted:Unhandled Error
Traceback (most recent call last):
[...]

PR #199, opened 2021-09-11
Changes to loggingproxy to use byte strings #198

joachimmetz:


sibson:
Are you using Python2 or 3?

joachimmetz:
I was using python 3.9 (also captured in the traceback of https://github.com/sibson/vncdotool/issues/198), these changes should work for python 2.7 as well (since b'' is equivalent to '' on that version)

sibson:
thanks, ✨ 🍰 ✨

PR #200, opened 2021-10-07
Add support for Apple Remote Desktop (ARD)

andywgrant:
This adds support for Apple Remote Desktop (ARD) in order to support Apple's screen sharing / remote management feature. This is done by adding support for a username parameter, and RFB protocol version 3.889 and auth type 30.

### DIFF
--- a/vncdotool/client.py
+++ b/vncdotool/client.py
@@ -464,6 +464,7 @@ class VMWareClient(VNCDoToolClient):
 
 
 class VNCDoToolFactory(rfb.RFBFactory):
+    username = None
     password = None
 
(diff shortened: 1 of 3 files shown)


sibson:
lgtm, thanks!  You may want to add some tests to help ensure this doesn't get accidentally broken in future.

pmhahn:
You forgot to add [PyCryptodome](https://pycryptodome.readthedocs.io/en/latest/src/introduction.html) also as a new dependency to `setup.py` as `pip install -e .` will not install it:
> E   ModuleNotFoundError: No module named 'Crypto'

PR #205, opened 2021-11-05
Enforce 16 BPP support

Kochise:
VNC servers delivering 16 BPP were not handled (black screen)

### DIFF
--- a/vncdotool/client.py
+++ b/vncdotool/client.py
@@ -346,7 +346,9 @@ class VNCDoToolClient(rfb.RFBClient):
     def setImageMode(self):
         """ Extracts color ordering and 24 vs. 32 bpp info out of the pixel format information
         """
-        if self._version_server == 3.889:
+        if (self.truecolor and (not self.bigendian) and self.depth == 16
+                and self.redmax == 31 and self.greenmax == 63 and self.bluemax == 31) \
+            or self._version_server == 3.889:
             self.setPixelFormat(
                     bpp = 16, depth = 16, bigendian = 0, truecolor = 1,
                     redmax = 31, greenmax = 63, bluemax = 31,


vncdotool/client.py
```diff
@@ -346,7 +346,9 @@ def mouseDrag(self, x, y, step=1):
     def setImageMode(self):
         """ Extracts color ordering and 24 vs. 32 bpp info out of the pixel format information
         """
-        if self._version_server == 3.889:
+        if (self.truecolor and (not self.bigendian) and self.depth == 16
+                and self.redmax == 31 and self.greenmax == 63 and self.bluemax == 31) \
+            or self._version_server == 3.889:
```
sibson, line 351: This likely gets the desired result. Simply, adding a complex or clause doesn't feel like the right approach.  There may even be a larger issue with how this method is implemented.  I suspect the else that casll setPixelFormat with default args rather than values from self is the culprit.

Kochise: Indeed. It wasn't working on 16 BPP, this way it does, but I like your idea, yet why hasn't it be done already that way ?

sibson: Introduced in https://github.com/sibson/vncdotool/commit/9bd3e24619a7464c4a3ecd013d8f43e17efc41a9.  Probably, best to properly support multiple depths and figure out a different way to handle OSX servers, if even still a problem.

pmhahn: Maybe I don't get this, but this PR looks strange:
1. The client has successfully connected the server and `client.vncConnectionMade()` has been called.
2. During the initial handshake the server has just previously sent its _native_ pixel format, which was handled by `rfb._handleServerInit()` and which has saved the values for us.
3. `vncConnectionMade()` now calls `setImageMode()` to check those values to see, if PIL does support that image format or not, in which case `rfb.setPixel()` gets called to
  - in the AppleRemoteDesktop-case `_version_server==3.889`  `BGR;16` is requested instead
  - for _TrueColor_ modes with 8 bits-per-pixel tell PIL how to decode the colors in the right order
  - otherwise select `RGBX` by default.

Now this PR adds a check, to handle _one_ case differently. For that let's compare the _server sent format_
> (self.truecolor and (not self.bigendian) and self.depth == 16 and self.redmax == 31 and self.greenmax == 63 and self.bluemax == 31)

with the then _requested format_
> bpp=16, depth=16, bigendian=0, truecolor=1, redmax=31, greenmax=63, bluemax=31, redshift=11, greenshift=5, blueshift=0

| name | server | req |
|------|--------|-----|
| bpp | :grey_question:  | 16 |
| depth | 16 | 16 |
| bigEndian | False | 0 |
| trueColor | True | 1 |
| redmax | 31 | 31 |
| greenmax | 63 | 63 |
| bluemax | 31 | 31 |
| redshift | :grey_question:  | 11 |
| greenshift | :grey_question:  | 5 |
| blueshift | :grey_question:  | 0 |

I see no (relevant) change and so calling `setPixelFormat(…)` for that case seems pointless. The only needed line for that case is
> self.image_mode = "BGR;16"

I already have a local branch, where I work on handling more image formats.

sibson:
Closing this PR as its not seen activity from the author in over a year and now has conflicts.

Kochise:
> Closing this PR as its not seen activity from the author in over a year and now has conflicts.

Because I don't work on it since nov 2021 and cannot modify/improve it anymore.

Good to know that @pmhahn seems to have taken over this PR and provided a more elaborate and universal solution.

For me, it was a quick hack to make it work. In my case.

Regards.

PR #207, opened 2021-12-22
add missing pycryptodome dependency to requirements.txt to pass build

DAcodedBEAT:
Hello 👋 

I was looking at making use of this project and I noticed that the build is broken because the code references modules which are not marked as dependencies, which [fails the build](https://github.com/sibson/vncdotool/runs/3854662017?check_suite_focus=true).

This PR is to outline this dependency in requirements.txt and ensure that the unit tests in the pipeline can run again.  Please let me know if there is anything else I need to do, as this is my first time contributing to this project.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 Pillow==6.2.0
 Twisted==21.7.0
 zope.interface==5.4.0
+pycryptodome==3.12.0


DAcodedBEAT:
Hi @sibson !  Would it be possible to get this PR reviewed (and hopefully merged)?

sibson:
I've not kept up with python crypto  and I've been meaning to look into what the status of pycrypto and pycryptodome.  Are there other forks of pycrypto that we could use?  I'm generally +1 on this change and simply cautious about introducing a new untrusted dependency.

DAcodedBEAT:
According to [issues on the `pycrypto`'s Github repository](https://github.com/pycrypto/pycrypto/issues/301), it sounds like pycrypto is left unmaintained, with the community recommending [`pycryptodome`](https://pypi.org/project/pycryptodome/) as an appropriate drop-in replacement.

From the other python projects I've looked at, I would've went with the [`cryptography`](https://pypi.org/project/cryptography/) package but that would require reimplementation of https://github.com/sibson/vncdotool/pull/200 (not sure if you or the author of that contribution @andywgrant required API compatibility with pycrypto).

Ideally this change should've been done in https://github.com/sibson/vncdotool/pull/200 to not break the build but such is life. (side note: it might be beneficial to include `pull_request` in the Github actions triggers in order to ensure that new contributions pass the build before merging)

sibson:
Thank you for the description, cryptography is the alternative I'm more familiar with.  I'll accept this now to get the build unblocked.  I'm open to converting to cryptography if that is desirable, but I don't have strong opinions.

PR #213, opened 2022-02-01
Use only logging for output to avoid unwanted such on stdout

pevogam:
There are two functions in the client module that output explicitly
to stdout instead of using the provided loggers. Some remote sessions
could break down because of some of this output and there isn't any
easy way to prevent the output from being emitted. As the client module
already makes full use of the logging levels and filters let's make
the following code respect that too.


sibson:
thanks, ✨ 🍰 ✨

Issue #227, opened 2022-12-20
Require Python 3.7+ / drop support for older versions


pmhahn:
`vncdotool` currently still supports Python 2.7, which is end-of-life since 3 years in a few months.
Also all versions of Python 3.0 up to and including to 3.6 are end-of-life.
Python 3.7 reaches end-of-life in half a year: 2023-06-27

See https://devguide.python.org/versions/

- [tox.ini](tox.ini) is still using `py35`
- [.travis.yml](.travis.yml) is still using `2.7` and `3.5`
- [setup.py](setup.py) ony lists `Programming Language :: Python :: 3` but no individual versions.
- [requirements-dev.txt](requirements-dev.txt) and [requirements.txt](requirements.txt) seem to lock some old versions; maybe by design to support some older versions, but updating them might be worhwile.

Because of the Python 2.7 support there is still some compatibility code left to handle the `bytes` / `unicode` differences, which could be removed if Python 2.7 support was dropped, especially all that `ord()` handling #219.  That would simplify some code paths and would also simplify adding PEP-484 type annotations (I have a local branch where I already started this.)

What are your plans regarding
- removing `tox.ini`
- replacing `nosetest`
- removing Travis-CI support
- removing Circle-CI support
- removing Python 3.6 and older support 
- adding type annotations

PS: Some _function_ tests use a tool called `vncev`, which I could not find. What is that and where can I get it?

sibson:
I think it's time to retire 2.7 support.  I'd kept it pass EOL as I knew there are users on platforms that are still using 2.7.  
 
  - tox.ini, not thought about.
  - I've started removing nosetests and non-std mock. 
  - Travis CI has got to go
  - CircleCI, could go.  I don't know if it'll be easier to make the intergration tests work on GH Action vs Circle.  That is probably the only reason to keep Circle around.
  -  All EOL python should be dropped
  - On type annontations, I'm not sure.  I'm not a fan but I've not used them.  I'd be open to looking at a PR and if its not too "noisy" include them.  If there is other value for static analisys or similar I could be convinced.

sibson:
vncev is an example program from https://github.com/LibVNC/libvncserver.  I use it for intergration testing, but never figured out a good way to build and test it using CI.  I usually do a manual run before packaging a release.

pmhahn:
Thank for the information. I have a local branch here which
- drops support for Python 2.7 and before 3.7
- adds type annotations (supported since Python 3.5 resp. Python 3.6)
[...]

PR #229, opened 2022-12-21
Move to Python 3.7+ code

pmhahn:
This addresses #227 to remove the old Python 2.7 compatibility code and moved the code base up to Python 3.7+
This is only lightly tested; `vnclog` still does not work.
Please have a look and review carefully; any (early) feedback from you is very much appreciated. Hopefully I will find time during the Christmas holidays to work on that.
If this is too much I can try to split this into separate PRs, but that is not trivial as some of those patches depend on each other: the code might not `git bisect` at all because some intermediate steps might be "broken".


tests/functional/test_screen.py
```diff
@@ -2,16 +2,21 @@
...
 import pexpect
+import pytest
 
 
 DATADIR = os.path.join(os.path.dirname(__file__), 'data')
 SIMPLE_PNG  = os.path.join(DATADIR, 'simple.png')
 EXAMPLE_PNG = os.path.join(DATADIR, 'example.png')
 EXAMPLE_NOCURSOR_PNG = os.path.join(DATADIR, 'example_nocursor.png')
 
+SERVER = "exmaple"
```
sibson, line 16: typo?

tests/unit/helpers.py
```diff
@@ -2,7 +2,7 @@
 from unittest.mock import Mock, _importer
 from functools import wraps
 
-class _isolate(object):
+class _isolate:
```
sibson, line 5: fyi, I'm working on removing all this isolate stuff

vncdotool/command.py
```diff
@@ -16,15 +15,20 @@
...
 from twisted.python.log import PythonLoggingObserver
 from twisted.internet import reactor, protocol
 from twisted.internet.error import ConnectionDone
+from twisted.internet.interfaces import IConnector
 from twisted.python.failure import Failure
 
 from .client import VNCDoToolFactory, VNCDoToolClient, factory_connect
 from .loggingproxy import VNCLoggingServerFactory
 
+V = TypeVar("V", bound="VNCDoCLIClient")
```
sibson, line 30: Is using V an common idiom?  What is this actually doing?

sibson (approved):
Generally LGTM and few clarifying questions.    

How have you tested this?  

I'm wondering if this should be a 2.0 type change.  I'm also thinking that when this goes in I'll close any open and stale PRs because most likely they'll need updating due to conflicts.  What do you think?

tests/unit/helpers.py
sibson: I think I'll rebase my work over this, as it also was doing the mock changes.

tests/functional/test_screen.py
pmhahn: Yes, will fix.
While at it maybe you can clarify how that  _functional_ unit test worked: It looks like you have an executable named `example` in your test  environment as this is `pexcept.spawn()`ed. Maybe it would be more useful to make that executable configurable, e.g. get it from some environment variable via `os.environ.get("VNCDOTOOL_TEST_SERVER", "example")`

vncdotool/command.py
pmhahn: `V`, `K`, `VT`, `VK` are often used for _unbound_ `TypeVar`s, but as now it is _bound_ `V` is no longer a good name. Thank you for spotting this. I changed it to `vncdotool.client.TClient`.

[TypeVar](https://mypy.readthedocs.io/en/stable/generics.html#generics)s are used to bind some type to a name: Take this simple function for an example: `def same(var: V) -> V: return var`
You can pass any type (int, str, object, ….) into that function and will get the same type back.
The additional `bound="VNCDoCLIClient"` adds one more restriction: the passed in value must be an instance of `VNCDoCLIClient` (or any subclass of it) and you will (still) get the same type back. This is important if you actually pass in a sub-class of `VNCDoCLIClient` as then you're guaranteed to get back that sub-class, and not only the super-class as you would get with `def f(var: VNCDoCLIClient) -> VNCDoCLIClient: ...`: There you could still pass in a sub-class of `VNCDoCLIClient`, but you're only guaranteed to get back a `VNCDoCLIClient`, which might then not have the added methods of your sub-class.

pmhahn:
> How have you tested this?

You got me: I'm still on it which is why I called this PR a `WIP`. I only did some simple `vncdotool` testing so far.
My company [Univention GmbH](https://www.univention.de/) is using it for our automated testing, which is why I'm working on this at all. Our tests where failing after switching to Python 3 so I had to go down very deep. I had to spend many hours on understanding the low-level details and adding type-hints helped me to understand all the relations.

> I'm wondering if this should be a 2.0 type change. I'm also thinking that when this goes in I'll close any open and stale PRs because most likely they'll need updating due to conflicts. What do you think?

I tried to not break the ABI, but all those `bytes` vs. `unicode` differences might have changed details, especially my commit 1ecaafc9d107 named `fix(rfb): Missing encoding/decoding`: Previously it was _unspecified_ if you still work with low-level `bytes` or already with high-level `unicode`.
The change to use `bytearray()` to buffer the packages should be an internal change and not affect any using application.
The rest is mostly refactoring and style changes, which would not require a major version bump. But I'm okay with this.

I only have looked shortly at the pending PRs: Most of them will no longer work because adding PEP-484 type annotations changed most context. If you take them before this PR I'll rework my branch to incorporate them, but the other way around would also be okay.

pmhahn:
Last run looks better, but the [build for 3.11 is failing](https://github.com/univention/vncdotool/actions/runs/3757803532/jobs/6385410415) because my branch (based on `main`) still uses `isolate` while your branch `master` has it already removed.
I cherry-picked your patches 9ca371f8da99399e553eb10f230b509eb2601d38 3c17a6583a403e3730db6187cbfcee1db3fdea04 30d4db5db9c41f6634b611939fb3d9fef1474a4c 84de3c922b03b21fd95ea8c44ae1503f19e34651 and squashed them into one and fixed `tests/unit/test_command.py`.

tests/functional/test_screen.py
sibson: +1, will add some notes on testing and have a go at making them work in CI too.

sibson:
Let me know when you think this is mostly good enough to merge.  I think we should merge sooner rather than later.  It will be easier for us to make additional cleanups against main than holding a PR open.

pmhahn:
From my PoV it's ready; I undid my change to `pytest` as you already merged your change to plain `unittest` into `main`.
My testing so far found no issues, but  your eyes and testing as the expert are much appreciated.

pmhahn:
> > I'm wondering if this should be a 2.0 type change. I'm also thinking that when this goes in I'll close any open and stale PRs because most likely they'll need updating due to conflicts. What do you think?
> 
> I tried to not break the ABI

Removing Python 2.x support is a breaking change, so using Semantic versioning the version should be `2.0.0`.

sibson:
huge thanks @pmhahn for putting all this together.

sibson:
@pmhahn fyi, https://github.com/sibson/vncdotool/commit/54bcdb23a9e0c560774009680774041af07038f5, sets up the environment to run the functional tests.  They are currently failing due to a bytes/string issue.

pmhahn:
> @pmhahn fyi, [54bcdb2](https://github.com/sibson/vncdotool/commit/54bcdb23a9e0c560774009680774041af07038f5), sets up the environment to run the functional tests. They are currently failing due to a bytes/string issue.

Thanks for that; see #231 which fixes those bytes/string issues.

PR #232, opened 2023-01-17
Cleanups and old PR updates

pmhahn:
While going through old issues and PR I did some more work:
- sort `import` statements with [isort](https://pypi.org/project/isort/) for consistency
- convert constants to `IntEnum`s
- rename "Apple Auth" to "Diffie-Hellman-Auth" as it was already registered as that
- limit `PSEUDO_DESKTOP_SIZE` to 16 bit max, e.g. 65535
- document a bug in `VMWareClient`
- Re-implement `PSEUDO_QEMU_EXTENDED_KEY_EVENTS`  #153
- raise exception in case of protocol errors
- some minor code cleanups

Issue #95, opened 2017-04-23
please add support for ipv6


rpodgorny:
CRITICAL:root:DNS lookup failed: mj95.asterix.cz.

but:

radek@pupik /t/vncdotool-git [10]> systemd-resolve mj95.asterix.cz                                                                                              (+00:00:00) 2017-04-24 00:50:36
mj95.asterix.cz: 2002:2e24:2741:201:d1e3:29c0:8256:4573

...which is a valid ipv6 address (yes, that is a ipv6-only host).

PR #233, opened 2023-01-19
Add IPv6 support

pmhahn:
~~This fixed the two issues you found in my previous #232 and also~~ adds IPv6 support, which is #95

While `reactor.connectTCP()` already supports IPv6, it only works when given an IPv6 address but not, when a DNS name is given which resolved to one. This is a bug in `twisted/internet/tcp.py` where `self.addressFamily = socket.AF_INET6` does not get set in that case.

My next try was to call [socket.getaddrinfo()](https://docs.python.org/3/library/socket.html#socket.getaddrinfo) myself, which returns multiple addresses. Any client must iterate over them in turn and try to connect it until one works. That way most clients will try IPv6 first and fall back to IPv4.
But doing the iteration manually does not work well with the Twisted reactor design as you have to hook the iteration to `clientConnectionFailed()`, which I did not get to work.

So I decided to switch to the newer [endpoints API](https://docs.twistedmatrix.com/en/stable/core/howto/endpoints.html), which already has code to work with `getaddrinfo()`. But it works with `Factory` instead of `Clientfactory`, so all those `clientConnection*` callbacks are no longer invoked. Because of that I added code to still call them.

I have testes this successfully on my local system with IPv4 (`127.0.0.1:0`), IPv6 (`[::1]:0`), DNS (`localhost`), UNIX (`/tmp/qemu`), but please have a closer look yourself.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -32,6 +32,10 @@ The `vncserver` argument needs to be in the format `address[:display|::port]`. F
     # connect to myvncserver.com on port 5902 (two colons needed)
     vncdo -s myvncserver.com::5902 type "hello world"
 
+    # connect via IPv6 to localhost on display :3 (port 5903)
+    vncdo -s '[::1]:3' type "hello IPv6"
+    #         ^   ^ mind those square brackets around the IPv6 address
+
 You can also take a screen capture with::
 
     vncdo -s vncserver capture screen.png
--- a/tests/unit/test_command.py
+++ b/tests/unit/test_command.py
@@ -177,11 +177,45 @@ class TestParseServer(unittest.TestCase):
         assert host == '127.0.0.1'
         assert port == 5910
 
+    def test_missing_display(self):
+        with self.assertRaises(ValueError):
+            command.parse_server(":")
+
+    def test_missing_port(self):
+        with self.assertRaises(ValueError):
+            command.parse_server("::")
+
+    def test_invalid(self):
+        with self.assertRaises(ValueError):
+            command.parse_server(":::")
+
     def test_just_port(self):
         family, host, port = command.parse_server('::1111')
         assert family == socket.AF_INET
         assert host == '127.0.0.1'
         assert port == 1111
+
+    def test_ipv6_host_display(self):
+        family, host, port = command.parse_server('[::1]:10')
+        assert family == socket.AF_INET6
+        assert host == '::1'
+        assert port == 5910
+
+    def test_ipv6_host_port(self):
+        family, host, port = command.parse_server('[::1]::4444')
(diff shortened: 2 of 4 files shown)


README.rst
```diff
@@ -32,6 +32,9 @@ The `vncserver` argument needs to be in the format `address[:display|::port]`. F
     # connect to myvncserver.com on port 5902 (two colons needed)
     vncdo -s myvncserver.com::5902 type "hello world"
 
+    # connect via IPv6 to localhost on dusplay :3 (port 5903)
```
sibson, line 35: typo

sibson (commented):
My only concern is using `[]` to denote the ipv6 address.  It might be confusing to users as that is also used in the help for optional arguments.   

```  -s SERVER, --server=SERVER
                        connect to VNC server at ADDRESS[:DISPLAY|::PORT]
                        [127.0.0.1]```

Is there another way we could detect v6?  I'm naively thinking something along the lines of 

```
split = server.split(':')
if len(split) > 6: 
   # assume v6 host
else:
   # assume v4 host and default to localhost if no host
# handle port/display colons too, which makes it tricky

README.rst
pmhahn: Fixed.

setup.cfg
```diff
@@ -0,0 +1,47 @@
+[metadata]
```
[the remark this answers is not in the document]
pmhahn: I can separate it if you prefer. For me it was just a by-product of fixing the previous small issues.

pmhahn:
> My only concern is using `[]` to denote the ipv6 address. It might be confusing to users as that is also used in the help for optional arguments.
...
> Is there another way we could detect v6? I'm naively thinking something along the lines of
> 
> ```
> split = server.split(':')
> if len(split) > 6: 
>    # assume v6 host
> else:
>    # assume v4 host and default to localhost if no host
> # handle port/display colons too, which makes it tricky
> ```

Yes, that syntax looks a bit confusing, but `vncdotool` is not the first and any protocol where `:` is used to separate the protocol is affected by this; Using `[…]` if quite common, see URLs for example: https://en.wikipedia.org/wiki/IPv6_address#Literal_IPv6_addresses_in_network_resource_identifiers

And the number of `:` may be anything between 2 and 7; all of these examples are valid IPv6 addresses:
- `::1`
- `200d::1`
- `0000:1111:::6666:7777`
- `0000:1111:2222:3333:4444:5555:6666:7777`

sibson (approved):

sibson:
Looks like I need to configure running CI on PRs.  I'll assume the tests pass and merge.

Issue #202, opened 2021-11-02
[BUG] pip installation fails (somehow)


Kochise:
```bash
C:\Program Files\Spyder\Python>python --version
Python 3.7.9

C:\Program Files\Spyder\Python>python -m pip install --upgrade pip
Requirement already satisfied: pip in c:\program files\spyder\python\lib\site-packages (21.3.1)

C:\Program Files\Spyder\Python>python -m pip cache purge
Files removed: 1

C:\Program Files\Spyder\Python>set
HOMEDRIVE=C:
HOMEPATH=\Users\dkoch
NUMBER_OF_PROCESSORS=4
OS=Windows_NT
Path=C:\WINDOWS\system32;C:\WINDOWS;C:\WINDOWS\System32\Wbem;C:\WINDOWS\System32\WindowsPowerShell\v1.0\;C:\WINDOWS\System32\OpenSSH\;C:\windows\SysWow64;C:\Program Files\Spyder\Python;C:\Program Files\Spyder\Python\Lib;C:\Program Files\Spyder\Python\Scripts;
PATHEXT=.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC
PROCESSOR_ARCHITECTURE=x86
PROCESSOR_ARCHITEW6432=AMD64
PROCESSOR_IDENTIFIER=Intel64 Family 6 Model 78 Stepping 3, GenuineIntel
PROCESSOR_LEVEL=6
PROCESSOR_REVISION=4e03
ProgramData=C:\ProgramData
ProgramFiles=C:\Program Files (x86)
ProgramFiles(x86)=C:\Program Files (x86)
ProgramW6432=C:\Program Files
PROMPT=$P$G
PYSDL2_DLL_PATH=C:\Program Files\Spyder\lib
PYTHONHOME=C:\Program Files\Spyder\Python
PYTHONPATH=C:\Program Files\Spyder\Python;C:\Program Files\Spyder\Python\Lib;C:\Program Files\Spyder\Python\Scripts
SystemDrive=C:
SystemRoot=C:\WINDOWS
windir=C:\WINDOWS
[...]

C:\Program Files\Spyder\Python>python -m pip install --upgrade vncdotool
Collecting vncdotool
  Downloading vncdotool-1.0.0.tar.gz (50 kB)
     |████████████████████████████████| 50 kB 615 kB/s
  Preparing metadata (setup.py) ... error
[...]

Issue #160, opened 2019-11-21
CI not building automatically :(


sibson:

pmhahn:
What's missing?
- To upload to PyPI see https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python#publishing-to-package-registries

sibson:
This was probably from the days of Travis/Circle where the webhook to trigger a build was needed.

PR #237, opened 2023-01-25
Improve GitHub actions

pmhahn:
- Update the issue template to not create a task list for every new issue
- Enable PyPI caching
- Also build and test package on Windows #202 #160
- Fix running unit tests on Windows as `socket.AF_UNIX` is undefined there. #202


sibson:
thanks, ✨ 🍰 ✨

PR #239, opened 2023-01-26
Code style issues and more minor fixes

pmhahn:
`flake8` runs as part of the GitHub actions each time and prints many issues
- Enable `flake8` for [rfb.py](vncdotool/rfb.py)
- Enable `flake8` for foreign-code of [pyDes.py](vncdotool/pyDes.py)
- reformat all code to follow PEP-8

On the way to that I also
- removed some unused `imports`
- removed some dead variable assignments
- fixed the fallback logic for `import PIL` again
- removed unneeded options to `optparse.OptionParser().add_option()`
- removed one old Python 2 related comment
- converted the tests to also use f-strings
- added type hints for most tests

If you like it I can also integrate `isort`, `black`, `myp`, `pre-commit`, … for automatic code formatting and checking in the future. In my other projects these tools help to keep the code clean, consistent and working.


vncdotool/command.py
```diff
@@ -257,30 +256,45 @@ def build_proxy(options: optparse.Values) -> VNCLoggingServerFactory:
...
 
 
 def setup_logging(options: optparse.Values) -> None:
     # route Twisted log messages via stdlib logging
     if options.logfile:
-        handler = logging.handlers.RotatingFileHandler(options.logfile,
-                                      maxBytes=5*1024*1024, backupCount=5)
+        handler = logging.handlers.RotatingFileHandler(
+            options.logfile,
+            maxBytes=5 << 20,
```
sibson, line 295: nit, I find N*1024*1024 more readable, as I don't have to think as much about how big the number is.  Maybe we add a comment for size?

sibson (approved):
I'd been avoiding making large changes to pyDes and rfb, with the intent of merging them back upstream.  That never happened and with py3k I think that ship has sailed, so it makes sense to "fix" the formatting of them.

vncdotool/command.py
pmhahn: Okay, I'll try to remember this for the next time.

pmhahn:
> I'd been avoiding making large changes to pyDes and rfb, with the intent of merging them back upstream

For [pyDes.py](vncdotool/pyDes.py) I had found an old version on [PyPI](https://pypi.org/project/pyDes/) from 2010, but it linked to a broken homepage. It looks like it once was on [SourceForce](https://pydes.sourceforge.net/) but now lives on [GitHub](https://github.com/twhiteman/pyDes). As `vncdotool` already carried a heavily modified version with all that Python 2/3 `bytes` vs. `unicode` stuff, I did not bother to re-synchronize the upstream version back into ours.

For [rfb.py](vncdotool/rfb.py) I did not look for any upstream version, from which it has diverged largely by now. I had no idea that there was an upstream version, which seems to have been on [GoogleCode](https://code.google.com/archive/p/python-vnc-viewer/), now also on [GitHub](https://github.com/techtonik/python-vnc-viewer).

It might be a good idea to add those original projects to the [LICENSE.txt](LICENSE.txt) to help other (and me) to find them.

PR #251, opened 2023-02-08
Add username parameter to connect call

JacobHearst:
Just exposing the username parameter to the python API. I'm _very_ unfamiliar with the internals of this library so please let me know if there are other changes required

### DIFF
--- a/vncdotool/api.py
+++ b/vncdotool/api.py
@@ -117,11 +117,11 @@ class ThreadedVNCClientProxy:
 
 def connect(
     server: str,
-    username: Optional[str] = None,
     password: Optional[str] = None,
     factory_class: Type[VNCDoToolFactory] = VNCDoToolFactory,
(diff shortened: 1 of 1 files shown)


vncdotool/api.py
```diff
@@ -117,6 +117,7 @@ def __dir__(self) -> List[str]:
 
 def connect(
     server: str,
+    username: Optional[str] = None,
```
pmhahn, line 120: Adding a new parameter between of existing parameters breaks the API: Anyone doing a `connect("localhost:5900", "password") will now pass the 2nd argument to `username` instead.

Best append it at the end as the last parameter.

JacobHearst: Good catch, been a while since I've done python

pmhahn:
`username` is only used with the DH authentication type, which _Apple Remote Desktop_ seems to use; Sadly I found no other (free) alternative to test this, so this is only by code-review.
Thank you for your contribution.

PR #254, opened 2023-02-24
bugfix to avoid always calling cmake

sibson:
Extra BUILD_DIR in path meant we never created the target file


pmhahn:
Looks good, cherry-picked for testing, works-for-me.

Issue #256, opened 2023-04-14
1.1.0 client triggers twisted.internet.defer.AlreadyCalledError on disconnect()


reticulatedpines:
**vncdotool version**
1.1.0

**VNC server and version**
Qemu 4.2.1 via -vnc.  Hopefully not relevant.

**Steps to reproduce**

```  
#!/usr/bin/env python3

import vncdotool
from vncdotool import api

print(vncdotool.__version__)
client = vncdotool.api.connect(":1234")
client.keyPress("m") # just to test the connection really established
client.disconnect() # throws UnhandledError and twisted AlreadyCalledError with 1.1.0,
                    # silent with 1.0.0
vncdotool.api.shutdown()
```  

**Expected result**
Clean disconnect with no error when calling disconnect().  Unless there's some API change with 1.1.0 and I'm doing it wrong?

**Which erroneous result did you get instead**
UnhandledError and twisted AlreadyCalledError.

**Additional information**

I did some digging, and notice that command.py does this:
```  
    def clientConnectionLost(self, connector: IConnector, reason: Failure) -> None:
        if reason.type == ConnectionDone:
            self.done(0)
        else:
            self.error(reason)
```  

If I change client.py so it does similar, I don't see the error...  but I definitely don't understand this code well enough to know if that's an appropriate fix or I'm just hiding something:
```  
    def clientConnectionLost(self, connector: IConnector, reason: Failure) -> None:
        #self.deferred.errback(reason)
        if reason.type == ConnectionDone:
            pass
        else:
            self.deferred.errback(reason)
```  

It does seem sensible that ConnectionDone is a special case error, and currently client.py isn't special casing it.

reticulatedpines:
Slightly bad report, my mistake.  The error doesn't trigger on the disconnect().  It triggers on shutdown().  And without the shutdown(), the script never exits, similar to issue 255.  But there's no explicit multiprocessing or threads in this one.

sibson:
fixed in 1.2.0

PR #257, opened 2023-04-20
rework handling of error, ref #256, #189

sibson:
revert #189, connectionLost is called after the deferred was called for a successful or failed connection. Leading to AlreadyCalledError being raised

Further investigation uncovered a few bugs
 - api.ahutdown() if called multiple times would try to join a dead thread
 - renamed functions to be more readable to aid debugging
 - ensure we shutdown the reactor and associated threads when exiting program due to an exception.  This prevents hangs because the reactor creates a ThreadPool with non-daemon threads
 - BREAKING, may now raise Twisted Exceptions rather than VNCDoException wrapping the underlying exceptions

### DIFF
--- a/vncdotool/api.py
+++ b/vncdotool/api.py
@@ -142,9 +142,6 @@ def connect(
     http://twistedmatrix.com/documents/13.0.0/core/howto/choosing-reactor.html
     for a better method of intergrating vncdotool.
     """
-
-
-
     if not reactor.running:
 
         # For api.shutdown to kill reactor threads before trying to exit due to an Exception
--- a/vncdotool/api.py
+++ b/vncdotool/api.py
@@ -94,12 +94,13 @@ class ThreadedVNCClientProxy:
             return reason
 
         def callable_threaded_proxy(*args: Any, **kwargs: Any) -> Any:
-            reactor.callFromThread(
-                    self.factory.deferred.addCallbacks, # ensure we're connected
-                                   threaded_call,
-                                   errback_not_connected,
-                                   args,
-                                   kwargs)
+        reactor.callFromThread(
+            self.factory.deferred.addCallbacks,  # ensure we're connected
+            threaded_call,
+            errback_not_connected,
+            args,
+            kwargs,
+        )
             try:
                 result = self.queue.get(timeout=self._timeout)
             except queue.Empty:
--- a/vncdotool/api.py
+++ b/vncdotool/api.py
@@ -161,8 +161,6 @@ def connect(
         )
         _THREAD.daemon = True
         _THREAD.name = 'Twisted Reactor'
-
-
         _THREAD.start()
 
         observer = PythonLoggingObserver()
--- a/vncdotool/api.py
+++ b/vncdotool/api.py
@@ -9,6 +9,7 @@ import queue
 import socket
 import sys
 import threading
+from types import TracebackType
 from typing import Any, List, Optional, Type, TypeVar
 
 from twisted.internet import reactor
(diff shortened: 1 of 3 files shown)


vncdotool/api.py
```diff
@@ -137,14 +141,28 @@ def connect(
     http://twistedmatrix.com/documents/13.0.0/core/howto/choosing-reactor.html
     for a better method of intergrating vncdotool.
     """
+
+
+
```
pmhahn, line 146: PEP-8 calls for only one blank line:
```suggestion

```

vncdotool/api.py
```diff
@@ -137,14 +141,28 @@ def connect(
...
     if not reactor.running:
+
+        # For api.shutdown to kill reactor threads before trying to exit due to an Exception
+        sys_excepthook = sys.excepthook
+        def ensure_reactor_stopped(etype, value, traceback):
```
pmhahn, line 151: Nit: Missing type annotation
```suggestion
        def ensure_reactor_stopped(etype: Type[BaseException], value: BaseException, traceback: TracebackType) -> None:
```

That will require some additional `imports` *before* line 12:
```python
from types import TracebackType

```

vncdotool/api.py
```diff
@@ -82,32 +81,37 @@ def disconnector(protocol: VNCDoToolClient) -> None:
...
             d = maybeDeferred(method, protocol, *args, **kwargs)
             d.addBoth(result_callback)
             return d
 
-        def proxy_call(*args: Any, **kwargs: Any) -> Any:
-            reactor.callFromThread(self.factory.deferred.addCallbacks,
-                                   callback, errback, args, kwargs)
+        def errback_not_connected(reason: Failure, *args: Any, **kwargs: Any) -> None:
+            self.queue.put(reason)
+            return reason
```
pmhahn, line 94: The type annotation does not match the implementation: ~Either delete the `return reason` to return `None` or~ change the signature to be
```suggestion
        def errback_not_connected(reason: Failure, *args: Any, **kwargs: Any) -> Failure:
            self.queue.put(reason)
            return reason
```
According to [addCallback](https://docs.twistedmatrix.com/en/stable/api/twisted.internet.defer.Deferred.html#addCallbacks) the function should return `Failure`.

pmhahn (approved):
I only had a brief look, but noticed the hang during shutdown myself in one of my previous tests. Thank you for finding the cause.

I found https://stackoverflow.com/questions/49992329/the-workers-in-threadpoolexecutor-is-not-really-daemon which sheds additional light on the `ThreadPoool.daemon` issue.

According to the documentation [Thread.join()](https://docs.python.org/3.11/library/threading.html#threading.Thread.join) can be called multiple times, even on a dead thread. So it should not have been a problem, but for consistency I'm with you, especially when you want to call `api.connect()` and `shutdown()` multiple times from the same process.

I have not yet (extensively) tested your change, but your change looks good and `make test` succeeds.

sibson:
I wonder if using atexit() would be cleaner than what I've done. I'll play with it to see if it works.

vncdotool/api.py
```diff
@@ -82,32 +81,38 @@ def disconnector(protocol: VNCDoToolClient) -> None:
...
+
+        def callable_threaded_proxy(*args: Any, **kwargs: Any) -> Any:
+        reactor.callFromThread(
+            self.factory.deferred.addCallbacks,  # ensure we're connected
+            threaded_call,
+            errback_not_connected,
+            args,
+            kwargs,
+        )
```
pmhahn, line 103: The indention is again wrong here. 
```suggestion
        def callable_threaded_proxy(*args: Any, **kwargs: Any) -> Any:
            reactor.callFromThread(
                self.factory.deferred.addCallbacks,  # ensure we're connected
                threaded_call,
                errback_not_connected,
                args,
                kwargs,
            )
```

PR #258, opened 2023-04-22
run black code formatter

sibson:


pmhahn:
:+1: 
Maybe also [show your style](https://pypi.org/project/black/) and add this snippet to the [`README.rst`](https://github.com/sibson/vncdotool/blob/main/README.rst):
```rst
.. image:: https://img.shields.io/badge/code%20style-black-000000.svg
    :target: https://github.com/psf/black
```

If you look at this maybe also fix the broken PyPI-Link, which always shows this to me:
> package or version not found

sibson:
thanks for the tip on the README, I've wondered why it doesn't work but never investigated

PR #268, opened 2023-08-27
Simplify RPM support to Python 3 only

pevogam:
Now that the VNCDoTool no longer officially supports Python 2 we can simplify the RPM spec file and restore compatibility with most recent RPM packager distros.

### DIFF
--- a/rpm/python-vncdotool.spec
+++ b/rpm/python-vncdotool.spec
@@ -1,7 +1,7 @@
-# A basic RPM packaging for VNCDoTool with both Python 2 and 3.
+# A basic RPM packaging for VNCDoTool with Python 3.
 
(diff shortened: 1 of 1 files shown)


pevogam:
Hi @sibson, have you considered a later update of the python packaging via a toml file as described in the latest packaging guidelines from https://packaging.python.org/en/latest/tutorials/packaging-projects/#creating-pyproject-toml? I am just wondering. Otherwise I will be glad if you consider this pull request restricts RPM packaging to python 3 only, updates, and simplifies it.

sibson:
thanks, ✨ 🍰 ✨

I've not thought about toml.  Assuming there are no regressions I wouldn't object to upgrading.

PR #278, opened 2024-01-08
Switch from pycryptodomex to cryptography.io

geofft:
cryptography.io wraps an existing well-known cryptography library (OpenSSL) instead of implementing routines itself, and it's already a dependency of Twisted, so this reduces the overall number of cryptography libraries needed.


geofft:
I've done a little bit of unit testing by hand but let me try to make some real connections with this too, since I don't think there's much test coverage for this code. Also I'm open to feedback on whether this is a desirable thing at all.

geofft:
OK, I've confirmed that this works in both of the cases that are affected, i.e., password auth (with Xtightvnc) and Apple Remote Desktop (with macOS 13.6), so I think I got the code right :)

Neustradamus:
@sibson: Have you seen this @geofft PR?

sibson:
@geofft could you resolve conflicts?

I don't have a strong opinion here, my quick google suggests either is fine but gives cryptography a slight edge due to fewer sharp edges

@pmhahn you most recently updated the dependency here.  Do you have an opinion?

requirements.txt
```diff
@@ -1,4 +1,4 @@
-Pillow>=9.3.0
-Twisted>=22.10.0
-zope.interface>=5.4.0
-pycryptodomex>=3.12.0
+Pillow==9.3.0
```
sibson, line 1:
```suggestion
Pillow==9.3.0
```
```suggestion
Pillow>=9.3.0
```

requirements.txt
```diff
@@ -1,4 +1,4 @@
...
+Twisted==22.10.0
```
sibson, line 2:
```suggestion
Twisted>=22.10.0
```

Issue #288, opened 2024-06-14
Programs using api.connect hang unless api.shutdown is explicitly called before exit


cdleonard:
Please include the following information:

Client: vncdotool version: 1.2.0
Server: TigerVNC Server version 1.13.1, built Jan 30 2024 00:00:00

Run the following program:
```python
#! /usr/bin/env python3
import sys
import threading

import vncdotool.api


def main():
    server = sys.argv[1]
    passwd = sys.argv[2]
    with vncdotool.api.connect(server, passwd):
        pass
    for th in threading.enumerate():
        sys.stdout.write(f"thread name {th.name} daemon {th.daemon}\n")


if __name__ == "__main__":
    main()
```

Expected result: shuts down nicely

Actual result:
```
thread name MainThread daemon False
thread name Twisted Reactor daemon True
thread name PoolThread-twisted.internet.reactor-0 daemon False
*HANG*
```

Apparently the twisted reactor created by vncdotool is marked as a background thread but it spawns workers which are not marked as background and the program does not stop. Perhaps this used to work but was broken by updates to twisted itself? I know nothing about the twisted framework.

My goals is to automate clicking through VNC inside pytest and I found that this workaround works for me:

```python
def pytest_sessionfinish():
    from vncdotool import api

    api.shutdown()
```

One idea to work around the reactor not shutting down automatically would be to reference-count it from connect/disconnect.

pmhahn:
The Twisted reactor is not restartable!  A reactor is only started by the API when no-one is yet running. Stopping it after  `with vncdotool.api.connect(…): …` would prevent you from using the API again in the same process. Therefore `vncdotool` decided to keep the reactor running in a background thread, which is started on first use and needs to be shutdown explicitly by running `shutdown()` explicitly as you found out. That is why the function is there.

FYI: Twisted is `async` from a time when Python did not have nativ `async` itself, it sometimes matches badly with Threads and leads to all kind of problems when combined. Read [Using Threads in Twisted](https://docs.twisted.org/en/twisted-18.7.0/core/howto/threading.html) and [Choosing a Reactor](https://docs.twisted.org/en/twisted-18.7.0/core/howto/choosing-reactor.html). Even running it in a background threads leads to all kind of strange issues, e.g. `signal` handling not working correctly.

pevogam:
[...]

Issue #287, opened 2024-06-09
'SelectReactor' object has no attribute 'doPoll'


JackYrp:
When I use mouseDrag function ,this error has occurred:
 File "C:\ProgramData\Anaconda3\envs\mh\lib\site-packages\vncdotool\api.py", line 119, in callable_threaded_proxy
    result.raiseException()
  File "C:\ProgramData\Anaconda3\envs\mh\lib\site-packages\twisted\python\failure.py", line 475, in raiseException
    raise self.value.with_traceback(self.tb)
AttributeError: 'SelectReactor' object has no attribute 'doPoll'

What should I do now,I am in hurry
Thank you very much guys

JackYrp:
can some one help me?

pmhahn:
> can some one help me?

OpenSource if about being able to *contribute* to others, not *demanding* help from others ;-)

Nevertheless you might want to try the branch from #289 I just pushed which gets rid of some nastiness in `mouseDrag()`, which
- used `time.sleep()` with is a no-no with Twisted
- used some internal knowledge of the [SelectReactor](https://docs.twisted.org/en/twisted-18.4.0/core/howto/choosing-reactor.html#select-based-reactor)

Issue #260, opened 2023-05-05
1.10 use mouseDrag not work，err is AttributeError: 'SelectReactor' object has no attribute 'doPoll'


jian-kuang:
Please include the following information:

**vncdotool version**

**VNC server and version**

**Steps to reproduce**
-

**Expected result**

**Which erroneous result did you get instead**

**Additional information**

jian-kuang:
env: windows10 system 
err detail: 
`  File "G:\Anaconda\envs\mhxy\lib\site-packages\vncdotool\api.py", line 109, in callable_threaded_proxy
    result.raiseException()
  File "G:\Anaconda\envs\mhxy\lib\site-packages\twisted\python\failure.py", line 504, in raiseException
    raise self.value.with_traceback(self.tb)
AttributeError: 'SelectReactor' object has no attribute 'doPoll'`

jian-kuang:
Do not support dopoll under win10, just replace it with Doselet IO model

pmhahn:
Probably fixed by #289

PR #289, opened 2024-06-17
A bunch of fixes and improvemenbs

pmhahn:
- Fix building documentation
- Improve documentation, e.g. Sphinx reST syntax, Python 3.12 support, API #288
- Improve type hints (again)
- Fix functional tests
- Fix `mouseDrag` using `sleep()` and depending on [SelectReactor](https://docs.twisted.org/en/twisted-18.4.0/core/howto/choosing-reactor.html#select-based-reactor) internal knowledge. #287 #260

As always more testing (from you) is appreciated 😉


sibson:
Thank you for the attentiveness and thoroughness as always.  I delayed looking at this for a while because it seemed so large, but it looks like the semantic changes are minimal.

PR #293, opened 2025-01-04
Allow specifying a format with captureScreen

erjiang:
This allows it to be used with file handles, otherwise pillow doesn't know what format to write.

e.g.:

```
    buffer = BytesIO()
    # Capture screen to memory buffer in PNG format
    client.captureScreen(buffer, format="png")
```

### DIFF
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -1,4 +1,5 @@
 from unittest import TestCase, mock
+import io
 
 from vncdotool import client, rfb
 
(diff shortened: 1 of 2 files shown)


Neustradamus:
@sibson: Have you seen this @erjiang PR?

sibson:
Thanks, @erjiang.  Could you document what formats are valid and include a test case to ensure we don't regress this functionality.

I don't think the usage of format are very discoverable, even setting the default to 'png' or whatever PIL defaults to, would make it more clear.  An alternative might be to name it image_format or file_type.

erjiang:
> I don't think the usage of format are very discoverable, even setting the default to 'png' or whatever PIL defaults to, would make it more clear. An alternative might be to name it image_format or file_type.

I think this is mostly because of using Pillow - the list of formats is likely to change depending on the Pillow version and we don't want to set a default because the default should be to guess based on filename. Added some documentation to point to the Pillow image format list.

sibson:
👍

Issue #267, opened 2023-08-22
Capture Command Hangs


squat:
Please include the following information:

**vncdotool version**
1.2.0

**VNC server and version**
WayVNC 0.6.2

**Steps to reproduce**:
```
vncdotool -i -v -v -v -s 192.168.0.1::5900 capture out.png
```
Output:
```
INFO:root:connecting to 192.168.0.1:5900
INFO:twisted:Starting factory <vncdotool.command.VNCDoCLIFactory object at 0x7fb290c80750>
INFO:twisted:Using protocol version 3.8
INFO:twisted:Offered <AuthTypes.NONE: 1>
INFO:twisted:Native PixelFormat(bpp=32, depth=24, bigendian=False, truecolor=True, redmax=255, greenmax=255, bluemax=255, redshift=16, greenshift=8, blueshift=0) bytes=4
INFO:twisted:Offering <Encoding.RAW: 0>
INFO:twisted:Offering <Encoding.PSEUDO_DESKTOP_SIZE: -223>
INFO:twisted:Offering <Encoding.PSEUDO_LAST_RECT: -224>
INFO:twisted:Offering <Encoding.PSEUDO_QEMU_EXTENDED_KEY_EVENT: -258>
INFO:root:connected to b'WayVNC'
DEBUG:vncdotool.client:captureScreen out.png
INFO:twisted:x=0 y=0 w=0 h=0 <Encoding.PSEUDO_QEMU_EXTENDED_KEY_EVENT: -258>
```

**Expected result**
I expected the command to complete and produce a file.

**Which erroneous result did you get instead**
The command never completes. I have to interrupt it.

**Additional information**
Other `vncdotool` commands _do_ work just fine, such as `type`.

pmhahn:
- [WayVNC](https://github.com/any1/wayvnc)
- [ ] Need to check why you get back a `PSEUDO_QEMU_EXTENDED_KEY_EVENT` instead of the `RAW` screen.

josch:
Same issue with vncdotool 1.2.0 and wayvnc 0.8.0. Any updates?

josch:
Since `vncdotool` is not working for me, I re-implemented the needed functionality in a short Python script. Maybe it helps others:

```python3
#!/usr/bin/env python3
#
# SPDX-License-Identifier: GPL-3.0
#
# Copyright 2024 Johannes Schauer Marin Rodrigues <josch@mister-muffin.de>
#
[...]

PR #315, opened 2026-04-08
Fix captureScreen hang when server responds with only pseudo-encodings

HZNAN:
## Problem
`captureScreen` (and the CLI `capture` command) hangs indefinitely when the VNC server responds to a `FramebufferUpdateRequest` with only pseudo-encoding rectangles and no actual pixel data.
This has been reported in #267 and affects multiple VNC servers including TigerVNC and WayVNC.
### Reproduction
$ vncdo -s ::5900 -p capture screenshot.png

Debug log shows the server returns only pseudo-encodings:
DEBUG:vncdotool.client:captureScreen screenshot.png INFO:twisted:x=0 y=0 w=0 h=0 <Encoding.PSEUDO_QEMU_EXTENDED_KEY_EVENT: -258> INFO:twisted:x=0 y=0 w=0 h=0 <Encoding.PSEUDO_LAST_RECT: -224>

hangs forever here
### Root Cause
In `rfb.py`, `_doConnection()` only calls `commitUpdate()` when `rectanglePos` is non-empty:
```python
if self.rectanglePos:
    self.commitUpdate(self.rectanglePos)
Pseudo-encoding handlers (e.g. PSEUDO_QEMU_EXTENDED_KEY_EVENT) append to rectanglePos then immediately delete their entry. When a FramebufferUpdate contains only pseudo-encodings, rectanglePos ends up empty → commitUpdate() is never called → the Deferred from captureScreen is never resolved → infinite hang.

Fix
Always call commitUpdate() when all rectangles in a FramebufferUpdate have been processed, regardless of whether rectanglePos is empty. This ensures the Deferred is always resolved.

Test Environment
vncdotool main branch
VNC Server: TigerVNC 1.13 (Windows)
Client: Python 3.11 (Linux, Docker container)

### DIFF
--- a/vncdotool/rfb.py
+++ b/vncdotool/rfb.py
@@ -698,8 +698,7 @@ class RFBClient(Protocol):  # type: ignore[misc]
         if self.rectangles:
             self.expect(self._handleRectangle, 12)
         else:
-            if self.rectanglePos:
-                self.commitUpdate(self.rectanglePos)
+            self.commitUpdate(self.rectanglePos)
             self.expect(self._handleConnection, 1)
 
     def _handleRectangle(self, block: bytes) -> None:


sibson:
thanks, ✨ 🍰 ✨

PR #316, opened 2026-04-11
Fix ThreadedVNCClientProxy.disconnect() to block until disconnected

### LINKED ISSUE #314: ThreadedVNCClientProxy.disconnect() does not block until disconnection complete
I use vncdotool synchronously via the API it exposes, my server is multi screen and I have two clients.
My code, simplified, looks something like this:
```python
class X:
    def __init__(self):
        self._client1 = None
        self._client2 = None

    @property
    def client1(self):
        if not self._client1:
            client = api.connect(VNC_ADDRESS, password=VNC_PASSWORD, timeout=10)
            client.mouseMove(0, 0)
            self._client1 = client
        return self._client1

    @property
    def client2(self):
        if not self._client2:
            client = api.connect(VNC_ADDRESS + ":1", password=VNC_PASSWORD, timeout=10)
            client.mouseMove(0, 0)
            self._client2 = client
        return self._client2

    def reconnect_clients(self):
        self.log.info("Reinitializing device")
        if self._client1:
            self._client1.disconnect()
        if self._client2:
            self._client2.disconnect()
        self._client1 = None
        self._client2 = None
        do_stuff(self.client1)
        do_stuff(self.client2)
```

The vncdotool code for disconnect in `api.py` looks like this:
```python
    def disconnect(self):
        def disconnector(protocol):
            protocol.transport.loseConnection()
        reactor.callFromThread(self.factory.deferred.addCallback, disconnector)
```
It schedules `disconnector` for execution but then returns before the connection has actually disconnected. As the function has returned in the main thread, my code carries on and tries to use the client again. I see regular issues where `do_stuff(self.client1)` blows up with `TimeoutError: Timeout while waiting for client response` (coming from `proxy_call` on `ThreadedVNCClientProxy`).
I think this is because my VNC server (TightVNC on Windows) is doing something racy and closes the new and old connections when the disconnect actually is executed by the reactor in certain cases.

I've not been able to reliably reproduce this in isolation, but the problem would be solved if `ThreadedVNCClientProxy.disconnect()` blocked until disconnection had actually occurred. Given `ThreadedVNCClientProxy` is acting as a syncronous API, it should also act synchronous here.

sibson:
disconnect() previously scheduled loseConnection() and returned immediately, before the TCP connection was actually closed. This caused race conditions where code running after disconnect() could still be affected by the old connection being torn down (e.g. a VNC server closing both old and new connections in certain conditions).

Fix by registering a disconnect callback on VNCDoToolFactory and waiting on a threading.Event in disconnect() until clientConnectionLost fires, making the synchronous API truly synchronous on disconnect as well.

Fixes #314

https://claude.ai/code/session_01AA4gPSov9NB2ewZLgtAp16

### DIFF
--- a/vncdotool/api.py
+++ b/vncdotool/api.py
@@ -71,10 +71,17 @@ class ThreadedVNCClientProxy:
         reactor.callWhenRunning(factory_connect, self.factory, host, port, family)
 
     def disconnect(self) -> None:
+        event = threading.Event()
+
+        def on_disconnected(reason: Any) -> None:
+            event.set()
+
         def disconnector(protocol: VNCDoToolClient) -> None:
+            self.factory._disconnect_callbacks.append(on_disconnected)
             protocol.transport.loseConnection()
 
         reactor.callFromThread(self.factory.deferred.addCallback, disconnector)
+        event.wait(timeout=self.timeout)
 
     def __getattr__(self, attr: str) -> Any:
         method = getattr(self.factory.protocol, attr)
--- a/vncdotool/client.py
+++ b/vncdotool/client.py
@@ -12,7 +12,7 @@ import math
 import socket
 from pathlib import Path
 from struct import pack
-from typing import IO, Any, Iterator, TypeVar, Union
+from typing import IO, Any, Callable, Iterator, TypeVar, Union
 
 from twisted.internet import reactor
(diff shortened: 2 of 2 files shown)


Jamie-:
So this does change the behaviour to be synchronous now as expected, but strangely doesn't fix my issue.
I definitely think this is worth merging though :+1:
Thank you!

Issue #322, opened 2026-05-15
vncdotool silently disconnects –  hangs on shared tigervnc


TeofilisMartisius:
Hello,

Using vncdotool from Python, on Debian Linux. 

When vncdotool connection is made and screen refresh is requested, vncdotool hangs.

Python code:

```
from vncdotool import api as _vnc_api
self._client = _vnc_api.connect(server=server, password=password)
self._client.refreshScreen(incremental=False) # hangs here
```
vncdotool version used is 1.3.1.dev0

VNC server running is tigervnc, version 1.15.0+dfsg-2+b1. VNC server is launched using following command:

`/usr/bin/vncserver -SecurityTypes VncAuth -alwaysshared -depth 24 -geometry 1600x1000 :1 -xstartup /usr/bin/icewm
`
(TLS encryption disabled to allow inspection of protocol exchange via wireshark. It operates the same way with TLS enabled).

This server already has a single VNC client connected, VNC client is connected using following command:

`/usr/bin/xvncviewer -Shared`

I'm using vncdotool to open a second connection.

This failure happens because tigervnc in such conditions sends some VNC Fence requests, and vncdotool does not recognize them and silently disconnects, while refreshScreen() call is left waiting forever.

To fix this I have added a very simple VNC fence support to vncdotool. It does fix this issue at least for me. I will send a PR. This implementation was produced with the help of AI, so please review it before merging. I have checked the code myself and it does look reasonable, but I do not have sufficient familiarity with vncdotool codebase to be sure.

Pull request here: https://github.com/sibson/vncdotool/pull/323

Please do not hesitate to contact me if would like some changes in the PR or if you need anything else.

Sincerely,
Teofilis Martisius

sibson:
🤖 Claude Code:

Thanks for the report and for #323 — Fence support is landing in #471, built on your implementation.

Before merging it I tried to reproduce the mechanism you describe, and could not. Writing up what was tried, because it means something else is going on in your setup and the capture would tell us what.

**What was tested.** A socket-level client, so exactly which encodings are offered is controlled, against Xvnc started the way yours is (`-SecurityTypes None -geometry 256x192 -depth 24 -AlwaysShared`), on two versions: TigerVNC 1.12.0, and 1.15.0 from Debian trixie — the same `1.15.0+dfsg` you are running.

| Scenario | Result |
|---|---|
[...]

PR #323, opened 2026-05-15
Initial VNC Fence implementation

TeofilisMartisius:
Add basic ServerFence/ClientFence implementation, which is required to work with shared tigervnc #322

### DIFF
--- a/vncdotool/rfb.py
+++ b/vncdotool/rfb.py
@@ -189,6 +189,14 @@ class HextileEncoding(IntFlag):
     SUBRECTS_COLORED = 16
 
 
+class FenceFlags(IntFlag):
+    """Client/Server fence: flags."""
+    BLOCK_BEFORE = 1 << 0
+    BLOCK_AFTER = 1 << 1
+    SYNC_NEXT = 1 << 2
+    REQUEST = 1 << 31
+
+
 class AuthTypes(IntEnumLookup):
     """:rfc:`6143` §7.1.2. Security Handshake."""
 
@@ -456,6 +464,7 @@ class RFBClient(Protocol):  # type: ignore[misc]
         Encoding.PSEUDO_DESKTOP_SIZE,
         Encoding.PSEUDO_LAST_RECT,
         Encoding.PSEUDO_QEMU_EXTENDED_KEY_EVENT,
+        Encoding.PSEUDO_FENCE,
     }
 
     _HEADER = b"RFB 000.000\n"
@@ -684,6 +693,8 @@ class RFBClient(Protocol):  # type: ignore[misc]
             self.expect(self._handleConnection, 1)
         elif msgid == MsgS2C.SERVER_CUT_TEXT:
             self.expect(self._handleServerCutText, 7)
+        elif msgid == MsgS2C.SERVER_FENCE:
+            self.expect(self._handleServerFence, 8)
         else:
             log.msg(f"unknown message received {MsgS2C.lookup(msgid)!r}")
             self.transport.loseConnection()
@@ -1220,6 +1231,22 @@ class RFBClient(Protocol):  # type: ignore[misc]
         self.copy_text(block.decode("iso-8859-1"))
         self.expect(self._handleConnection, 1)
 
+    def _handleServerFence(self, block: bytes) -> None:
+        # 3 bytes padding + U32 flags + U8 payload-length
+        flags, length = unpack("!xxxIB", block)
+        if length:
+            self.expect(self._handleServerFencePayload, length, flags)
+        else:
+            self._doFence(flags, b"")
+
+    def _handleServerFencePayload(self, block: bytes, flags: int) -> None:
+        self._doFence(FenceFlags(flags), block)
+
+    def _doFence(self, flags: FenceFlags, payload: bytes) -> None:
+        if flags & FenceFlags.REQUEST:
+            self.clientFence(flags & ~FenceFlags.REQUEST, payload)
+        self.expect(self._handleConnection, 1)
+
     # ------------------------------------------------------
     # incomming data redirector
     # ------------------------------------------------------
@@ -1299,6 +1326,10 @@ class RFBClient(Protocol):  # type: ignore[misc]
         """
         data = message.encode("iso-8859-1")
         self.transport.write(pack("!BxxxI", 6, len(data)) + data)
+
+    def clientFence(self, flags: FenceFlags, payload: bytes = b"") -> None:
+        """Respond to or initiate a Fence synchronisation message."""
+        self.transport.write(pack("!BxxxIB", 248, flags, len(payload)) + payload)
 
     # ------------------------------------------------------
     # callbacks
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -38,6 +38,7 @@ class TestVNCDoToolClient(TestCase):
             client.rfb.Encoding.PSEUDO_DESKTOP_SIZE,
             client.rfb.Encoding.PSEUDO_LAST_RECT,
             client.rfb.Encoding.PSEUDO_QEMU_EXTENDED_KEY_EVENT,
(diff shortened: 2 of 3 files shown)


vncdotool/rfb.py
```diff
@@ -456,8 +456,11 @@ class RFBClient(Protocol):  # type: ignore[misc]
         Encoding.PSEUDO_DESKTOP_SIZE,
         Encoding.PSEUDO_LAST_RECT,
         Encoding.PSEUDO_QEMU_EXTENDED_KEY_EVENT,
+        Encoding.PSEUDO_FENCE,
     }
 
+    FENCE_FLAG_REQUEST = 0x8000_0000
```
pmhahn: Can you please move it before/after `class RFBClient`, change this to
```python
from enum import IntFlag
class FenceFlags(IntFlag):
    """Client/Server fence: flags."""
    BLOCK_BEFORE = 1 << 0
    BLOCK_AFTER = 1 << 1
    SYNC_NEXT = 1 << 2
    REQUEST = 1 << 31
```
and use that below?

vncdotool/rfb.py
```diff
@@ -1220,6 +1225,22 @@ def _handleServerCutTextValue(self, block: bytes) -> None:
...
+            self.expect(self._handleServerFencePayload, length, flags)
+        else:
+            self._doFence(flags, b"")
+
+    def _handleServerFencePayload(self, block: bytes, flags: int) -> None:
+        self._doFence(flags, block)
+
+    def _doFence(self, flags: int, payload: bytes) -> None:
+        if flags & self.FENCE_FLAG_REQUEST:
+            self.clientFence(flags & ~self.FENCE_FLAG_REQUEST, payload)
```
pmhahn:
```suggestion
    def _handleServerFencePayload(self, block: bytes, flags: int) -> None:
        self._doFence(FenceFlags(flags), block)

    def _doFence(self, flags: FenceFlags, payload: bytes) -> None:
        if flags & FenceFlags.REQUEST:
            self.clientFence(flags & ~FenceFlags.REQUEST, payload)
```

vncdotool/rfb.py
```diff
@@ -1300,6 +1321,10 @@ def clientCutText(self, message: str) -> None:
         data = message.encode("iso-8859-1")
         self.transport.write(pack("!BxxxI", 6, len(data)) + data)
 
+    def clientFence(self, flags: int, payload: bytes = b"") -> None:
```
pmhahn:
```suggestion
    def clientFence(self, flags: FenceFlags, payload: bytes = b"") -> None:
```

pmhahn (approved):
Looks okay from reading the code/PR.
I would find it clearer to add a `class FencFlags(IntFlag)`, which improves type safety and debugging and future additions to implement `Fence` for real.

I've no yet run any tests.

vncdotool/rfb.py
```diff
@@ -456,8 +456,11 @@ class RFBClient(Protocol):  # type: ignore[misc]
         Encoding.PSEUDO_DESKTOP_SIZE,
         Encoding.PSEUDO_LAST_RECT,
         Encoding.PSEUDO_QEMU_EXTENDED_KEY_EVENT,
+        Encoding.PSEUDO_FENCE,
```
pmhahn, line 467: Adding this  here breaks the unit test:
```
FAIL: test_vncConnectionMade (test_client.TestVNCDoToolClient)
----------------------------------------------------------------------
  File "/home/runner/work/vncdotool/vncdotool/tests/unit/test_client.py", line 35, in test_vncConnectionMade
    self.client.setEncodings.assert_called_once_with([
  File "/opt/hostedtoolcache/Python/3.10.20/x64/lib/python3.10/unittest/mock.py", line 941, in assert_called_once_with
    return self.assert_called_with(*args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.10.20/x64/lib/python3.10/unittest/mock.py", line 929, in assert_called_with
    raise AssertionError(_error_message()) from cause
AssertionError: expected call not found.
Expected: mock([<Encoding.RAW: 0>, <Encoding.PSEUDO_CURSOR: -239>, <Encoding.PSEUDO_DESKTOP_SIZE: -223>, <Encoding.PSEUDO_LAST_RECT: -224>, <Encoding.PSEUDO_QEMU_EXTENDED_KEY_EVENT: -258>])
Actual: mock([<Encoding.RAW: 0>, <Encoding.PSEUDO_CURSOR: -239>, <Encoding.PSEUDO_DESKTOP_SIZE: -223>, <Encoding.PSEUDO_LAST_RECT: -224>, <Encoding.PSEUDO_QEMU_EXTENDED_KEY_EVENT: -258>, <Encoding.PSEUDO_FENCE: -312>])
```

Please add `client.rfb.Encoding.PSEUDO_FENCE,` to `tests/unit/test_client.py:41`

coder111111:
Hi, thanks for your comments, I'll try to fix this as soon as I have some time.

pmhahn:
Functional test is failing: `make test-func` – more exactly: all 3 tests from `tests/functional/test_proxy.py`.

For debugging: `PATH=$PATH:.vncdo/libvncserver-LibVNCServer-0.9.14/examples python3 -m unittest discover -s tests/functional -t . -p test_proxy.py -k test_key_alpha -v`

If I remember correctly `vncdotool/loggingproxy.py` functions as a pass-through proxy, e.g. all traffic between any client and server goes completely through it. As such it also must be able to understand/parse the new fence messages completely – handle the variable length payload – to pass the message from server-to-client and client-to-server.

I've lost most of my knowledge about this project in the last 2 years, so maybe @sibson can help and knows, where the proxy must be extended.

pmhahn:
> Functional test is failing: `make test-func` – more exactly: all 3 tests from `tests/functional/test_proxy.py`.

```diff
diff --git a/vncdotool/loggingproxy.py b/vncdotool/loggingproxy.py
index 37b0177..c51ec7d 100644
--- a/vncdotool/loggingproxy.py
+++ b/vncdotool/loggingproxy.py
@@ -171,6 +171,8 @@ class RFBServer(Protocol):  # type: ignore[misc]
             else:
                 log.debug("Unhandled subtype %r", QemuClientMessage.lookup(subtype))
                 raise ProtocolError(subtype)
+        elif ptype == MsgC2S.CLIENT_FENCE:
+            pass
         else:
             log.debug("Unhandled response %r", MsgC2S.lookup(ptype))
             raise ProtocolError(ptype)
@@ -342,6 +344,7 @@ class VNCLoggingServerFactory(portforward.ProxyFactory):  # type: ignore[misc]
     pseudodesktop = True
     qemu_extended_key = True
     last_rect = True
+    fence = False
     force_caps = False
 
     password_required = False
```
The 2nd hunk is the missing one; the 1st one should be the location, where proper handling for `CLIENT_FENCE` should go in if FENCE should be handled, too.

pmhahn (changes requested):
`vncdotool/loggingproxy.py` at least needs `fence = False`; more work if `True`.

sibson:
🤖 Claude Code:

Thank you for this, and for the detailed report in #322.

Closing in favour of #471, which is built on this implementation and carries your authorship in the changelog. The rebase onto the current `main` was large enough that a fresh branch was cleaner than updating this one: `rfb.py` has since split its constants into `const.py` and its decoders into `decoders/`, and `loggingproxy.py` grew the message-length table that @pmhahn's review here correctly pointed at. #471 also picks up the `FenceFlags` `IntFlag` and the `test_client.py` fix from that review.

One thing worth knowing, written up in full on #322: the fence traffic your patch handles does not appear to be what hung your session. Neither TigerVNC 1.12.0 nor 1.15.0 sends a `ServerFence` to a client that never offered `PSEUDO_FENCE`, which vncdotool did not. So #471 is a real fix — offering the encoding without answering the message would break every TigerVNC session — but it may not be a fix for yours. Asked over there for a capture so the actual cause can be found.

PR #331, opened 2026-07-29
Move constants into its own module

pmhahn:
While doing a review on #323 I stumbled over several _magic numbers_  in `vncdotool/rfb.py`, which really are _client-to-server message identifiers_. We already have `MsgC2S` and `MsgS2C` and others defined, but hidden in several different files. Adding additional cross-`import` statement to get them would lead to `import` cycles. As such move them into one common module `const` and let the old locations import from there.

`KEY_…` is more tricky as it declared all those constants as top-level symbols inside the module `vncdotool.rfb`. I have converted them into an `enum.IntEnum` and added `__getattr__()` implementation for backward compatibility. Accessing those symbols via `vncdotool.rfb.KEY_X` still works – you will get a deprecation warning – but `dir(vncdotool.rfb)` will **not** list `KEY_…`s anymore.


sibson:
For a long time I have avoided touching rfb.py in the hope of unifying with upstream.  I don't see that happening so I think this cleanup is overdue.

PR #334, opened 2026-08-13
Spike: Docker Compose VNC test servers for compatibility testing

sibson:
## What this is

Viability spike for **Tier 1** of the server compatibility plan: a Docker-Compose-defined set of Linux VNC servers usable identically for local dev and CI. Verdict: **viable** — green end to end (~70s: image builds + healthcheck-gated startup, all per-server tests pass, screenshots captured and published, clean teardown).

## Contents

- `tests/servers/Dockerfile` — one multi-stage build: a `base` stage with what every server shares (X client to draw, `EXPOSE`, `HEALTHCHECK`) plus a stage per server, so adding a server is a stage rather than another near-copy. Built on `debian:bookworm-slim` so versions come from apt in a distro we pin, not from whatever an upstream desktop image bundles.
- `tests/servers/docker-compose.yml` — services `tigervnc` (no auth, :5931), `tigervnc-auth` (VNC password, :5932), `x11vnc` (Xvfb-backed, :5933), each selecting its stage with `target:`. The two TigerVNC services are the same image, distinguished only by `VNC_PASSWORD`.
- `tests/servers/{tigervnc,x11vnc}-entrypoint.sh` + shared `draw-content.sh` — the TigerVNC entrypoint serves `SecurityTypes None` or VncAuth depending on `VNC_PASSWORD`; `draw-content.sh` draws an X client and signals readiness once its window is mapped.
- `tests/servers/servers.mk` (included from `Makefile`) — `servers-up`, `servers-down`, `test-servers`, `screenshots`
- `tests/functional/test_servers.py` — parameterized connect → type → capture test per server; skips cleanly when a server port isn&#39;t open
- `tests/functional/capture_screenshots.py` — screenshots every running server and builds the gallery; used by both `make screenshots` and CI
- `.github/workflows/spike-servers.yml` — branch-push + manual-dispatch workflow proving the servers on `ubuntu-latest`

## Screenshots are kept, not thrown away

- captures land in `tests/servers/screenshots/` (gitignored, `VNCDOTOOL_SCREENSHOT_DIR` overrides) instead of a tempfile, so they survive the run
- the capture script writes a self-contained `index.html` gallery with every server&#39;s screenshot inlined — one file to open, locally or out of the CI artifact
- in CI it also appends a server/port/resolution/size table to the job summary, so the run page shows what was captured without downloading the artifact
- `make screenshots` does all of the above locally against whatever servers are up

## What the per-server test proves

Passing means, in order: the RFB handshake completed for that server&#39;s security type (None, or VNC password auth), a key event was accepted, a framebuffer update was decoded to PNG, and that PNG is both the size the server serves and not a single flat colour. Previously an all-black framebuffer passed on PNG magic alone.

## Findings worth keeping

- `api.connect()` leaves the Twisted reactor thread running until `api.shutdown()` is called — without it, test processes/CI steps hang forever *after passing*. Hit twice during the spike; now guarded with `tearDownModule` and an explicit shutdown in the capture script. Also fresh evidence for the plan&#39;s Phase 1 &#34;fail loudly, never hang&#34; workstream.
- **An open RFB port is not readiness.** Xvnc accepts connections the instant it launches, before anything is drawn, so a port-only healthcheck reports healthy against a blank framebuffer — a latent race that only surfaced once the test started asserting captures aren&#39;t flat. Readiness now also requires the drawn-content marker.
- Debian/Ubuntu package split: `vncpasswd` lives in `tigervnc-tools`, needed alongside `tigervnc-standalone-server`.
- Runners start with an empty Docker cache, so image builds are paid in full every run (~35s of ~70s) — layer caching only helps within a run.

## Spike status

This is spike-quality code proving the approach. Follow-ups are recorded in `docs/server-compatibility-plan.md`: publish the screenshot gallery to GitHub Pages, stop paying the per-run image build tax (gha cache or GHCR), pin images by digest, fold into the main CI workflow, extend coverage (TightVNC, QEMU, TLS variants), deepen what the scenario suite asserts, and wire captures into the fixture/recording pipeline.


.github/workflows/spike-fleet.yml
```diff
@@ -0,0 +1,103 @@
...
+      - name: Show fleet status
+        if: always()
+        run: docker compose -f tests/servers/docker-compose.yml ps
+
+      - name: Run fleet functional tests
+        run: |
+          mkdir -p fleet-screenshots
+          python -m unittest discover -v -s tests/functional -t . -p 'test_fleet.py'
+
+      - name: Capture fleet screenshots for the artifact upload
```
sibson: For readability, this should be a script rather than inline code

.github/workflows/spike-fleet.yml
```diff
@@ -0,0 +1,103 @@
+name: Spike - VNC server fleet
```
sibson: Not keen on the name fleet.

.github/workflows/spike-fleet.yml
sibson: Moved out to `tests/functional/capture_screenshots.py`; every `run:` in the workflow is a one-liner now. The script does a bit more than the old heredoc — it also builds the `index.html` gallery and writes the job summary table — and it's reusable via `make screenshots`.

---

.github/workflows/spike-fleet.yml
sibson: Dropped. It's "test servers" throughout now: `spike-servers.yml`, `tests/servers/servers.mk`, `tests/functional/test_servers.py`, `make test-servers`, and `VNC_SERVERS`/`VNCServer` in code. Stale references in `docs/server-compatibility-plan.md` updated too.

---

.github/workflows/spike-servers.yml
```diff
@@ -0,0 +1,72 @@
...
+        if: always()
+        timeout-minutes: 3
+        run: python tests/functional/capture_screenshots.py
+
+      - name: Dump test server container logs
+        if: always()
+        run: docker compose -f tests/servers/docker-compose.yml logs
+
+      - name: Upload screenshots
+        if: always()
```
sibson, line 69: why if always()?

tests/servers/tigervnc-auth/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+# is never baked into the image.
+FROM debian:bookworm-slim
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        tigervnc-standalone-server \
+        tigervnc-common \
+        tigervnc-tools \
+        x11-apps \
+        xauth \
+        netcat-openbsd \
```
sibson: why is netcat needed?

tests/servers/tigervnc/Dockerfile
```diff
@@ -0,0 +1,25 @@
+# Minimal TigerVNC (Xvnc) server, no authentication.
```
sibson: can we layer these dockerfiles to avoid duplication?

tests/servers/tigervnc/entrypoint.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
```
sibson: can we not use a single entrypoint that optionally spawns with password?

tests/servers/tigervnc/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+#
+# Deliberately built from a small in-repo Dockerfile rather than a
+# third-party desktop image so the exact server version is controlled by
+# apt in the base distro rather than by whatever a random upstream image
+# happens to bundle.
+FROM debian:bookworm-slim
```
sibson: will gh actions cache these custom images or are we paying a build tax everytime.  Address as todo rather than blocker.

Makefile
```diff
@@ -7,6 +7,9 @@ REQUIREMENTS_TXT?=requirements-dev.txt
 help:
 	@echo "test:		run unit tests"
 	@echo "test-func:	run functional tests"
+	@echo "servers-up:	start the docker VNC test servers"
```
sibson, line 10: servers-down?

.github/workflows/spike-servers.yml
sibson: Because those trailing steps are diagnostics and teardown, and the case you most want them in is the failing one: when a test fails, the container status, a screenshot of what the server was actually showing, and the server logs are the evidence you need — without `if: always()` a failed test step skips all of it, and `down` would leak containers. Added a comment in the workflow saying so.

The steps that are part of the test itself (start servers, run tests) deliberately don't have it, so a failure there still fails the job.

---

tests/servers/tigervnc-auth/Dockerfile
sibson: It was only there for the `HEALTHCHECK`/compose readiness probe (`nc -z 127.0.0.1 5900`), which is what makes `up --wait` a reliable barrier. Not worth a package: switched to `bash -c 'exec 3<>/dev/tcp/127.0.0.1/5900'`, same probe using bash's built-in, and netcat is gone from the images.

---

tests/servers/tigervnc/Dockerfile
sibson: Done — the three Dockerfiles are now one multi-stage `tests/servers/Dockerfile`: a `base` stage with what every server shares (X client to draw, `EXPOSE`, `HEALTHCHECK`), then a stage per server that installs only its own packages. Compose picks the stage with `target:`. Adding a server is one stage rather than another near-copy.

---

tests/servers/tigervnc/entrypoint.sh
sibson: Yes — merged. `tigervnc-entrypoint.sh` reads `VNC_PASSWORD`: unset/empty gives `-SecurityTypes None`, set writes a password file with `vncpasswd` and gives `-SecurityTypes VncAuth -PasswordFile ...`. The `tigervnc-auth` image is gone; both services are the same image (pinned to one build via `image:` in compose) and differ only by that env var.

The wait-for-X-then-draw block both entrypoints duplicated is now a shared `draw-content.sh` in the base image, so `x11vnc-entrypoint.sh` is down to three lines of actual work.

---

tests/servers/tigervnc/Dockerfile
sibson: We're paying it. GitHub-hosted runners start each job with an empty Docker cache, so layer caching only helps within a run — every run rebuilds from scratch, ~35s of the ~70s total. Recorded as a Tier 1 follow-up in `docs/server-compatibility-plan.md` with the two options: `docker/build-push-action` with `cache-from/to: type=gha`, or build once into GHCR and have CI pull pinned digests (which folds into the digest-pinning item). Not treating it as a blocker.

The layering change above helps a little in the meantime — the TigerVNC services now share one image instead of building two, so it's two builds per run rather than three, over a shared base stage.

---

Makefile
sibson: Missed it — added to `help`.

---

PR #335, opened 2026-08-13
Spike: live UltraVNC (Windows) and Apple Screen Sharing (macOS) on hosted runners

sibson:
## What this is

Viability spike for **Tier 2** of the server compatibility plan: can GitHub-hosted `windows-latest` / `macos-latest` runners host live OS-bound VNC servers for vncdotool to test against? Verdict: **viable on both**, with one macOS caveat. Final [spike-os-servers run](https://github.com/sibson/vncdotool/actions/runs/31730610001) is green on both jobs after four evidence-driven rounds. Single file added: `.github/workflows/spike-os-servers.yml`.

## Windows / UltraVNC — works end-to-end

- `choco install ultravnc`; `winvnc.exe` is at the fixed path `C:\Program Files\uvnc bvba\UltraVNC\winvnc.exe` (do not search `Program Files` recursively — minutes-slow on runner images)
- UltraVNC refuses **all** connections until a password is set, regardless of `AuthRequired` — no no-auth shortcut exists (server literally replies "Until a password is set, incoming connections cannot be accepted")
- The `ultravnc.ini` `passwd=` hex uses the classic VNC password-*file* obfuscation (DES with the fixed `vncauth.c` key, bit-reversed) — computed in a few lines of Python on the runner; note this is *not* the challenge-response transform
- Must run as a Windows service (`winvnc.exe -install` + `Start-Service`); `winvnc.exe -run` opens an interactive settings dialog instead of serving
- `vncdo type` + `capture` succeed with genuine, non-black desktop content → Windows supports real visual assertions

## macOS / Screen Sharing — works at the protocol level

- Create a dedicated user (`sysadminctl -addUser`), enable Remote Management via `kickstart`; Screen Sharing is socket-activated on :5900 out of the box
- Auth is **ARD/Diffie-Hellman (type 30)** with username+password — vncdotool's existing ARD support handles the full round trip (connect, auth, type, capture)
- Legacy VNC password is confirmed dead on modern macOS: kickstart accepts `-setvnclegacy`/`-setvncpw` silently but the server still demands ARD auth
- **Caveat**: the captured framebuffer is fully black — the runner has no rendered desktop. macOS jobs therefore validate protocol/auth/input, not pixels, pending a follow-up into whether a display can be attached
- Bonus robustness finding: with no username supplied, `rfb.py` falls into an interactive `input("username:")` prompt and dies with `EOFError` on a TTY-less runner — filed under the plan's Phase 1 fail-loudly items

## Spike status

Spike-quality proof, triggered only on pushes to this branch + manual dispatch. Graduation (per `docs/server-compatibility-plan.md`): both jobs move under the change-triggered/path-filtered policy — Windows as a full type+capture check, macOS with pixel assertions explicitly excluded — with passwords moved to repository secrets and the dedicated-user/service-mode setup retained.

---
_Generated by [Claude Code](https://claude.ai/code/session_01STWkReabCK9sJW2qQ45brM)_

### DIFF
--- a/.github/workflows/spike-os-servers.yml
+++ b/.github/workflows/spike-os-servers.yml
@@ -1,0 +1,294 @@
+name: Spike - OS-hosted VNC servers
+
+# Viability spike: can GitHub-hosted windows-latest / macos-latest runners host a
+# live, OS-bound VNC server (UltraVNC on Windows, Apple Screen Sharing / Remote
+# Management on macOS) that vncdotool can connect to, drive, and screenshot?
+#
+# The two jobs are fully independent - one OS failing must never take down the
+# other, so there is no `needs:`, no matrix, and each job carries its own
+# artifact uploads and diagnostics.
+
+on:
+  push:
+    branches: [claude/spike-os-servers]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  windows-ultravnc:
+    name: Windows - UltraVNC
+    runs-on: windows-latest
+    timeout-minutes: 20
+    continue-on-error: true
+    defaults:
+      run:
+        shell: pwsh
+    steps:
+      - name: Check out source
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.13'
+
+      - name: Search available UltraVNC choco packages (diagnostic)
+        run: choco search ultravnc --all-versions
+        continue-on-error: true
+
+      - name: Install UltraVNC via chocolatey
+        run: choco install ultravnc -y --no-progress
+
+      - name: Locate winvnc.exe
+        id: locate
+        run: |
+          $candidates = Get-ChildItem -Path 'C:\Program Files','C:\Program Files (x86)' -Recurse -Filter winvnc.exe -ErrorAction SilentlyContinue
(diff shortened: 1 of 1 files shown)


.github/workflows/spike-os-servers.yml
```diff
@@ -0,0 +1,425 @@
...
+# Viability spike: can GitHub-hosted windows-latest / macos-latest runners host a
+# live, OS-bound VNC server (UltraVNC on Windows, Apple Screen Sharing / Remote
+# Management on macOS) that vncdotool can connect to, drive, and screenshot?
+#
+# The two jobs are fully independent - one OS failing must never take down the
+# other, so there is no `needs:`, no matrix, and each job carries its own
+# artifact uploads and diagnostics.
+#
+# Round-1 findings baked into this version:
+#  - Recursively scanning C:\Program Files for winvnc.exe took 5+ minutes on the
```
sibson, line 12: too much omment here.  narrative should be in docs and linked here not inline

sibson:
🤖 Claude Code: Superseded — Tier 2 graduation work already merged to main: setup/diagnostic scripts under tests/servers/ultravnc/ and tests/servers/screen-sharing/, tests/functional/test_os_servers.py, and docs/server-compatibility-plan.md updated with this spike's findings. This raw spike workflow is no longer needed.

PR #345, opened 2026-08-14
Report vncdo failures accurately, with exit codes grouped by cause

sibson:
## What changed

`VNCDoCLIFactory` decided the exit status from transport teardown: `clientConnectionLost` mapped `ConnectionDone` to `done(0)`. A clean TCP close only proves the socket shut down, so any clean close that was not the end of a successful run still exited 0.

Two parts:

**1. Exit status now comes from the `factory.deferred` command chain**, which is what represents the requested work — `done(SUCCESS)` as its final callback, `factory.error` as its errback. `clientConnectionLost` always reports an error and never signals success. First outcome wins, keyed off `reactor.exit_status` (initialised to `None`, mapped to 1 at the `sys.exit` site), checked in `error()` too so the expected close after a completed run does not log CRITICAL.

**2. Exit codes now say what went wrong.** Single digits are for input we were given, tens and up for what happened on the wire, grouped by cause:

| Code | Meaning |
|---|---|
| 0 | all actions completed |
| 1 | unexpected error |
| 2 | bad command line or unknown action |
| 3 | authentication failed |
| 10 | could not connect: refused, unreachable, name lookup failed |
| 11 | connection closed before the actions finished |
| 20 | server spoke something we could not understand |
| 30 | an action failed, e.g. writing a capture to an unwritable path |
| 40 | `--timeout` elapsed |

Authentication sits at 3 rather than inside the protocol family: a wrong password is bad input from whoever invoked vncdo, the same class as a bad flag, not a fault out on the wire. It is also the outcome scripts branch on most.

Classification is by exception type, falling back to the status of the reporting call site, since the same generic exception means different things arriving from `clientConnectionFailed`, `clientConnectionLost`, or the command chain.

Protocol failures previously logged and dropped the connection, so they surfaced as connection losses reading `Connection was closed cleanly`. Six sites in `rfb.py` now go through a new `vncProtocolError` hook, mirroring `vncAuthFailed` — the default implementation still only logs, so library behaviour is unchanged, and `VNCDoToolClient` turns it into a `ProtocolError` on the factory deferred.

`CommandParseError` exits 2 rather than 1. That 1 was never chosen: `sys.exit()` returns 1 for any non-int argument and the line always passed the message text. Both sibling usage errors (bad flag, missing command) already exited 2.

## Why

Measured against a stub listener and the tigervnc-auth compose server:

| Scenario | main | this branch |
|---|---|---|
| Server accepts, hangs up without speaking RFB | exit 0 | exit 11 |
| Server hangs up after the version handshake | exit 0 | exit 11 |
| Server sends a non-RFB header | exit 0 | exit 20 |
| Bad password | exit 0 | exit 3 |
| `capture /nonexistent/dir/x.png` | hangs forever | exit 30 |
| `--timeout` with an expect that never matches | exit 10 | exit 40 |
| Connection refused / DNS failure | exit 10 | exit 10 |
| Unknown action | exit 1 | exit 2 |
| Normal run | exit 0 | exit 0 |

The hang is the same defect seen from the other side: the chain errbacked with nothing attached to handle it, so the `loseConnection` callback was skipped and nothing closed the transport. `--timeout` was the only escape.

Scripts and CI relying on vncdo's exit code got a false OK in every failing row above.

## Reviewer notes

- Every row in that table was verified live, not just unit tested; each also has a unit test.
- Exit codes were previously undocumented, so this establishes the contract: docs/usage.rst gains a table and a worked example, and a unit test pins the numeric values.
- Version bumped to 1.4.0 and the changelog entry is marked `[BREAKING]`, since exit codes that used to be 0 are now non-zero.
- Ordering works because `loseConnection` is async: the `done(SUCCESS)` callback runs before `connectionLost` fires.
- `vnclog` is unaffected: `VNCLoggingServerFactory` extends `portforward.ProxyFactory`, has no `done`/`error`, and `build_proxy` sets `exit_status = 0` explicitly.
- The live `test_bad_password_exits_nonzero` check on the Phase 1 testing-framework branch is still marked `expectedFailure` there; that marker gets dropped once this lands beneath it. Note it should assert 3.

### DIFF
--- a/tests/unit/test_command.py
+++ b/tests/unit/test_command.py
@@ -2,7 +2,11 @@ import socket
 import unittest
 from unittest import mock, skipUnless
 
+from twisted.internet.error import ConnectionDone
+from twisted.python.failure import Failure
+
 from vncdotool import command
+from vncdotool.client import AuthenticationError
 
 
 class TestBuildCommandList(unittest.TestCase):
@@ -252,3 +256,21 @@ class TestVNCDoCLIClient(unittest.TestCase):
         assert command.getpass.getpass.called
         assert cli.factory.password == password
         cli.sendPassword.assert_called_once_with(password)
+
+
+class TestVNCDoCLIFactory(unittest.TestCase):
+
+    @mock.patch('vncdotool.command.reactor')
+    def test_auth_failure_exits_nonzero(self, reactor):
+        factory = command.VNCDoCLIFactory()
+        factory.clientConnectionFailed(
+            None, Failure(AuthenticationError('Authentication failure'))
+        )
+
+        assert reactor.exit_status == 10
+
+        # the server closing the rejected connection afterwards must not
+        # overwrite the failure exit status with a clean-close success
+        factory.clientConnectionLost(None, Failure(ConnectionDone()))
+
+        assert reactor.exit_status == 10
--- a/tests/unit/test_command.py
+++ b/tests/unit/test_command.py
@@ -260,17 +260,78 @@ class TestVNCDoCLIClient(unittest.TestCase):
 
 class TestVNCDoCLIFactory(unittest.TestCase):
 
-    @mock.patch('vncdotool.command.reactor')
-    def test_auth_failure_exits_nonzero(self, reactor):
-        factory = command.VNCDoCLIFactory()
-        factory.clientConnectionFailed(
+    def setUp(self) -> None:
+        patcher = mock.patch('vncdotool.command.reactor')
+        self.reactor = patcher.start()
+        self.addCleanup(patcher.stop)
+        self.reactor.exit_status = None
+        self.factory = command.VNCDoCLIFactory()
+
+    def test_auth_failure_exits_nonzero(self) -> None:
+        self.factory.clientConnectionFailed(
             None, Failure(AuthenticationError('Authentication failure'))
         )
 
-        assert reactor.exit_status == 10
+        assert self.reactor.exit_status == 10
 
-        # the server closing the rejected connection afterwards must not
-        # overwrite the failure exit status with a clean-close success
-        factory.clientConnectionLost(None, Failure(ConnectionDone()))
+    def test_clean_close_before_commands_run_exits_nonzero(self) -> None:
+        # the server hanging up cleanly is not evidence the commands ran
+        self.factory.clientConnectionLost(None, Failure(ConnectionDone()))
 
-        assert reactor.exit_status == 10
+        assert self.reactor.exit_status == 10
+
+    def test_clean_close_after_commands_run_keeps_success(self) -> None:
+        self.factory.done(0)
+        self.factory.clientConnectionLost(None, Failure(ConnectionDone()))
+
+        assert self.reactor.exit_status == 0
+
+    def test_first_outcome_wins(self) -> None:
+        self.factory.error(Failure(AuthenticationError('denied')))
+        self.factory.done(0)
+
+        assert self.reactor.exit_status == 10
+
+
+class TestBuildTool(unittest.TestCase):
+
+    def setUp(self) -> None:
+        patcher = mock.patch('vncdotool.command.reactor')
+        self.reactor = patcher.start()
+        self.addCleanup(patcher.stop)
+        self.reactor.exit_status = 'unset'
+
+        connect_patcher = mock.patch('vncdotool.command.factory_connect')
+        connect_patcher.start()
+        self.addCleanup(connect_patcher.stop)
+
+        self.options = mock.Mock(
+            verbose=False,
+            delay=None,
+            warp=1.0,
+            incremental_refreshes=0,
+            host='127.0.0.1',
+            port=5900,
+            address_family=socket.AF_INET,
+        )
+
+    def test_undecided_exit_status_starts_none(self) -> None:
+        command.build_tool(self.options, [])
+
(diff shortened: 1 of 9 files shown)


vncdotool/command.py
```diff
@@ -73,6 +74,11 @@ def error(self, reason: Failure) -> None:
         self.done(10)
 
     def done(self, exit_code: int) -> None:
+        # an authentication failure reports its error before the server
+        # closes the socket; the later clean close must not overwrite it
+        if self.finished:
```
sibson: this seems extranous.  Can we not key off exit_status != None?

sibson: We should maybe dig deeper into why done gets called twice and needs the guard.   It seems like maybe we're not handling correctly further up the stack.

sibson: Right on both counts, and the second one opened up more than I expected.

On keying off `exit_status`: it was never `None` to key off — `build_tool` set it to `1` before the reactor ran. Moved that to `None`, mapped `None → 1` at the `sys.exit` site in `vncdo()`, and dropped the `finished` flag. The guard now lives on `exit_status` in both `done()` and `error()`, the latter so the expected close after a completed run does not log CRITICAL.

On digging deeper: the guard was papering over the wrong signal. `clientConnectionLost` mapped `ConnectionDone` to success, but a clean close only proves the socket shut down. Measured against a stub listener and the tigervnc-auth container, on `main`:

| Scenario | main | now |
|---|---|---|
| Server accepts, hangs up without speaking RFB | exit 0 | exit 10 |
| Server hangs up after the version handshake | exit 0 | exit 10 |
| `capture /nonexistent/dir/x.png` | hangs forever | exit 10, reports the errno |
| Bad password | exit 0 | exit 10, `Authentication failure` |
| Normal run | exit 0 | exit 0 |

The hang is the same defect from the other side: the chain errbacked with nothing attached, so the `loseConnection` callback was skipped and nothing ever closed the transport. `--timeout` was the only escape.

Fix drives the exit status from the deferred chain, which is what actually represents the requested work — `done(0)` as its final callback, `error` as its errback — and `clientConnectionLost` now always reports an error. Ordering works out because `loseConnection` is async, so `done(0)` runs before `connectionLost` fires. The `--timeout` path at command.py:531 stops racing a completed run as a side effect.

In 8718f53, with unit tests for each row.

tests/unit/test_command.py
```diff
@@ -252,3 +256,115 @@ def test_vncRequestPassword_prompt(self, getpass):
         assert command.getpass.getpass.called
         assert cli.factory.password == password
         cli.sendPassword.assert_called_once_with(password)
+
+
+class TestVNCDoCLIFactory(unittest.TestCase):
+
+    def setUp(self) -> None:
+        patcher = mock.patch('vncdotool.command.reactor')
+        self.reactor = patcher.start()
```
sibson: Can we use decorators?

sibson (commented):
This change alters behaviour for protocol errors, so should be a minor version bump not a point

tests/unit/test_command.py
sibson: Done in 8027fa4 — class-level `mock.patch` decorators, no more `patcher.start()`/`addCleanup` pairs.

One wrinkle worth flagging: the stand-in is a small `FakeReactor` rather than the default `MagicMock`. A Mock auto-creates `exit_status` on attribute access, and the factory reads any non-`None` value as "an outcome was already decided", so every guard short-circuits and the tests pass while exercising nothing. That is why the old `setUp` had the `self.reactor.exit_status = None` line — the decorator form has nowhere to put it, so the fake owns the default instead.

Checked the tests still bite by breaking the classification mapping: `AuthenticationError` remapped to `TIMEOUT` fails 2 of the 9.

sibson:
Bumped to 1.4.0 in 8027fa4 — `vncdotool/__init__.py` to `1.4.0.dev0` and the CHANGELOG heading to match, so `make release` still finds `1.4.0 (UNRELEASED)` to stamp.

Agreed on the reasoning, and it is broader than the protocol errors: exit codes that used to be 0 are now non-zero for auth failures and truncated sessions, and `CommandParseError` moved from 1 to 2. Anything parsing vncdo exit codes can see different values after this, which is minor-bump territory on its own.

sibson:
Changelog now flags it, in f4ecd62 — `[BREAKING]` prefix matching the precedent from the 1.0 entry, spelling out the values that change: 21 for auth failure and 11 for a truncated session (both previously 0), and 2 for an unknown action (previously 1).

On preserving 1 for `CommandParseError`: I dug into whether that code was ever deliberate, and it was not. On main the three usage mistakes already disagree:

    vncdo --nosuchflag              -> 2   (optparse)
    vncdo -s HOST                   -> 2   (optparse, "no command provided")
    vncdo -s HOST nosuchcommand     -> 1   (CommandParseError)

So two of the three already exit 2 and only the unknown action exits 1. That 1 is a side effect rather than a choice: `sys.exit()` returns 1 for any non-int argument, and the line has always passed the message. 317de66 changed `sys.exit(exc)` to `sys.exit(str(exc))` as a typing cleanup, which left the code at 1 without anyone picking it.

Preserving it would mean an unknown flag exits 2 while an unknown action exits 1, and 1 would keep double duty as "unexpected error" in the new table. Happy to put it back if you would rather hold the line on the observable value, but the compatibility argument is weaker than it looks given the sibling cases were already 2.

PR #348, opened 2026-08-14
Run fleet smoke tests through the vncdo CLI; add event sinks; retire pexpect (Phase 1)

sibson:
Phase 1 of `docs/testing-framework-design.md` (merged in #347). Second of six stacked PRs.

## Hang containment: subprocess, not in-process

Every fleet scenario (connect, keypress, mousemove, capture) now shells out to the real `vncdo` CLI via `subprocess.run` with a timeout. No reactor ever starts in the test process, so there is no `api.shutdown()` ordering to get wrong, and any hang in the client, server, or network is contained by the kernel reaping the subprocess rather than by anything in-process.

## Verify input against event sinks, not pixels

- A new **vncev** compose service (LibVNCServer's event-printing example, built in-container from a pinned 0.9.14 tarball) answers *what did the client put on the wire*.
- An **`xev -root`** logger in the x11vnc container answers *did the server deliver a real X event*.

LibVNCServer's example server joins the smoke fleet as its own compose service, replacing the native `libvncserver.mk` build — no host toolchain needed anywhere, so `libvncserver.mk` and its CI job go away.

## pexpect retired

`test_screen.py`, `test_send_events.py`, the old `test_proxy.py`, `test_xvnc.py` and their harness (`cli.py`, `cli_runner.py`) are deleted. Their unique coverage is ported:

- CLI coverage → subprocess tests in `tests/functional/test_cli.py`
- the `vnclog` recording proxy → a subprocess `Popen` scenario in the new `test_proxy.py`, using `--forever` directory mode (the one recorder mode that flushes on client disconnect)

## Bug found while porting

`test_cli.py` pins a real client bug: `vncdo` exits 0 on a rejected VNC password, because `rfb.py`'s `vncAuthFailed` only logs. Kept as `expectedFailure` here — #345 is the fix, and drops the marker.

## Verification

Against the live five-service fleet: 27 functional tests OK (1 expected failure, 4 OS-server skips). Clean skips with the fleet down. `flake8 --count --statistics vncdotool tests` clean. Unit suite untouched and green (68 tests, 1 expected failure).

## Stack

1. ~~design + plan reconciliation~~ — #347, merged
2. **this PR** — fleet smoke through the `vncdo` CLI, event sinks, pexpect retired
3. Phase 2 — decoder golden unit tests
4. Phase 3 — `vnclog --capture`
5. Phase 4 — `replay_server.py`
6. Phase 5 — in-process API lifecycle suite


tests/functional/test_cli.py
```diff
@@ -0,0 +1,85 @@
+"""CLI-surface checks against the Docker Compose test fleet.
+
+The pexpect suite this replaces (tests/functional/test_screen.py,
```
sibson: We don't need to exaplin about pexpect

tests/functional/test_cli.py
```diff
@@ -0,0 +1,85 @@
...
+    HOST,
+    port_open,
+    run_vncdo,
+    screenshot_dir,
+)
+
+# tigervnc-auth is the one fleet server with a password, so it's the one
+# that can prove a *wrong* password is rejected rather than merely ignored.
+_AUTH_SERVER = next(s for s in DOCKER_SERVERS if s.name == "tigervnc-auth")
+_PLAIN_SERVER = next(s for s in DOCKER_SERVERS if s.name == "tigervnc")
```
sibson: do we really need a plainserver and auth? Aren't they eqivilent for this test?

.github/workflows/ci.yml
```diff
@@ -126,6 +92,8 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install -r requirements-dev.txt
+          # The functional suite shells out to `vncdo`, so it has to be on PATH.
```
sibson, line 95: much better comment

tests/functional/test_proxy.py
```diff
@@ -1,72 +1,80 @@
...
+    def _stop_proxy(self) -> None:
+        if self.proxy.poll() is None:
+            self.proxy.terminate()
+            try:
+                self.proxy.wait(timeout=PROXY_SHUTDOWN_TIMEOUT)
+            except subprocess.TimeoutExpired:
+                self.proxy.kill()
+                self.proxy.wait(timeout=PROXY_SHUTDOWN_TIMEOUT)
+
+    def test_keypress_is_recorded_and_forwarded(self) -> None:
```
sibson, line 67: Need handling for various classes of special keys?  Special keys is also an area we may need to have more live server tests against but that can be a todo.

PR #351, opened 2026-08-14
Add vnclog --capture: scrubbed raw wire captures for unhosted servers (Phase 3)

sibson:
Phase 3 of the server testing framework: the capture kit from `docs/testing-framework-design.md`.

`vnclog --capture DIR` records the raw server-to-client and client-to-server streams (`s2c.bin`, `c2s.bin`) alongside the semantic `session.vdo`, plus a `meta.json` fingerprint (server, versions, security types offered, geometry, what was scrubbed). Contributors running a server we cannot host in CI run one documented command and attach the directory to an issue. `docs/capture.rst` is the paved road.

## Scrubbing

Scrubbing happens before bytes touch disk. A handshake state machine in the new `vncdotool/capture.py` walks both directions of the proxied RFB handshake and replaces the 16-byte VNC-auth challenge and response with a same-length marker.

It branches on the version the *client* actually negotiates, not the server's greeting, so a client-side downgrade to 3.3 cannot route the scrubber off the real exchange. Auth types it cannot scrub (ARD/DH) are named in `meta.json`'s `unscrubbed_auth`, and `docs/capture.rst` warns that those captures contain encrypted credentials.

The capture tap runs before the proxy's own RFB parsers, and both are exception-isolated, so a parse failure can never truncate what reaches disk. One capture per directory: the first byte-bearing connection claims it, later connections are refused, and a zero-byte TCP open/close releases the claim.

## Found and fixed along the way

- **vnclog desynced against RFB 3.7+ servers using VNC password auth.** `RFBServer` never skipped the 16-byte auth response on the negotiated path, misread it as `ClientInit`, and lost the client stream. Real proxy bug, CHANGELOG entry included.
- **TigerVNC's brute-force blacklist counts failed auth attempts cumulatively per client IP** for the server process lifetime, so the suite's deliberate bad-password test poisoned `tigervnc-auth` for later runs — suite results depended on run history. The fleet now runs `Xvnc` with the blacklist disabled.
- **x11vnc raced its own Xvfb at container start** and exited 1 under load; the entrypoint now waits for the display, like the xev sink already did.
- **Capture-mode readiness is detected from vnclog's `accepting connections` stderr line, never by port-probing.** A probe races the forwarded server greeting into looking like a session and steals the one capture. A greeting-only session (server rejects before the client gets a byte through) is deliberately kept — that rejection is exactly the evidence a contributor is trying to capture.

## Verification

- `make test` — 117 tests, OK (expected failures=1)
- Functional suite against the docker fleet — 30 tests green, three consecutive runs

Stack note: Phase 2 (#350) is parked in draft. Its fixtures were hand-derived from the decoder source, so they pin behaviour rather than verify it independently. Phase 2 will be rebuilt on top of this PR and Phase 4, using real server-generated captures round-tripped against a source PNG.

### DIFF
--- a/docs/capture.rst
+++ b/docs/capture.rst
@@ -23,18 +23,13 @@ Making a capture
 
        vnclog --capture-raw ./my-bug-capture.zip --listen 5902 -s YOURSERVER::5900
 
-   ``vnclog`` will refuse to run if ``./my-bug-capture.zip`` already exists
-   -- that's deliberate, so a capture is never silently overwritten.
-
 3. Point your VNC client at the proxy instead of the real server --
    ``localhost::5902`` -- and reproduce the problem. This can be your usual
    GUI viewer, or a ``vncdo`` script if the bug is scriptable::
 
        vncdo -s localhost::5902 key enter type "hello" capture screen.png
 
-4. Stop the proxy (Ctrl-C) once you're done. It writes the archive when your
-   client disconnects, not only on a clean process exit, so a Ctrl-C is
-   safe.
+4. Disconnect. ``vnclog`` writes the archive and exits.
 
 5. Attach ``my-bug-capture.zip`` to the GitHub issue.
 
@@ -51,91 +46,34 @@ What's in the archive
                   # types it offered, which encodings it actually sent,
                   # and the framebuffer geometry
 
-The archive is written whole, under a temporary name and renamed into place,
-so a half-written capture never looks like a complete one.
-
-Only one session is ever recorded per ``--capture-raw``: if a second client
-connects (a viewer retrying after a dropped connection, say) it is refused
-outright rather than silently overwriting the capture you already have.
-Start a fresh ``vnclog --capture-raw`` with a new filename to record another
-session.
-
-``meta.json`` is plain, human-readable JSON -- open it and read it before
-you attach the capture, so you know what is about to leave your machine.
-
-``encodings_seen`` is counted from the rectangles the server actually sent,
-not from what the client asked for -- a server is free to ignore the
-client's ``SetEncodings`` -- so it answers "which encodings does this server
-really use?" directly::
-
-    "encodings_seen": [
-      {"encoding": 0,  "name": "raw",  "rectangles": 3},
-      {"encoding": 16, "name": "zrle", "rectangles": 118}
-    ]
-
-An encoding vncdotool has no name for shows as ``"name": null`` with its
-number intact, which is itself a useful bug report.
-
 What is scrubbed
 ------------------
 
-Credentials are redacted at capture time, before any byte touches disk.
-Redactions are always replaced by the same number of zero bytes, so the
-surrounding byte offsets don't shift and the capture still replays:
+Credentials are redacted at capture time, before any byte touches disk:
 
-* **VNC authentication** -- the 16-byte challenge (from the server) and the
-  16-byte response (from the client);
-* **ARD / Apple Screen Sharing** (``diffie-hellman``, macOS) -- the 128-byte
-  AES block carrying your username and password. The Diffie-Hellman values
-  around it (generator, modulus, both public keys) are public by
-  construction and are kept deliberately: ARD compatibility bugs live in
-  those values.
+* **VNC authentication** -- the 16-byte challenge and the 16-byte response;
+* **ARD / Apple Screen Sharing** -- the AES block carrying your username and
+  password. The Diffie-Hellman values around it are public by construction
+  and are kept: ARD compatibility bugs live in them.
 
-vncdotool can only redact an exchange it can parse. If the session
-negotiates an auth type it cannot -- ``tight``, ``vencrypt``, ``rsa-aes``,
-UltraVNC's MS-Logon -- the capture is aborted and the error tells you how to
-proceed anyway.
+vncdotool implements those two auth types and no others, so it cannot locate
+the secret inside ``tight``, ``vencrypt``, ``rsa-aes`` or UltraVNC's
+MS-Logon. A session negotiating one of those aborts the capture rather than
+write the credential exchange out whole; the error says how to override it.
 
 .. warning::
 
-   ``--capture-raw-unsafe-auth`` writes the credential exchange to the
-   archive verbatim, and ``meta.json``'s ``unscrubbed_auth`` records which
-   type it was. Some of these exchanges are attackable offline -- ARD's
-   512-bit Diffie-Hellman is within reach of ordinary compute today. Use a
-   **disposable password, and rotate it afterwards**, and prefer emailing
-   such a capture to a maintainer over attaching it to a public issue.
+   Overriding writes your credential exchange into the archive verbatim.
+   Some of these are attackable offline -- ARD's 512-bit Diffie-Hellman is
+   within reach of ordinary compute today. Use a **disposable password and
+   rotate it afterwards.**
 
 What is **not** scrubbed
 --------------------------
 
-Scrubbing stops at the end of the handshake. Everything the session itself
-carries is recorded as-is -- deliberately: a capture with its input stream
-rewritten is no longer evidence of what the server did. In particular:
+Scrubbing stops at the end of the handshake. Everything the session carries
+is recorded as sent: screen contents, clipboard text, and every keystroke --
+including anything typed into a password field *inside* the remote desktop.
 
-* **screen contents** -- framebuffer updates are the whole point of the
-  protocol, and they are not redacted;
-* **clipboard text** -- cut/paste text sent in either direction is captured
-  in the clear;
-* **typed text** -- keystrokes you send through the proxy (including
-  anything you type into a password field *inside* the remote session, as
-  opposed to the VNC connection password itself) are recorded exactly as
-  sent.
-
-Do not reproduce a bug through the capture proxy while entering anything you
-would not want to see attached to a public GitHub issue -- in particular, do
-not type a password into the remote desktop during a capture. If your
-reproduction requires it, unzip the archive, redact the relevant bytes from
-``s2c.bin``/``c2s.bin`` by hand, and rezip -- the format is deliberately
-simple (three flat files plus JSON metadata) so that's a plain byte-offset
-edit, not a special tool.
-
-Replaying a capture
----------------------
-
-Captures are discovery evidence attached to an issue, not a fixture that
-lives in the repository or runs in CI -- see ``docs/testing-framework-design.md``
-in the source tree for the full picture. A maintainer investigating your
-issue may replay ``s2c.bin`` at a real client using an in-repo development
-tool to narrow down the bug, and the end result of that investigation is a
-small, deterministic unit test with the relevant bytes inlined -- not the
-capture itself.
+So reproduce the bug and nothing else. Do not type a password into the
+remote desktop while capturing.
--- a/docs/capture.rst
+++ b/docs/capture.rst
@@ -1,0 +1,126 @@
+Capturing a session for a bug report
+=====================================
+
+vncdotool is tested against the servers we can run ourselves (see
+``tests/servers/``). If you hit a bug against a server we can't host --
+RealVNC, Proxmox, a vendor's embedded VNC implementation, whatever -- the
+most useful thing you can attach to the issue is a *capture*: a recording of
+the raw bytes vncdotool exchanged with that server while reproducing the
+problem.
+
+You do not need a checkout of this repository to make one. A released,
+``pip``-installed vncdotool is enough.
+
+Making a capture
+------------------
+
+1. Install vncdotool::
+
+       pip install vncdotool
+
+2. Start the recording proxy, pointed at the server you're seeing the bug
+   against, with an empty (or not-yet-existing) directory to write into::
+
+       vnclog --capture ./my-bug-capture --listen 5902 -s YOURSERVER::5900
+
+   ``vnclog`` will refuse to run if ``./my-bug-capture`` already exists and
+   is non-empty -- that's deliberate, so a capture is never silently mixed
+   with an older one.
+
+3. Point your VNC client at the proxy instead of the real server --
+   ``localhost::5902`` -- and reproduce the problem. This can be your usual
+   GUI viewer, or a ``vncdo`` script if the bug is scriptable::
+
+       vncdo -s localhost::5902 key enter type "hello" capture screen.png
+
+4. Stop the proxy (Ctrl-C) once you're done. It flushes and closes the
+   capture when your client disconnects, not only on a clean process exit,
+   so a Ctrl-C is safe.
+
+5. Attach the whole ``my-bug-capture`` directory to the GitHub issue (zip it
+   first if your attachment tool doesn't take directories).
+
+What's in the directory
+-------------------------
+
+::
+
+    session.vdo   # the vncdo commands/events driven through the proxy
+    s2c.bin       # raw server-to-client byte stream, in arrival order
+    c2s.bin       # raw client-to-server byte stream, in arrival order
+    meta.json     # server address, vncdotool version, capture timestamp,
+                  # the server's protocol version string, the security
+                  # types it offered, the framebuffer geometry, and what
+                  # was (and was not) scrubbed
+
+Only one session is ever recorded per ``--capture DIR``: if a second client
+connects (a viewer retrying after a dropped connection, say) it is refused
+outright rather than silently overwriting the capture you already have.
+Start a fresh ``vnclog --capture`` into a new directory to record another
+session.
+
+``meta.json`` is plain, human-readable JSON -- open it and read it before
+you attach the capture. It tells you exactly what got redacted, so you can
+confirm nothing else in the two ``.bin`` files needs to be trimmed by hand
+before it leaves your machine.
+
+What is scrubbed
+------------------
+
+Two things are redacted at capture time, before any byte touches disk:
+
+* the 16-byte VNC authentication challenge (sent by the server) and the
+  16-byte response (sent by the client), each replaced with an all-zero
+  marker of the same length so the surrounding byte offsets don't shift;
+* any bytes matching the VNC password vncdotool itself was given (via
+  ``-p``/``--password``), wherever they occur in either stream.
+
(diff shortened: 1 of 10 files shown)


docs/capture.rst
```diff
@@ -0,0 +1,166 @@
...
+
+* **VNC authentication** -- the 16-byte challenge (from the server) and the
+  16-byte response (from the client);
+* **ARD / Apple Screen Sharing** (``diffie-hellman``, macOS) -- the 128-byte
+  AES block carrying your username and password. The Diffie-Hellman values
+  around it (generator, modulus, both public keys) are public by
+  construction and are kept deliberately: ARD compatibility bugs live in
+  those values;
+* **your VNC password** -- any bytes matching the password vncdotool itself
+  was given (via ``-p``/``--password``), wherever they occur in either
```
sibson: this is just pointless extra code for a weak guarenttee,  remove

docs/capture.rst
```diff
@@ -0,0 +1,166 @@
...
+  stream. Best-effort only: VNC auth never puts the password on the wire in
+  the clear, so this is a backstop, not the main defence.
+
+Each is listed by name in ``meta.json``'s ``scrubbed`` field, so you don't
```
sibson: pointless, proves nothing

docs/capture.rst
```diff
@@ -0,0 +1,166 @@
...
+Auth types with no scrubber
+-----------------------------
+
+vncdotool can only redact an exchange it can parse. If the session
+negotiates something else -- ``tight``, ``vencrypt``, ``rsa-aes``,
+UltraVNC's MS-Logon -- the capture is **aborted**: nothing is written and
+the connection is closed, with the reason printed to stderr::
+
+    --capture aborted: the session negotiated tight(16), which vncdotool
+    cannot scrub; the capture would contain the credential exchange verbatim.
```
sibson: error should give clue as to how to get around it.  ALso, don't need to docuement this behaviour

docs/capture.rst
```diff
@@ -0,0 +1,141 @@
...
+
+vncdotool is tested against the servers we can run ourselves (see
+``tests/servers/``). If you hit a bug against a server we can't host --
+RealVNC, Proxmox, a vendor's embedded VNC implementation, whatever -- the
+most useful thing you can attach to the issue is a *capture*: a recording of
+the raw bytes vncdotool exchanged with that server while reproducing the
+problem.
+
+You do not need a checkout of this repository to make one. A released,
+``pip``-installed vncdotool is enough.
```
sibson, line 12: strike overkill

docs/capture.rst
```diff
@@ -0,0 +1,141 @@
...
+   -- that's deliberate, so a capture is never silently overwritten.
+
+3. Point your VNC client at the proxy instead of the real server --
+   ``localhost::5902`` -- and reproduce the problem. This can be your usual
+   GUI viewer, or a ``vncdo`` script if the bug is scriptable::
+
+       vncdo -s localhost::5902 key enter type "hello" capture screen.png
+
+4. Stop the proxy (Ctrl-C) once you're done. It writes the archive when your
+   client disconnects, not only on a clean process exit, so a Ctrl-C is
```
sibson: we're telling them to Ctrl-z, saying it's safe is redundent

docs/capture.rst
```diff
@@ -0,0 +1,141 @@
...
+
+    session.vdo   # the vncdo commands/events driven through the proxy
+    s2c.bin       # raw server-to-client byte stream, in arrival order
+    c2s.bin       # raw client-to-server byte stream, in arrival order
+    meta.json     # server address, vncdotool version, capture timestamp,
+                  # the server's protocol version string, the security
+                  # types it offered, which encodings it actually sent,
+                  # and the framebuffer geometry
+
+The archive is written whole, under a temporary name and renamed into place,
```
sibson: strike, implementation detail users don't care about.  This would be in the tech spec not user facing docs.

docs/capture.rst
```diff
@@ -0,0 +1,141 @@
...
+so a half-written capture never looks like a complete one.
+
+Only one session is ever recorded per ``--capture-raw``: if a second client
+connects (a viewer retrying after a dropped connection, say) it is refused
```
sibson: users can discover this behaviour and it should be obvious what happens next.  again belongs in tech spec not user docs.

docs/capture.rst
```diff
@@ -0,0 +1,141 @@
...
+outright rather than silently overwriting the capture you already have.
+Start a fresh ``vnclog --capture-raw`` with a new filename to record another
+session.
+
+``meta.json`` is plain, human-readable JSON -- open it and read it before
```
sibson: we're telling people to open a zip and inspect meta.json,  that's insane.

docs/capture.rst
```diff
@@ -0,0 +1,141 @@
...
+
+* **VNC authentication** -- the 16-byte challenge (from the server) and the
+  16-byte response (from the client);
+* **ARD / Apple Screen Sharing** (``diffie-hellman``, macOS) -- the 128-byte
+  AES block carrying your username and password. The Diffie-Hellman values
+  around it (generator, modulus, both public keys) are public by
+  construction and are kept deliberately: ARD compatibility bugs live in
+  those values.
+
+vncdotool can only redact an exchange it can parse. If the session
```
sibson: why can't we parse these types?

docs/capture.rst
```diff
@@ -0,0 +1,141 @@
...
+negotiates an auth type it cannot -- ``tight``, ``vencrypt``, ``rsa-aes``,
+UltraVNC's MS-Logon -- the capture is aborted and the error tells you how to
+proceed anyway.
+
+.. warning::
+
+   ``--capture-raw-unsafe-auth`` writes the credential exchange to the
+   archive verbatim, and ``meta.json``'s ``unscrubbed_auth`` records which
```
sibson: we dropped this key

docs/capture.rst
```diff
@@ -0,0 +1,141 @@
...
+   type it was. Some of these exchanges are attackable offline -- ARD's
+   512-bit Diffie-Hellman is within reach of ordinary compute today. Use a
+   **disposable password, and rotate it afterwards**, and prefer emailing
```
sibson: we don't provide email support and I don't want to handle thes via my inbox

docs/capture.rst
```diff
@@ -0,0 +1,141 @@
...
+* **clipboard text** -- cut/paste text sent in either direction is captured
+  in the clear;
+* **typed text** -- keystrokes you send through the proxy (including
+  anything you type into a password field *inside* the remote session, as
+  opposed to the VNC connection password itself) are recorded exactly as
+  sent.
+
+Do not reproduce a bug through the capture proxy while entering anything you
+would not want to see attached to a public GitHub issue -- in particular, do
+not type a password into the remote desktop during a capture. If your
```
sibson: this is insane to ask someone to do and will more than likely breka the capture

tests/functional/test_proxy.py
```diff
@@ -3,21 +3,90 @@
 A `vnclog` process sits between `vncdo` and the vncev fleet service, and
 the test asserts both sides of the proxy did their job: the events reached
 the server and the recorder wrote them down.
+
+The `--capture` classes run against vncev (auth None) and against
```
sibson: why is this a top level doc comment rather than closer to the relevent test.  Also seems like we're over explaining again

tests/functional/test_proxy.py
```diff
@@ -78,3 +147,131 @@ def test_keypress_is_recorded_and_forwarded(self) -> None:
             time.sleep(0.2)
         self.assertIn("keydown z", recorded, f"vnclog recorded:\n{recorded!r}")
         self.assertIn("keyup z", recorded, f"vnclog recorded:\n{recorded!r}")
+
+
+class TestVnclogCapture(TestCase):
+    """`vnclog --capture-raw FILE.zip` end to end against vncev.
+
+    vncev speaks auth None, so this is the no-scrub path: the archive lands
+    with all four members and stays byte-identical when there is nothing to
```
sibson: nothing to redact? we don't redact anything

tests/functional/test_proxy.py
```diff
@@ -78,3 +147,131 @@ def test_keypress_is_recorded_and_forwarded(self) -> None:
...
+    redact.
+    """
+
+    def setUp(self) -> None:
+        if not port_open(HOST, VNCEV.port):
+            self.skipTest(
```
sibson, line 147: this allows silent failure to sneak into the test suite

tests/functional/test_proxy.py
```diff
@@ -78,3 +147,131 @@ def test_keypress_is_recorded_and_forwarded(self) -> None:
...
+            ],
+            capture_output=True,
+            text=True,
+            timeout=PROXY_STARTUP_DEADLINE,
+        )
+        self.assertNotEqual(second.returncode, 0)
+        self.assertIn("already exists", second.stderr.lower())
+
+
+class TestVnclogCaptureVncAuth(TestCase):
```
sibson: CaptureVNCAuth because VNC is an acroymn we shouldn't mix case it

vncdotool/capture.py
```diff
@@ -0,0 +1,326 @@
...
+
+from .const import AuthTypes, Encoding
+
+MARKER = b"\x00" * 16
+
+# AES-ECB over a fixed 64-byte username + 64-byte password struct; see
+# RFBClient._encryptArd().
+ARD_CREDENTIALS_LEN = 128
+
+Direction = str  # "s2c" or "c2s"
```
sibson: feels like a magic value

vncdotool/capture.py
```diff
@@ -0,0 +1,326 @@
...
+
+
+def check_capture_target(path: str) -> None:
+    """Validate `path` is usable as a --capture-raw target.
+
+    Creates nothing: the archive is written when the session ends, so a
+    later `op.error()` leaves nothing behind.
+    """
+    if not path.endswith(".zip"):
+        raise ValueError(f"--capture-raw target {path!r} must end in .zip -- captures are written as one archive")
```
sibson, line 348: is there a test for this?

vncdotool/command.py
```diff
@@ -320,6 +321,7 @@ def build_tool(options: optparse.Values, args: list[str]) -> VNCDoCLIFactory:
 def build_proxy(options: optparse.Values) -> VNCLoggingServerFactory:
     factory = VNCLoggingServerFactory(options.host, int(options.port))
     factory.password_required = options.password_required
+    factory.server_str = options.server
```
sibson: what is a server_str?  can we find a better name

vncdotool/command.py
```diff
@@ -449,15 +451,41 @@ def vnclog() -> None:
...
 
     setup_logging(options)
 
     options.address_family, options.host, options.port = parse_server(options.server)
 
-    if len(args) != 1:
+    output = None
+    if options.capture_raw:
+        if args:
+            op.error("OUTPUT is implied by --capture-raw (session.vdo in the archive); do not also pass OUTPUT")
```
sibson, line 489: should argparser not already handle this?

vncdotool/command.py
```diff
@@ -449,15 +451,41 @@ def vnclog() -> None:
...
+        try:
+            check_capture_target(options.capture_raw)
+        except ValueError as exc:
+            op.error(str(exc))
+    elif options.capture_raw_unsafe_auth:
```
sibson, line 494: does this skip the existing zip check?

vncdotool/loggingproxy.py
```diff
@@ -72,7 +75,14 @@ def _handle_security(self) -> None:
         sectype = self.buffer[0]
         log.debug("Client selected %r", AuthTypes.lookup(sectype))
         del self.buffer[:1]
-        self._handler = self._handle_clientInit, 1
+        if sectype == AuthTypes.VNC_AUTHENTICATION:
```
sibson, line 79: why was this added?

vncdotool/loggingproxy.py
```diff
@@ -234,22 +272,100 @@ class VNCLoggingServerProxy(portforward.ProxyServer, RFBServer):  # type: ignore
...
+            if self.capture.s2c or self.capture.c2s:
+                # Bytes in either direction count: a server rejecting the
+                # client before it speaks is itself the evidence. So nothing
+                # may port-probe a capture-mode vnclog -- the forwarded
+                # greeting would look like a session and claim the capture.
+                self._writeCapture()
+            else:
+                # A bare TCP open/close is not a session. Release the claim
+                # rather than lock out the real connection behind it.
+                log.debug("--capture-raw: connection produced no bytes, not claiming the capture")
```
sibson: claiming is weird.  What is the purpose

vncdotool/loggingproxy.py
```diff
@@ -234,22 +272,100 @@ class VNCLoggingServerProxy(portforward.ProxyServer, RFBServer):  # type: ignore
...
+        if self.capture is not None:
+            self.capture.feed_c2s(data)
+            if self.capture.abort_reason is not None:
+                self._abortCapture(self.capture.abort_reason)
+                return
+        try:
+            RFBServer.dataReceived(self, data)
+        except Exception:
+            log.warning(
+                "RFBServer failed to parse %d client bytes; capture and forwarding continue",
```
sibson: this changes client behaviour, we no longer propogate exceptions?

vncdotool/loggingproxy.py
```diff
@@ -299,23 +415,45 @@ class VNCLoggingServerFactory(portforward.ProxyFactory):  # type: ignore[misc]
...
     password_required = False
+    # Declared so a factory built without the CLI parser still has a value
+    # here rather than raising AttributeError.
+    password: str | None = None
 
     output: IO[str] | str = sys.stdout
     _out: IO[str] | None = None
 
+    # `capture_claimed` latches on the first connection accepted in capture
+    # mode, so a later one is refused rather than clobbering the capture.
```
sibson: would it be better to have a --one-shot option or something similar that we could set when in capture mode? And be useable elsewhere?  one-shot would exit cleanly after a single connection.

docs/capture.rst
sibson: vncdotool only implements NONE, VNC auth and Diffie-Hellman (rfb.py:155-157). Scrubbing needs the exchange structure to know where the secret starts and where the handshake resumes, and we have neither for tight/vencrypt/rsa-aes/MS-Logon. The proxy still forwards them fine, it just cannot find the secret inside. Docs now say this in one sentence.

docs/capture.rst
sibson: unscrubbed_auth was still there -- `scrubbed` was the one dropped. But you are right that it should go: it is the same self-report, and with abort-by-default it is only non-null when the user typed the opt-in themselves. Dropped from meta.json; the name survives only as an internal attribute used to build the abort message.

vncdotool/capture.py
sibson: Yes: test_refuses_non_zip_suffix for this branch, plus test_refuses_existing_file and test_refuses_missing_parent_directory for the other two.

vncdotool/command.py
sibson: vnclog uses optparse, not argparse. Neither can express "this option consumes the positional argument" -- argparse mutually-exclusive groups only cover options, not option-vs-positional -- so it stays a manual check either way.

vncdotool/command.py
sibson: No. That elif is only reachable when --capture-raw is absent, so there is no target to validate; check_capture_target runs in the if branch above it.

vncdotool/loggingproxy.py
sibson: That is the bug this PR fixes: RFBServer never skipped the 16-byte auth response on the 3.7+ negotiated path, read its first byte as ClientInit shared, and lost the client stream from there. Covered by test_vnc_auth_over_negotiated_security_does_not_desync_rfbserver and the live tigervnc-auth functional test, with a CHANGELOG entry.

vncdotool/loggingproxy.py
sibson: Purpose was: refuse a second connection rather than let it overwrite or interleave into the one capture. Replaced by --one-shot per your other comment, so the concept is now "this run serves one session" rather than anything capture-specific. The latch survives as session_taken, only to refuse a concurrent second connection while the session is live.

vncdotool/loggingproxy.py
sibson: Good catch, and the answer differs per call. The two wrapped calls are not the same thing: RFBServer.dataReceived / vnclog.dataReceived are observers feeding our own log, while super().dataReceived is the forwarding the real client depends on. Swallowing an observer error cannot make the client survive anything -- but swallowing a forwarding error leaves it waiting on bytes that never arrive. So forwarding is now unwrapped and propagates as before; observers log and continue. Note the pre-change behaviour was itself wrong for capture: an unparseable message raised out of the *logger* and tore down a session the server was happy with, recording a vnclog artefact. New test: test_forwarding_failures_are_not_swallowed.

tests/functional/test_proxy.py
sibson: Agreed, and it bit me while working on this -- a local run showed 31/31 OK with the fleet stopped. Leaving it for its own PR since the skip pattern predates this branch and applies to every functional module (Phase 0/1). Suggest an env var CI sets that turns unreachable servers into failures, keeping the local no-docker path.

tests/functional/test_proxy.py
sibson: Fair -- it was describing classes further down. Cut to one line; the per-class detail now sits on the class it describes.

tests/functional/test_proxy.py
sibson: Wrong on my part: vncev uses auth None so there is no auth exchange to redact in that particular test, but the phrasing read as if we never redact. Docstring is now one line naming the server and its auth type.

vncdotool/loggingproxy.py
sibson: Done -- added --one-shot, exits cleanly after one session, and --capture-raw implies it. It errors with --forever. A connection carrying no bytes still does not count, so a port probe cannot consume the run before the real client connects.

PR #352, opened 2026-08-14
Strip auth at capture time, and add vncdo-replay (Phase 4)

sibson:
Phase 4 of the server testing framework: the replay side of the capture kit. Review feedback reshaped both halves, so this now touches capture as well as replay.

## Auth is stripped at capture time, not redacted

`vnclog --capture-raw` no longer zeroes credentials in place. It replaces the handshake: the archive records the real greeting, a `none`-only security list, a SecurityResult where the negotiated version calls for one, and then the real stream from ServerInit on. The auth exchange between them is never written.

"This archive contains no credential bytes" is a stronger thing for a contributor to reason about than "the credential bytes are zeros", and it means the archive describes a session anyone can replay without a password — whatever the original server demanded.

Two consequences worth flagging:

- **The equal-length-zeroing invariant is gone**, and with it the per-step `scrub` flag. Which step a chunk belongs to is often only knowable *after* reading it — the security-type count and its list have to be re-emitted together as one `none` offer — so the grammar now decides what to record once it has seen the bytes.
- **Stripping still needs a grammar for the auth type**, because skipping the exchange means knowing where it ends. Types we have none for (tight, vencrypt, rsa-aes, MS-Logon) still abort the capture by default. Only the reason changed: not "we cannot find the secret" but "we cannot find the end of it".

`--capture-raw-unsafe` supersedes `--capture-raw-unsafe-auth`, removed rather than aliased since the kit is unreleased. It now means "preserve the handshake whole" rather than "permit an unscrubbable one" — which is also the only way to get ARD's Diffie-Hellman into an archive, since stripping takes it along with the credentials it surrounds.

## `vncdo-replay`, two commands

Shipped as a console script (`vncdotool/replay.py`, out of `tests/tools/`), so a contributor can replay their own capture before filing:

```
vncdo-replay --server ./my-bug-capture.zip     # one terminal
vncdo-replay ./my-bug-capture.zip              # another
```

Client mode hands the archive's own `session.vdo` to `vncdo`, defaulting `-s` to the replay server's `127.0.0.1::5999`, with trailing arguments appended as further `vncdo` commands (`vncdo-replay capture.zip capture screen.png`). Replaying the recorded events is the point: a replay driven by different events is a different session, and a replay of a different session is not evidence.

An earlier revision of this branch had `--server` fork its own client so one command reproduced everything. Serving bytes and driving a session turned out to be different jobs, and fusing them made the tool awkward to point at a GUI viewer or at a `vncdo` line with an extra command appended. Nothing about the client side is privileged now.

Replay is smaller for the capture-side change: one protocol rather than two, no handshake substitution, and no security-type divergence check — a stripped archive offers `none` and cannot diverge, and an unsafe one is served as-is and left to desync if the live client chose differently. What remains is pacing the handshake against the client's real replies, using the same grammar the capture side walks, and holding the recorded framebuffer until the client's first FramebufferUpdateRequest.

## Two bugs a real client exposed

- **The framebuffer was firehosed before the client asked for it.** A capture holds one finite recording of the screen, so those bytes get exactly one chance to be useful; sent early they go past a client that asks a moment later, which then waits forever. This is what made `vncdo <script-file>` hang where the same commands inline worked — a file argument inserts vncdo's 10ms `--delay` pause ahead of the `capture` command, and 10ms was enough to miss the update.
- **An exhausted capture closed the connection**, which `vncdo` reported as `CONNECTION_LOST` (exit 11). The original server hung up because its client did; closing here cut the client's own work short.

## Verification

- `make test` — 152 tests, OK (expected failures=1)
- `tests/functional/{test_replay,test_roundtrip,test_proxy}.py` — 7 tests, OK

`test_roundtrip.py` is the one that matters: it captures password-protected tigervnc through `vnclog`, then replays that archive with no password at all, to pixel-identical output. `test_proxy.py` asserts against live tigervnc-auth that the recorded stream carries a `none` offer where the challenge used to be.

## Rebase note

The branch was written against the pre-merge Phase 3 and reworked onto what landed in #351: captures are a zip archive (`--capture-raw ARCHIVE.zip`), not a directory, and the scrubber is driven per-direction through `Tap` objects. Two stale `--capture DIR` references in `docs/testing-framework-design.md` are corrected here too.


sibson:
CI's `Test server functional tests` failure here is not from this branch — `test_keypress_seen_by_xev` failed because x11vnc's `xev` sink lost its startup race against Xvfb and was never restarted (`xev: unable to open display ':0'` is the first line of that run's container log). Root cause and fix in #353; this branch will be rebased on it once that lands.

tests/tools/replay_server.py
```diff
@@ -0,0 +1,566 @@
...
+--capture-raw`` archive; ``c2s.bin`` drives the divergence check below and
+``session.vdo``/``meta.json`` are for a human to read. ``--script FILE``
+runs a hand-written scenario
+instead: a Python file defining a ``MESSAGES`` list of ``bytes`` to send,
+``("wait", nbytes)`` to block until the client has sent that much, or
+``("pause", seconds)``. It is run through ``exec()`` -- **scripts are
+trusted developer code**, like a local config file. Either way, the
+connection is closed once there is nothing left to send.
+
+Pacing (``--capture`` only)
```
sibson: what is pacing, --capture is invalid

docs/capture.rst
```diff
@@ -77,3 +77,63 @@ including anything typed into a password field *inside* the remote desktop.
...
+
+``tests/tools/replay_server.py`` serves a capture back at a real VNC client,
+so a bug against a server nobody here can run becomes reproducible on a
+laptop. It is a maintainer's tool: never shipped, never a CI fixture. The
+end product of a replay session is a unit test with the relevant bytes
+inlined, not the capture itself.
+
+::
+
+    python tests/tools/replay_server.py --capture ./my-bug-capture.zip --verbose
```
sibson: --capture is required so should be an argument not param

docs/capture.rst
```diff
@@ -77,3 +77,63 @@ including anything typed into a password field *inside* the remote desktop.
...
+
+then point a client at it::
+
+    vncdo -s 127.0.0.1::5999 key enter type "hello" capture screen.png
```
sibson: vncdo must run the vdo file in the capture otherwise we're not being faithful.  This may imply we need replay_capture to fork vncdo rather than allow user to run.

tests/tools/replay_server.py
```diff
@@ -0,0 +1,405 @@
+#!/usr/bin/env python3
+"""Replay a captured (or hand-scripted) RFB byte stream at a real VNC client.
+
+A maintainer's local tool, never shipped and never a CI fixture. See
```
sibson: extra comments

tests/tools/replay_server.py
```diff
@@ -0,0 +1,405 @@
...
+                    return
+                chunk = bytes(self.buffer[:nbytes])
+                del self.buffer[:nbytes]
+                tap.feed(chunk)
+            if self._diverged():
+                return
+        self._send_remainder()
+
+    def _diverged(self) -> bool:
+        """Has the live client picked a different security type than the capture?
```
sibson: We should skip auth and use no auth mode of vnc unless auth was captured.

vncdotool/replay.py
```diff
@@ -0,0 +1,272 @@
...
+    def _stalled(self, nbytes: int) -> None:
+        self._stall_call = None
+        log.warning(
+            "timed out after %ss waiting for the client to send %d bytes (got %d) -- "
+            "client may have stalled",
+            self.factory.client_timeout, nbytes, len(self.buffer),
+        )
+
+
+class ReplayFactory(ServerFactory):  # type: ignore[misc]
```
sibson: why do we need this ignore

vncdotool/replay.py
```diff
@@ -0,0 +1,272 @@
...
+                self.awaiting_request = True
+        if self.awaiting_request and not saw_update_request(self.buffer):
+            self.expect(10)
+            return
+        remainder = s2c[self.pos :]
+        if remainder:
+            self.transport.write(remainder)
+            self.pos = len(s2c)
+        self.exhausted = True
+        # The connection stays open: the original server hung up because its
```
sibson: unneeded

tests/functional/test_proxy.py
```diff
@@ -243,21 +245,20 @@ def test_capture_scrubs_vnc_auth_challenge_and_response(self) -> None:
...
-        # count of security types, the one type on offer, then the 16 bytes
-        # that must have been zeroed.
-        self.assertEqual(s2c[14:30], bytes(16), f"auth challenge was not scrubbed: {s2c[:32]!r}")
-
-        # The response sits at a fixed offset on the 3.8 path: 12-byte
-        # version reply, 1-byte security type, then the 16 bytes that must
-        # have been zeroed.
-        self.assertEqual(c2s[13:29], bytes(16), f"auth response was not scrubbed: {c2s[:32]!r}")
-        self.assertIsNone(meta.get("unscrubbable_auth"))
+        # tigervnc-auth negotiated VNC auth, but what was recorded is a
```
sibson: long comment

tests/functional/test_roundtrip.py
```diff
@@ -0,0 +1,112 @@
...
+        self.live = self.tmp / "live"
+        self.replayed = self.tmp / "replayed"
+        self.live.mkdir()
+        self.replayed.mkdir()
+        self.capture = self.tmp / "roundtrip.zip"
+        self.proxy = _start_vnclog(ROUNDTRIP_PROXY_PORT, f"{HOST}::{TIGERVNC.port}", self.capture)
+        self.addCleanup(_stop_proxy, self.proxy)
+
+    def test_replayed_capture_decodes_to_the_same_screenshot(self) -> None:
+        # A relative filename, run from its own directory: the replayed
```
sibson: long comment

tests/functional/test_roundtrip.py
```diff
@@ -0,0 +1,112 @@
...
+            self.assertEqual(
+                live.convert("RGB").tobytes(),
+                replay.convert("RGB").tobytes(),
+                "replayed capture decoded to different pixels than the session it recorded",
+            )
+
+    def _start_replay(self) -> subprocess.Popen:
+        """Serve the capture, waiting until the port answers.
+
+        --forever, so this readiness probe does not consume the one
```
sibson: extra comment

tests/functional/test_replay.py
```diff
@@ -0,0 +1,127 @@
...
+CLIENT_TIMEOUT = 20
+EXIT_TIMEOUT = 10
+
+
+def _pixel(r: int, g: int, b: int) -> bytes:
+    return bytes((r, g, b, 0))
+
+
+def _build_s2c() -> bytes:
+    """greeting + pre-3.8 AuthTypes.NONE handshake + ServerInit + one raw FBU,
```
sibson: what is this even saying?

tests/unit/test_replay.py
```diff
@@ -0,0 +1,248 @@
+"""Unit coverage for vncdotool.replay, the `vncdo-replay --server` half.
```
sibson: redundent

tests/unit/test_replay.py
```diff
@@ -0,0 +1,248 @@
...
+FRAMEBUFFER = b"THE-RECORDED-FRAMEBUFFER"
+UPDATE_REQUEST = pack("!BBHHHH", MsgC2S.FRAMEBUFFER_UPDATE_REQUEST, 0, 0, 0, 640, 480)
+
+# What a stripped capture holds: a none-auth handshake and then the session.
+NONE_38 = VERSION_38 + bytes([1, AuthTypes.NONE]) + pack("!I", 0) + SERVER_INIT_640x480 + FRAMEBUFFER
+NONE_33 = VERSION_33 + pack("!I", AuthTypes.NONE) + SERVER_INIT_640x480 + FRAMEBUFFER
+
+
+def written(protocol) -> bytes:
+    """Everything the protocol has sent, in order."""
```
sibson: gee thanks

tests/unit/test_replay.py
```diff
@@ -0,0 +1,248 @@
...
+    return b"".join(call.args[0] for call in protocol.transport.write.call_args_list)
+
+
+def start(s2c: bytes, **kwargs) -> replay.ReplayProtocol:
+    """Build a protocol on a mocked transport and open the connection."""
```
sibson: interesting

sibson:
🤖 Claude Code: Ran a verified multi-angle review of this branch (8 finder passes, adversarial verification of 18 candidates: 16 confirmed) and pushed fixes as 5bd1584. Highlights:

**Correctness**
- **Replay desynced on version mismatch**: the pacing scrubber branched on the *live* client's version reply while `s2c.bin` was rewritten for the *original* client's. `meta.json` now records `negotiated_version`, and `vncdo-replay --server` refuses a mismatched client with a clear error instead of silently firehosing. Meta-less archives are served best-effort as before.
- **Credential stripping failed open**: a forgotten `_record(b"")` on a future auth-exchange step would have recorded credentials verbatim with all tests green. Inside the auth exchange the default is now record-nothing; a regression test pins the default itself.
- **Pre-3.8 failed-auth captures were unparseable**: the real SecurityResult was recorded into a synthetic none handshake with no slot for it (only the 3.8 path was tested, which is coherent by luck). Fixed, plus `_flush` no longer emits grammar-invalid pending bytes at disconnect.
- **`session.vdo` broke its own parser**: unmapped keysyms recorded as raw `chr(keysym)`, so a captured `'` or `#` crashed or silently altered the shlex parse `vncdo-replay` runs it through — proven by execution. Tokens are now shlex-quoted with a round-trip test.
- Smaller: unbounded buffer growth after capture exhaustion; `--client-timeout 0` armed an immediate warning instead of disabling; a readiness probe consumed the one-shot serve; unfollowable preserved-auth handshakes firehosed silently (now a loud unpaced-serve warning); QEMU extended key events hit the unknown-message give-up; client mode gains `-p` for `--capture-raw-unsafe` archives; `HANDSHAKE_ONLY` in test_proxy was one byte short, making the truncation guard inert.

**Structure/docs**
- `vncdo(argv)` parameter replaces the `sys.argv` mutation re-entry; `capture: Capture` typing; `TemporaryDirectory`.
- One `start_replay_server` harness in `vncservers.py` replaces the two already-divergent copies in test_replay/test_roundtrip.
- Replay rationale deduplicated: full text in `docs/capture.rst`, design doc points there; "What is stripped" tightened ~35%.

**Deferred (deliberate)**: replay still drives the scrubber through its private `_want`/`pos - 24` arithmetic — the deeper fix is a shared public handshake-grammar API, a larger refactor than this pass should carry. Same for full unification of the C2S framing table with `loggingproxy.TYPE_LEN` (QEMU support added; tables still separate). Also flagged in passing: `TYPE_LEN[QEMU_CLIENT_MESSAGE] = 1` in loggingproxy looks one byte short of the subtype it needs to dispatch on.

Verified: 179 unit tests OK, flake8 clean, and the replay + proxy + roundtrip functional suites pass against live tigervnc.

sibson:
🤖 Claude Code: Pushed 782ac09 — the grammar refactor previously deferred, done now per maintainer direction since this PR introduces the code.

- `HandshakeScrubber._run` consumes the ServerInit server name, so the generator ends exactly where the handshake does.
- New public method `waiting() -> ("s2c"|"c2s", nbytes) | None` is the pacing contract; replay consumes only it plus the documented `s2c`/`c2s` taps.
- `server_init_end` and the `pos - 24` back-reckoning are deleted; the unpaceable-handshake warning and truncated-capture path are restated on the public contract (now also covering a capture cut mid-name).
- New tests pin the named-server case the old arithmetic broke silently (every prior fixture used name-len=0): name served before the framebuffer hold, recorded verbatim, partial name dropped at disconnect.

Verified: 180 unit tests OK, flake8 clean, full functional suite 34 tests OK — including the tigervnc roundtrip, whose real server name exercises the new grammar end.

tests/functional/test_roundtrip.py
```diff
@@ -0,0 +1,77 @@
+"""Capture and replay, joined: does what vnclog writes still decode the same?
+
+tigervnc demands a password; the archive of it does not."""
```
sibson: huh?

vncdotool/capture.py
```diff
@@ -56,36 +47,41 @@ class _Want(NamedTuple):
...
 
 
 class HandshakeScrubber:
     """Tracks an RFB handshake across both directions of a proxied stream.
 
-    Feed arriving bytes to :attr:`s2c` or :attr:`c2s`; each returns what that
-    direction should record, credentials replaced by zeros. Once the
-    handshake is done both pass bytes straight through.
-    """
+    Feed :attr:`s2c` / :attr:`c2s`; each returns what that side records."""
```
sibson: huh?

vncdotool/capture.py
```diff
@@ -56,36 +47,41 @@ class _Want(NamedTuple):
...
         self.protocol_version: str | None = None
         self.negotiated_version: tuple[int, int] | None = None
         self.security_types: list[int] = []
         self.security_type: int | None = None
-        self.unscrubbable_auth: str | None = None
-        # Set when an unscrubbable auth type is selected without an opt-in;
-        # the caller drops the capture.
+        self.unstrippable_auth: str | None = None
+        # Set when an auth type we cannot follow is selected without an
+        # opt-in; the caller drops the capture.
```
sibson: redudent?

vncdotool/capture.py
```diff
@@ -56,36 +47,41 @@ class _Want(NamedTuple):
...
         self.abort_reason: str | None = None
         self.width: int | None = None
         self.height: int | None = None
 
         self._gen: Generator[_Want, bytes, None] | None = self._run()
         self._want: _Want | None = None
+        # What to record for the chunk just read, decided by the handshake
```
sibson: what?

vncdotool/capture.py
```diff
@@ -103,18 +99,21 @@ def _advance(self, sent: bytes | None) -> None:
     def _give_up(self, why: str) -> None:
         """Stop the capture when we can no longer follow the handshake.
```
sibson: really, that's what give up means?

vncdotool/capture.py
sibson: 🤖 Claude Code: Fixed — restated Tap's own docstring instead of adding to it; deleted.

vncdotool/capture.py
sibson: 🤖 Claude Code: Fixed — deleted, redundant with the abort_reason name and the message set where it's actually raised.

vncdotool/capture.py
sibson: 🤖 Claude Code: Fixed — reworded to state the override plainly.

vncdotool/capture.py
sibson: 🤖 Claude Code: Fixed — dropped the summary line, kept only the non-obvious part (why preserve_auth is a no-op here).

tests/functional/test_roundtrip.py
sibson: 🤖 Claude Code: Fixed — reworded to a full sentence.

PR #353, opened 2026-08-14
fix(servers): supervise x11vnc's xev sink so a lost startup race self-heals

sibson:
Fixes the intermittent `test_keypress_seen_by_xev` failure that red-flagged #352 (whose changes touch none of this).

## What actually happened

The failure looks like a lost keypress:

```
AssertionError: Regex didn't match: 'xev: KeyPress event' not found in '' :
no KeyPress seen in x11vnc's xev log:
```

The empty string is the tell — not "the KeyPress is missing from the log" but "there is no log". The container-log dump from that same CI run has it, as the very first line the container ever wrote:

```
x11vnc-1  | xev: xev:  unable to open display ':0'
```

The X-side event sink lost the startup race against Xvfb, died, and nothing restarted it. Nothing else noticed: x11vnc kept serving, the container stayed healthy through `--wait`, and all four x11vnc smoke tests passed in that run. Every event test afterwards polled a log the dead sink would never write to again.

A successful `xdpyinfo` probe does not make this safe — Xvfb accepts connections before it has finished initialising, the same "accepting connections is not the same as ready" problem `draw-content.sh` already documents for Xvnc. Lengthening the probe only changes the odds.

## The fix

Three parts, all in the container rather than in the tests:

1. **Supervise `xev`** instead of starting it once: restart it if it exits, bounded at 60 attempts. A lost race becomes a half-second gap rather than a poisoned run. `|| true` around `xev` is load-bearing under `set -e`, which would otherwise take the supervising subshell down with the process it exists to restart.
2. **Exit if the sink cannot be kept alive.** After the 60th attempt the supervisor does `kill -TERM 1`. This container promises to serve VNC *and* to observe X events; one without the other is a container that passes every smoke test while answering no event question correctly. Failing as a dead container beats failing as a keypress assertion that points at vncdotool.
3. **Make the sink part of `healthy`.** The HEALTHCHECK adds `pgrep -f "xev -root"`, so `up --wait` gates on the sink being up, not just the port.

The entrypoint also clears `/tmp/.X0-lock`, the X socket, and the readiness file on start. Restarting a stopped container reuses its filesystem, so without that, the first exit is unrecoverable — `make servers-up` comes back to `Fatal server error: Server is already active for display 0` forever.

## Verification

Against the local fleet, with the fixed image:

- baseline `vncdo key x` → 1 `KeyPress` in the log
- `pkill -x xev` once → log shows `xev exited; restarting the X-side event sink` and the next keypress lands
- `pkill -x xev` every 0.2s until the supervisor gives up → `xev: xev kept exiting; taking the container down with it`, container `Exited (2)`
- `docker compose up -d --wait x11vnc` on that stopped container → `Healthy`, `test_events` green again

Unfixed, a single lost race loses every event for the container's lifetime — which is exactly the CI failure. `python -m unittest tests.functional.test_events` green (4 tests); unit suite and flake8 clean.

## Discarded hypothesis, recorded so it isn't re-tried

First read was stdio block buffering: `xev | sed -u`, where `-u` unbuffers sed but not xev upstream of it. Measured instead of assumed — the log grows 627 bytes per keypress with no 4KB granularity, so xev's output is not block-buffered here and `stdbuf` fixes nothing.

### DIFF
--- a/tests/functional/test_events.py
+++ b/tests/functional/test_events.py
@@ -128,6 +128,24 @@ class TestVncevSink(TestCase):
         )
 
 
+def _sink_alive(service: str, pattern: str) -> bool:
+    """Is the container's event sink process still running?
+
+    A sink that died leaves the container serving VNC quite happily, so
+    nothing else in the suite notices; asking the container directly is
+    the only way to tell "the event never arrived" from "nothing was
+    listening for it".
+    """
+    result = subprocess.run(
+        ["docker", "compose", "-f", str(COMPOSE_FILE), "exec", "-T", service, "pgrep", "-f", pattern],
+        capture_output=True,
+        text=True,
+        timeout=30,
+        stdin=subprocess.DEVNULL,
+    )
+    return result.returncode == 0
+
+
 class TestX11SideSink(TestCase):
     """Server processing: did x11vnc's X display see a real X event.
 
@@ -143,6 +161,14 @@ class TestX11SideSink(TestCase):
                 f"x11vnc not reachable on {HOST}:{X11VNC.port} -- start the servers first "
                 "with `make servers-up`"
             )
+        # Fail as the infrastructure problem it is. Without this the symptom
+        # of a dead sink is an assertion about a keypress, pointing at
+        # vncdotool for something vncdotool did not do.
+        self.assertTrue(
+            _sink_alive("x11vnc", "xev -root"),
+            "x11vnc's xev sink is not running, so no event can be observed no matter what "
+            "the client sends. Check `docker compose logs x11vnc` for why it exited.",
+        )
 
     def test_keypress_seen_by_xev(self) -> None:
         offset = len(_compose_logs("x11vnc"))
--- a/tests/functional/test_events.py
+++ b/tests/functional/test_events.py
@@ -128,24 +128,6 @@ class TestVncevSink(TestCase):
         )
 
 
-def _sink_alive(service: str, pattern: str) -> bool:
-    """Is the container's event sink process still running?
-
-    A sink that died leaves the container serving VNC quite happily, so
-    nothing else in the suite notices; asking the container directly is
-    the only way to tell "the event never arrived" from "nothing was
-    listening for it".
-    """
-    result = subprocess.run(
-        ["docker", "compose", "-f", str(COMPOSE_FILE), "exec", "-T", service, "pgrep", "-f", pattern],
-        capture_output=True,
-        text=True,
-        timeout=30,
-        stdin=subprocess.DEVNULL,
-    )
-    return result.returncode == 0
-
-
 class TestX11SideSink(TestCase):
     """Server processing: did x11vnc's X display see a real X event.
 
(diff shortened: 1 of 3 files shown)


sibson:
Second commit added: supervising the sink stops it staying dead, but nothing noticed it had died. The x11vnc HEALTHCHECK only proved the VNC port answered, so a container with no event sink reported healthy, satisfied `--wait`, and 27 tests passed before the one test that reads the sink failed with an assertion about a missing keypress.

Now the healthcheck requires `xev -root` to be running (a fleet started with a dead sink fails at `docker compose up --wait`, where the cause is still on screen), and `TestX11SideSink.setUp` asserts sink liveness so a mid-run death names the sink rather than the keypress. Verified the healthcheck predicate passes with the sink alive and fails immediately after `pkill -f "xev -root"`.

vncev needs neither: it is its container's main process, so its death takes the container with it.

tests/functional/test_events.py
```diff
@@ -128,6 +128,24 @@ def test_mousemove_seen_by_vncev(self) -> None:
         )
 
 
+def _sink_alive(service: str, pattern: str) -> bool:
+    """Is the container's event sink process still running?
+
+    A sink that died leaves the container serving VNC quite happily, so
```
sibson: why doesn't the container just crash/exit if the sink is dead?

tests/servers/Dockerfile
```diff
@@ -60,11 +60,21 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
         x11vnc \
         x11-utils \
+        procps \
     && rm -rf /var/lib/apt/lists/*
 
 COPY x11vnc-entrypoint.sh /entrypoint.sh
 RUN chmod +x /entrypoint.sh
 
+# Extends base's check with the X-side event sink, because this container
+# promises two things and the port only evidences one. A dead xev used to
```
sibson: review comment policy

tests/functional/test_events.py
sibson: Made the container own it. After 60 failed xev restarts the supervisor does `kill -TERM 1`, so the container exits (verified: `Exited (2)`, preceded by `xev: xev kept exiting; taking the container down with it`). The test-side `_sink_alive()` probe and its `setUp` assert are gone — wrong layer.

Verifying that turned up a follow-on: a container that exits leaves `/tmp/.X0-lock` behind, and restarting a stopped container reuses its filesystem, so `make servers-up` came back to `Fatal server error: Server is already active for display 0` forever. The entrypoint now clears the lock, the socket, and the readiness file on start. Kill loop -> exit -> `up -d --wait` -> healthy, tests green.

tests/servers/Dockerfile
sibson: Trimmed to why-only. The HEALTHCHECK comment is 3 lines (the port evidences half of what this container promises; the sink loses a startup race often enough that readiness has to wait for it) and the entrypoint sink comment dropped the bug narration, keeping that xdpyinfo succeeding does not mean the next X client can connect.

PR #356, opened 2026-08-15
Stop the readiness probe blacklisting itself, and gate the fleet on wait_for_servers

sibson:
This branch was written before #351/#353 and never got a PR, so it sat stranded while the stack moved on — and the problem it fixes kept resurfacing. Rebased onto current main and verified against a real fleet.

## The problem

The base HEALTHCHECK opened an RFB connection and hung up without speaking the protocol. Xvnc counts every connection closed before a successful authentication towards `BlacklistThreshold` (default 5), so the probe blacklisted 127.0.0.1 about ten seconds after start-up and refused every session after that — captures against tigervnc-auth silently produced no PNG.

The workaround at the time was `-BlacklistThreshold=1000000`, which turns the brute-force protection off rather than fixing the probe.

## The fix

Measured against TigerVNC 1.12.0, **no RFB-level probe avoids the counter**: a bare connect, a version exchange, and a version-plus-security-types exchange each trip it after five attempts, and only a successful authentication clears the mark. So the probe reads the kernel socket table with `ss` instead, confirming the port is listening without a connection ever existing.

The threshold is then *reduced* to a targeted 50 rather than removed, because the probe was not the only offender: `vncservers.py`'s `port_open()` guard leaves about seven unauthenticated closes per run, and `test_cli.py` adds one deliberate wrong password.

Readiness itself moves out of the container: `wait_for_servers.py` gates both server families by connecting, capturing, and requiring the content that server is supposed to render — the same assertion the tests make, so the gate and the tests cannot drift. The old marker file was written unconditionally, so its failure mode was a false green: a server that came up blank still reported healthy.

## Verification against a real fleet

- whole fleet rebuilt and started: all five containers reach `Healthy` on the connectionless probe
- `wait_for_servers.py docker`: every server ready on the first attempt
- full functional suite: **31 tests, OK** (4 skipped — OS-hosted servers)
- after that run, tigervnc's log shows **20 connection entries and 0 blacklist entries**, while roughly 35 healthcheck probes had elapsed. Connections track test sessions, not probes.

## One integration fix

x11vnc grew its own HEALTHCHECK while this branch sat unmerged, keeping both habits removed from base: an RFB connection via `/dev/tcp`, and the marker-file gate. The connection is harmless there — x11vnc has no blacklist — but it is the pattern that broke TigerVNC, and no reason to leave an exception lying around to be copied. Its check now uses `ss` plus the thing genuinely specific to that container, its event sink.

Supersedes #355, which raised the threshold on both entrypoint branches — spreading the workaround instead of deleting it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,6 +98,14 @@ jobs:
       - name: Build and start the VNC test servers
         run: docker compose -f tests/servers/docker-compose.yml up -d --build --wait
 
+      # `up --wait` only reaches the containers' HEALTHCHECK, which is
+      # liveness. This is the readiness gate: it captures until every server
+      # serves the content it should, so a server that comes up blank fails
+      # here, named, rather than as a flat capture inside a test.
+      - name: Wait for the test servers to serve their screens
+        timeout-minutes: 5
+        run: python tests/functional/wait_for_servers.py docker
+
       # The steps below are diagnostics and teardown, so they carry
       # `if: always()`: when a test fails, the container status, the
       # screenshot of what the server was actually showing, and the server
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,10 +98,9 @@ jobs:
       - name: Build and start the VNC test servers
         run: docker compose -f tests/servers/docker-compose.yml up -d --build --wait
 
-      # `up --wait` only reaches the containers' HEALTHCHECK, which is
-      # liveness. This is the readiness gate: it captures until every server
-      # serves the content it should, so a server that comes up blank fails
-      # here, named, rather than as a flat capture inside a test.
+      # `up --wait` only reaches the HEALTHCHECK, which is liveness. A
+      # server that comes up blank fails here, named, rather than as a flat
+      # capture inside a test.
       - name: Wait for the test servers to serve their screens
         timeout-minutes: 5
         run: python tests/functional/wait_for_servers.py docker
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,9 +98,6 @@ jobs:
       - name: Build and start the VNC test servers
         run: docker compose -f tests/servers/docker-compose.yml up -d --build --wait
 
-      # `up --wait` only reaches the HEALTHCHECK, which is liveness. A
-      # server that comes up blank fails here, named, rather than as a flat
-      # capture inside a test.
       - name: Wait for the test servers to serve their screens
         timeout-minutes: 5
         run: python tests/functional/wait_for_servers.py docker
--- a/tests/functional/vncservers.py
+++ b/tests/functional/vncservers.py
@@ -155,12 +155,7 @@ def screenshot_dir() -> Path:
 
 
 def port_open(host: str, port: int, timeout: float = PORT_PROBE_TIMEOUT) -> bool:
-    """Whether ``port`` accepts a connection. Cheap liveness, not readiness.
-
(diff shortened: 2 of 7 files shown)


tests/functional/vncservers.py
```diff
@@ -154,6 +155,12 @@ def screenshot_dir() -> Path:
 
 
 def port_open(host: str, port: int, timeout: float = PORT_PROBE_TIMEOUT) -> bool:
+    """Whether ``port`` accepts a connection. Cheap liveness, not readiness.
+
+    This hangs up without authenticating, which TigerVNC counts towards
```
sibson: over comment

.github/workflows/ci.yml
```diff
@@ -98,6 +98,14 @@ jobs:
       - name: Build and start the VNC test servers
         run: docker compose -f tests/servers/docker-compose.yml up -d --build --wait
```
sibson, line 100: over comment

tests/functional/vncservers.py
```diff
@@ -183,23 +190,44 @@ def capture_screenshot(server: VNCServer, path: Path, timeout: Optional[float] =
     return path
 
 
+def distinct_colours(image: Image.Image) -> Optional[int]:
+    """How many colours ``image`` contains, or None above ``MAX_COLOURS``."""
```
sibson, line 189: over comment

sibson: Deleted rather than shortened. `port_open`'s blacklist aside, `has_expected_content`'s flat-capture paragraph, `wait_until_ready`'s essay, and the CI step's comment (which restated its own name) are all gone -- 38 lines out, 8 in.

Two of your three flags were docstring paragraphs, which I had never audited because I scoped the rule to `#` comments. That was the gap.

tests/functional/vncservers.py
sibson: 🤖 Claude Code: Deleted in c604a9b — the blacklist aside now lives only in tigervnc-entrypoint.sh's comment, one home instead of two.

.github/workflows/ci.yml
sibson: 🤖 Claude Code: Deleted in c604a9b — restated that `up --wait` is liveness-only, which the Dockerfile's HEALTHCHECK comment already says once.

tests/servers/tigervnc-entrypoint.sh
```diff
@@ -15,12 +15,16 @@ if [ -n "$VNC_PASSWORD" ]; then
...
-    set -- -SecurityTypes VncAuth -PasswordFile /root/.vnc/passwd -BlacklistThreshold=1000000
+    # Xvnc counts every connection closed before a successful authentication
+    # towards BlacklistThreshold, and a successful authentication clears the
+    # host's blackmark. The harness leaves a few such closes even though the
+    # readiness probe never connects (see the HEALTHCHECK in Dockerfile):
+    # about seven port_open() reachability probes per full run, plus the one
+    # deliberate wrong password in tests/functional/test_cli.py. A full run
+    # interleaves enough successful authentications to stay under the default
+    # of 5, but running test_cli.py alone, repeatedly, chains them without one.
+    # 50 absorbs several such runs while still capping real password guessing.
```
sibson: we got verbos here...

sibson: 🤖 Claude Code: Trimmed in 54ae7cf — the arithmetic (seven probes, one wrong password) went to the PR discussion, comment keeps just the mechanism and why 50.

PR #357, opened 2026-08-15
Make `make test-func` use the venv it builds, and say so when it can't

sibson:
Working in a second checkout kept producing tests that passed against the wrong code — most recently in #352, where the functional suite had to be driven through hand-written shims to exercise the branch at all. Three causes, each of which hides the next.

## `make test-func` never used the venv it builds

It ran `$(PYTHON)`, which reads like a project variable but is **one of GNU make's built-ins**, defaulting to `python3`:

```
$ make -n -p | grep '^PYTHON ='
PYTHON = python3
```

So the target used the system interpreter while `make test-unit` beside it used `$(VENV)/python`. CI never noticed because it installs into the same system interpreter it invokes.

Now it runs `$(VENV)/python` with `$(VENV)` prepended to `PATH`, which also settles *which* `vncdo` the suite shells out to — previously whichever one `PATH` happened to offer.

## A too-old interpreter failed as a dependency-resolution error

`Makefile.venv` builds `.venv` with the first `python3` it finds. A fresh checkout has no interpreter pin (`.tool-versions` is gitignored), so a version manager has nothing to activate and the system python is used. When that is too old, the failure looks like this:

```
ERROR: Could not find a version that satisfies the requirement Pillow>=12.2.0
```

which reads like a broken pin rather than a wrong interpreter — and the tempting fix is to borrow another checkout's `.venv`, which is cause three.

`check-python` now states the floor, names `PY=`, and warns off that fix.

## The declared floor was wrong

`setup.cfg` advertised `>=3.9`. CI tests 3.10–3.13, and the development requirements cannot install on 3.9 at all. Now `>=3.10`, so the three agree.

## The suite refuses to run against a foreign install

Borrowing another checkout's `.venv` is silent: the editable install records the path it was created from, so its `vncdotool` and its `vncdo` still point there. The suite passes, reporting that copy's behaviour as this branch's.

It now asks the `vncdo` on `PATH` which vncdotool *its own* interpreter imports — what the test process imports proves nothing, since running from the working tree puts it on `sys.path` — and fails with both paths named:

```
AssertionError: `vncdo` on PATH (/Users/marcs/src/vncdotool/.venv/bin/vncdo) imports
vncdotool from /Users/marcs/src/vncdotool/vncdotool/__init__.py, not this checkout at
.../scratchpad/wt-env. These tests shell out to it, so they would report that copy's
behaviour as this branch's. Run `make venv` in this working tree...
```

## Verification

In a fresh working tree, against the docker fleet:

- `make test-func` alone — no `PYTHONPATH`, no shims, no symlink — **31 tests OK** (4 skipped, OS-hosted servers)
- `make test-func PY=python3` (3.9 here) — explains the floor and stops
- running the suite against another checkout's venv — names both paths and refuses
- `make test` — 125 tests OK; flake8 clean


docs/contributing.rst
```diff
@@ -3,6 +3,34 @@ Contributing
 
 Code and Issue tracking is provided by Github_.  There is also a mailing list setup via `Google Groups`_.
 
+Development environment
```
sibson: these belong in develop.rst rather htan contributing

sibson: Moved to a new `docs/develop.rst` and added it to the toctree; `contributing.rst` is back to exactly what it was on main.

Agreed on the split: contributing is how to bring a change to the project, this is what you do once you are working on one. Docs build clean and the page renders and links.

sibson: Right, and I should have looked before creating a page: `DEVELOP.rst` at top level already owned this. The notes are folded in there, `docs/develop.rst` is deleted, and `docs/index.rst` and `contributing.rst` are back to what main has.

Its Running Tests section also needed correcting while I was in there — it predated the Makefile building its own venv, so it told contributors to run `virtualenv venv` by hand and to `pip install -e .` so `vncdo` was on PATH. With this PR neither is true: `make` builds `.venv`, and `make test-func` puts it on PATH itself.

PR #360, opened 2026-08-16
Add the in-process API lifecycle suite (Phase 5)

sibson:
## Summary
- `tests/functional/test_api_lifecycle.py` covers `vncdotool.api`'s lifecycle in-process (connect/op/disconnect, context-manager use, sequential reconnects, TimeoutError, prompt ConnectError) against the tigervnc container -- the one module permitted to call `api.connect()`, since a process gets exactly one Twisted reactor. `make test-api` runs it alone; it also sorts first in the shared functional discover.
- `make test-api` target added; design doc and `vncservers.connect()` docstring updated to state the one-module rule and why it's safe.

## Test plan
- [x] `make test-api` -- 5/5 green
- [x] full functional discover (39 tests) -- green, 4 skipped
- [x] unit suite + flake8 -- clean

Last phase of the server testing framework (#347, #348, #350, #351, #352 already merged).

### DIFF
--- a/tests/functional/test_api_lifecycle.py
+++ b/tests/functional/test_api_lifecycle.py
@@ -2,14 +2,12 @@
 use, reconnection, error propagation, timeouts, ``shutdown()`` cleanliness.
 Server compatibility belongs to the subprocess grid in test_servers.py.
 
-THIS IS THE ONE MODULE ALLOWED TO CALL api.connect().
------------------------------------------------------
-
 ``api.connect()`` starts the Twisted reactor in a daemon thread, and a
 reactor cannot be restarted once stopped, so one process gets one reactor
-for its whole lifetime. A second in-process module would either race this
-one for it or leave its own reactor thread wedged on exit -- new in-process
-tests belong in *this* file, sharing *this* reactor.
+for its whole lifetime: this is the only module allowed to call it. A
+second in-process module would either race this one for it or leave its
+own reactor thread wedged on exit -- new in-process tests belong in
+*this* file, sharing *this* reactor.
 
 ``setUpModule`` makes the first connection itself, so the reactor is
 running before any test method does -- no test depends on being first.
@@ -77,9 +75,7 @@ class TestApiLifecycle(unittest.TestCase):
             client.keyPress("x")
 
     def test_sequential_reconnects(self) -> None:
-        """The reactor survives a disconnect and serves a second connection.
-
-        ``shutdown()`` is what's terminal, not ``disconnect()``: otherwise no
+        """``shutdown()`` is what's terminal, not ``disconnect()``: otherwise no
         long-running application could reconnect after a single drop.
         """
         for _ in range(2):
@@ -87,10 +83,7 @@ class TestApiLifecycle(unittest.TestCase):
                 client.refreshScreen()
 
     def test_timeout_raises_timeout_error(self) -> None:
-        """A call against a port that never speaks RFB raises TimeoutError
-        instead of hanging.
-
-        ``api.connect()`` is fire-and-forget and never raises here; the
+        """``api.connect()`` is fire-and-forget and never raises here; the
         timeout comes from the first call that has to wait on the
         never-completing handshake.
         """
@@ -113,9 +106,7 @@ class TestApiLifecycle(unittest.TestCase):
         self.assertLess(elapsed, SHORT_TIMEOUT + 1.0)
 
     def test_closed_port_raises_promptly(self) -> None:
-        """connect() to a port nothing listens on fails fast, not hangs.
-
-        Bounded twice over -- a per-client timeout, and a helper thread
+        """Bounded twice over -- a per-client timeout, and a helper thread
         joined with a deadline -- so even a wedge fails only this test,
         not the suite.
 
--- a/tests/functional/test_api_lifecycle.py
+++ b/tests/functional/test_api_lifecycle.py
@@ -33,7 +33,6 @@ TIGERVNC = next(s for s in DOCKER_SERVERS if s.name == "tigervnc")
 # observed to flake under load.
 HAPPY_TIMEOUT = TIGERVNC.timeout + SUBPROCESS_TIMEOUT_HEADROOM
 
-# Tight budget for the cases that want to observe a timeout quickly.
 SHORT_TIMEOUT = 2.0
 
 # join() budget for the closed-port helper thread: long enough to fail
@@ -43,7 +42,7 @@ HANG_GUARD_TIMEOUT = 10.0
 
 def setUpModule() -> None:
     if not port_open(HOST, TIGERVNC.port):
-        raise unittest.SkipTest(
+        raise RuntimeError(
             f"tigervnc not reachable on {HOST}:{TIGERVNC.port} -- "
             "start the servers first with `make servers-up`"
         )
@@ -188,8 +187,6 @@ class _SilentListener:
                 continue  # idle poll; socket.timeout IS an OSError, catch it first
             except OSError:
                 return
-            # Accept and go silent: read nothing, write nothing, until
-            # told to stop.
             while not self._stopped.is_set():
                 time.sleep(0.1)
             conn.close()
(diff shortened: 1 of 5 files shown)


tests/functional/test_api_lifecycle.py
```diff
@@ -0,0 +1,206 @@
+"""In-process coverage of ``vncdotool.api``'s *lifecycle* -- context-manager
+use, reconnection, error propagation, timeouts, ``shutdown()`` cleanliness.
+Server compatibility belongs to the subprocess grid in test_servers.py.
+
+THIS IS THE ONE MODULE ALLOWED TO CALL api.connect().
```
sibson: gratuitiious, Extraneous comments

tests/functional/test_api_lifecycle.py
```diff
@@ -0,0 +1,206 @@
...
+
+SHORT_TIMEOUT = 2.0
+
+# join() budget for the closed-port helper thread: long enough to fail
+# promptly, short enough that a real hang doesn't stall the suite.
+HANG_GUARD_TIMEOUT = 10.0
+
+
+def setUpModule() -> None:
+    if not port_open(HOST, TIGERVNC.port):
```
sibson: better to run against libvncserver rather than tiger?

tests/functional/test_api_lifecycle.py
```diff
@@ -0,0 +1,206 @@
...
+def tearDownModule() -> None:
+    # Exactly once per process: without it the interpreter hangs on the
+    # non-daemon worker threads Twisted runs under the reactor.
+    api.shutdown()
+
+
+class TestApiLifecycle(unittest.TestCase):
+    """Ordered lifecycle cases against the tigervnc:5931 container.
+
+    Names are prefixed to fix run order: well-behaved cases first, the
```
sibson: Dependent ordering is a code smell.  We should be reslient to failures.  If not let's dig into why to establish this is required with evidence rather than assumed.

tests/functional/test_api_lifecycle.py
```diff
@@ -0,0 +1,206 @@
...
+    potentially-hanging one last.
+    """
+
+    def test_a_connect_op_disconnect(self) -> None:
+        """connect -> one trivial op -> disconnect: clean, no exception."""
```
sibson: pointless?

tests/functional/test_api_lifecycle.py
```diff
@@ -0,0 +1,206 @@
...
+        client = api.connect(f"{HOST}::{TIGERVNC.port}")
+        client.timeout = HAPPY_TIMEOUT
+        try:
+            client.refreshScreen()
```
sibson: rather than op let's call it refresh.  op feels more like a keypress

tests/functional/test_api_lifecycle.py
```diff
@@ -0,0 +1,206 @@
...
+class TestApiLifecycle(unittest.TestCase):
+    """Ordered lifecycle cases against the tigervnc:5931 container.
+
+    Names are prefixed to fix run order: well-behaved cases first, the
+    potentially-hanging one last.
+    """
+
+    def test_a_connect_op_disconnect(self) -> None:
+        """connect -> one trivial op -> disconnect: clean, no exception."""
+        client = api.connect(f"{HOST}::{TIGERVNC.port}")
```
sibson: should connect/disco be part of setup/teardown?  Avoids duplicating code?

tests/functional/test_api_lifecycle.py
```diff
@@ -0,0 +1,206 @@
...
+        """The reactor survives a disconnect and serves a second connection.
+
+        ``shutdown()`` is what's terminal, not ``disconnect()``: otherwise no
+        long-running application could reconnect after a single drop.
+        """
+        for _ in range(2):
+            with connect(TIGERVNC, timeout=HAPPY_TIMEOUT) as client:
+                client.refreshScreen()
+
+    def test_d_timeout_raises_timeout_error(self) -> None:
```
sibson: what is the d for here?

tests/functional/test_api_lifecycle.py
```diff
@@ -0,0 +1,206 @@
...
+        client.timeout = SHORT_TIMEOUT
+        self.addCleanup(client.disconnect)
+
+        start = time.monotonic()
+        with self.assertRaises(TimeoutError):
+            client.refreshScreen()
+        elapsed = time.monotonic() - start
+
+        # Generous bound: proves the timeout fired, without being a
+        # flakiness trap on a loaded CI box.
```
sibson: pointless comment?

tests/functional/test_api_lifecycle.py
```diff
@@ -0,0 +1,206 @@
...
+        # failure satisfies the lifecycle guarantee.
+        self.assertIsInstance(outcome["exception"], ConnectError)
+
+
+class _SilentListener:
+    """A TCP listener that accepts and then never speaks, forcing a
+    protocol-level timeout rather than a connection-refused.
+    """
+
+    def __init__(self) -> None:
```
sibson: Is there not a more Twisted native version of this?

tests/functional/test_api_lifecycle.py
sibson: 🤖 Claude Code: Trimmed. Dropped the now-stale "ordering matters for the same reason" paragraph (the ordering claim itself is gone, see below) and kept only the one-reactor-per-process rationale, which is still load-bearing.

tests/functional/test_api_lifecycle.py
sibson: 🤖 Claude Code: Switched to libvncserver-example -- lighter, more deterministic, and no behavioural reason to prefer tigervnc for a lifecycle-only suite.

tests/functional/test_api_lifecycle.py
sibson: 🤖 Claude Code: Checked rather than assumed: the closed-port case is already bounded by its own timeout + a joined helper thread, so a wedge there fails only that test. Verified live by running it first (it now sorts first alphabetically) and confirming the tests after it still passed. Dropped the a/b/c/d/z ordering scheme entirely -- it was redundant with protection already in place.

tests/functional/test_api_lifecycle.py
sibson: 🤖 Claude Code: Removed as a separate test. setUpModule now makes this exact connect/refresh/disconnect itself, which also fixes the reactor-startup ordering dependency (see the ordering thread below) rather than relying on this test happening to run first.

tests/functional/test_api_lifecycle.py
sibson: 🤖 Claude Code: That exact docstring text is gone now (test_a folded into setUpModule, see above), but noting for the record: agreed, 'refresh' is the clearer word for what refreshScreen() does.

tests/functional/test_api_lifecycle.py
sibson: 🤖 Claude Code: Per-test setUp/tearDown doesn't actually fit here -- test_context_manager_connect, test_sequential_reconnects, and the timeout/closed-port cases each need a different connection pattern or target on purpose, so a shared setUp would just add an unused connection to most of them. Moved the one case that did want a plain connect/refresh/disconnect into setUpModule instead, which removes the duplication AND fixes the reactor-startup-ordering issue in the other thread.

tests/functional/test_api_lifecycle.py
sibson: 🤖 Claude Code: Letter prefixes dropped entirely -- tests are now named for what they check (test_context_manager_connect, test_sequential_reconnects, test_timeout_raises_timeout_error, test_closed_port_raises_promptly), no ordering scheme.

tests/functional/test_api_lifecycle.py
sibson: 🤖 Claude Code: Agreed on reflection -- the +5.0 was unexamined padding, and nothing established the tighter bound as actually flaky. Tightened to SHORT_TIMEOUT + 1.0 and rewrote the comment to state that reasoning instead of hand-waving 'generous.'

tests/functional/test_api_lifecycle.py
sibson: 🤖 Claude Code: Yes -- replaced the hand-rolled socket+threading listener with a Protocol that does nothing, registered on api.connect()'s own already-running reactor via reactor.callFromThread(reactor.listenTCP, ...) instead of a second, unrelated listener thread.

PR #366, opened 2026-08-17
Measure and report coverage for both test tiers, gating nothing

sibson:
Coverage was never measured anywhere in this repo. This adds it as a
number that is reported and never gated: no threshold, no build failing
on it.

## What CI does now

A new `coverage` job combines data from the unit matrix and the fleet
job and writes three tables to the run's job summary — unit tier, fleet
tier, and combined — plus an `htmlcov/` artifact.

Both tiers are reported separately because they mean different things. A
fall in the unit number is a regression in the diff; a fall in the fleet
number usually means a container did not come up. One blended number
cannot tell them apart. Measured locally on this branch:

| tier | coverage |
| --- | --- |
| Unit | 62% |
| Fleet | 65% |
| Combined | 75% |

The job carries `if: always()` and does not depend on the suites
passing. The fleet job is the flakiest thing here and its coverage is
most interesting exactly when it went red.

## Subprocess measurement

Almost every line the functional suite covers executes in a `vncdo`,
`vnclog` or `vncdo-replay` subprocess, and a subprocess only measures
itself if coverage starts before it imports anything. A `.pth` in
site-packages plus `COVERAGE_PROCESS_START` arranges that, in CI and in
`make coverage-func`. Written inline rather than pulling in
`coverage_enable_subprocess` so `requirements-dev.txt` stays a list of
things the tests import; `make` installs `coverage` on demand through
Makefile.venv's wildcard rule.

A child that is killed never runs `atexit` and so never writes its data.
`start_replay_server` now terminates its process instead: Twisted's own
SIGTERM handler stops the reactor, so the process exits cleanly with its
coverage data — and any file it still owes — flushed. **No signal
handling of ours is involved**; this was verified by probe before the
change, not assumed. The remaining `kill()` calls are failure paths
where the process is already wedged.

## Also fixed

Verifying this turned up a pre-existing bug: `make test-func` put the
*relative* `./.venv/bin` on `PATH`. `test_roundtrip` runs the CLI with a
`cwd=` of its own, so that entry resolved against the temp directory and
the test picked up whatever else was on `PATH` — one test failing
locally on main for reasons unrelated to the code under test. Now
`$(abspath $(VENV))`.

## Verification

- `make coverage` — 180 unit tests OK, 38 functional OK, combined 75%,
  against the live Docker fleet.
- `make test-func` — 38/38, previously 37/38 for the `PATH` reason above.
- `flake8 --count --statistics vncdotool tests` — clean.
- The report script was dry-run against real tier data before landing.
  It caught that `coverage combine --data-file=X` discovers inputs named
  `X.*`, so the obvious spelling silently reports nothing; the script
  combines normally and renames instead.

No `CHANGELOG.rst` entry: nothing here is user-visible.


.github/workflows/ci.yml
```diff
@@ -132,3 +166,73 @@ jobs:
...
         if: always()
         run: docker compose -f tests/servers/docker-compose.yml down
+
+  coverage:
+    name: Coverage
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    needs: [test, servers]
+    # Coverage is reported, never gated: no threshold, and no dependency on
+    # the suites passing. The fleet job is the flakiest one here, and its
```
sibson: we keep saying the fleet job is flakey based on what proof?

.github/workflows/ci.yml
```diff
@@ -132,3 +166,73 @@ jobs:
...
+          name: coverage-fleet
+          path: data/fleet
+
+      # Per tier as well as combined: a drop in the unit tier is a
+      # regression in the diff, a drop in the fleet tier usually means a
+      # container did not come up. One blended number cannot tell them
+      # apart. `--keep` leaves the inputs for the passes that follow.
+      - name: Report
+        run: |
+          mkdir -p combined
```
sibson: can this be a script rather than long exec blob inside yaml?

.github/workflows/ci.yml
sibson: 🤖 Claude Code: No proof — you are right to push on it. I checked: of the last 39 completed runs of the "Test server functional tests" job, 36 passed and 3 failed. All 3 failures were on `claude/ci-build-cache` and all 3 were the same broken workflow edit (`ERROR: open docker-compose.yml: no such file or directory`), not the fleet.

So the claim is unsupported and I have removed it from the comment and from DEVELOP.rst. `if: always()` still earns its place on a smaller claim that needs no evidence: a suite that fails should still report the coverage it reached before it stopped.

.github/workflows/ci.yml
sibson: 🤖 Claude Code: Done — it is `.github/scripts/coverage-summary.sh` now, and the workflow step is one line:

```yaml
run: .github/scripts/coverage-summary.sh unit=data/unit fleet=data/fleet
```

It takes `NAME=DIR` pairs and appends to `$GITHUB_STEP_SUMMARY`, falling back to stdout when that is unset, so the CI report can be reproduced locally against any pair of data directories. Verified both tiers populated (62/65/75%) and with an empty fleet directory (reports "No data", still emits a total, exits 0). shellcheck clean.

.github/workflows/ci.yml
```diff
@@ -44,10 +44,21 @@ jobs:
...
-        run: python -m unittest discover tests/unit
+        run: coverage run --parallel-mode -m unittest discover tests/unit
+
+      - name: Upload coverage data
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-unit-${{ matrix.python-version }}
+          path: .coverage.*
+          # The data files are dotfiles, which upload-artifact skips by
```
sibson: this is somewhat obvious from the previous and next lines

.github/workflows/ci.yml
```diff
@@ -92,10 +103,20 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -r requirements-dev.txt
+          pip install -r requirements-dev.txt 'coverage>=7'
```
sibson: coverage should be added to requirements-dev not added here

.github/workflows/ci.yml
```diff
@@ -92,10 +103,20 @@ jobs:
...
           # The functional suite shells out to `vncdo`, so it has to be on PATH.
           pip install -e .
 
+      # A subprocess measures itself only if coverage starts before it
+      # imports anything; see DEVELOP.rst.
+      - name: Enable coverage in subprocesses
```
sibson: should this be a command line option we add to scripts or something else that isn't quite a hacky?

sibson: 🤖 Claude Code: You were right, it was hacky and it is now unnecessary. coverage 7.10 added a `patch` setting that does exactly this:

```ini
[coverage:run]
patch = subprocess
```

It starts coverage in processes created by `subprocess`, `os.system()` and the `execv`/`spawnv` families. That was the entire job of the hand-written `.pth` plus `COVERAGE_PROCESS_START`, so both are gone — along with the CI step and the make stamp target that installed the `.pth`. One config line replaces about 20.

Verified locally: `make coverage` still combines 34 data files and reports the same 75% total, so the CLI subprocesses are measured exactly as before. `coverage>=7.10` is pinned in requirements-dev.txt for it.

.github/workflows/ci.yml
sibson: 🤖 Claude Code: Moved. requirements-dev.txt now carries `coverage>=7.10` and the three inline installs are gone; the coverage job installs requirements-dev.txt like the others, with the same pip cache.

For the record my reason for keeping it out was thin — I wanted requirements-dev.txt to stay a list of things the tests import — and it cost a version pin duplicated across three steps.

.github/workflows/ci.yml
sibson: 🤖 Claude Code: Agreed, deleted. `path: .coverage.*` and `include-hidden-files: true` on adjacent lines do say it.

PR #367, opened 2026-08-17
fix(servers): survive docker compose restart on TigerVNC

sibson:
## Summary
- `docker compose restart` reuses the container filesystem. If Xvnc didn't exit cleanly, a stale `/tmp/.X0-lock` survives and the next start dies with "Server is already active for display 0".
- x11vnc's entrypoint already clears its lock for the same reason (`tests/servers/x11vnc-entrypoint.sh`); TigerVNC's did not.
- Fix mirrors that pattern: clear the lock and socket before starting Xvnc.
- Also fixes `test_roundtrip.py`: it claimed to capture password-protected tigervnc and replay without a password, but its `DOCKER_SERVERS` lookup picked `"tigervnc"` (no-auth) instead of `"tigervnc-auth"`. The no-password replay assertion was vacuous — there was never a password to strip. Now points at `tigervnc-auth`, verified passing against the live fleet.

Diagnosed and originally fixed together with an x11vnc/Xvfb start-up race on a local branch (`claude/goofy-mendel-87f383`) that was never pushed as a PR. That branch's other half — clearing `draw-content.sh`'s readiness marker on restart — is now moot: #356 moved readiness off the marker file entirely (`wait_for_servers.py` does a real capture instead), so there's no stale-marker failure mode left to guard against.

## Test plan
- [x] `tests/functional/test_roundtrip.py` run against the live fleet with the fix — passes
- [ ] `make servers-up`, `docker compose -f tests/servers/docker-compose.yml restart tigervnc tigervnc-auth`, confirm both come back healthy
- [ ] `make test-func` against the restarted fleet

### DIFF
--- a/tests/functional/test_roundtrip.py
+++ b/tests/functional/test_roundtrip.py
@@ -11,7 +11,7 @@ from PIL import Image
 from .test_proxy import _await_capture, _start_vnclog, _stop_proxy
 from .vncservers import DOCKER_SERVERS, HOST, port_open, start_replay_server, vncdo_argv
 
-TIGERVNC = next(s for s in DOCKER_SERVERS if s.name == "tigervnc")
+TIGERVNC_AUTH = next(s for s in DOCKER_SERVERS if s.name == "tigervnc-auth")
 
 ROUNDTRIP_PROXY_PORT = 5996
 REPLAY_PORT = 5997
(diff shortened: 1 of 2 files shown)


tests/functional/test_roundtrip.py
```diff
@@ -11,7 +11,7 @@
 from .test_proxy import _await_capture, _start_vnclog, _stop_proxy
 from .vncservers import DOCKER_SERVERS, HOST, port_open, start_replay_server, vncdo_argv
 
-TIGERVNC = next(s for s in DOCKER_SERVERS if s.name == "tigervnc")
+TIGERVNC_AUTH = next(s for s in DOCKER_SERVERS if s.name == "tigervnc-auth")
```
sibson, line 14: how often do we have this line? Should we have a more explicit, simple, common way to get this?

sibson: 🤖 Claude Code: Named in 3921aae — TIGERVNC, TIGERVNC_AUTH, X11VNC, LIBVNCSERVER_EXAMPLE, matching VNCEV's existing pattern. 5 call sites import directly now instead of scanning DOCKER_SERVERS.

PR #370, opened 2026-08-17
ci: publish coverage to Codecov, one flag per tier

sibson:
Follow-up to #366, which measures and reports coverage but stops at the job
summary. That has no history, so the effect of the testing-framework work is
invisible as a trend.

- `coverage-summary.sh` writes `combined/<tier>.xml` alongside the data it
  already combines.
- The coverage job uploads `unit` and `fleet` as separate Codecov flags;
  Codecov merges them for the project number.
- `codecov.yml` turns the project and patch statuses off and suppresses the
  PR comment.
- README badge, DEVELOP.rst points at the browsable report.

`relative_files = True` is the load-bearing bit. Without it the Cobertura XML
carries an absolute `<source>` with `filename="api.py"`, and nothing matching
those against the repo tree finds a file — the upload would have succeeded and
matched nothing. Verified locally that the source is now the relative
`vncdotool`.

**Nothing gates.** Statuses off, `fail_ci_if_error: false`, both steps
`continue-on-error`. A Codecov outage cannot turn a green run red. Introducing
a status check that can fail a PR would undo the point of #366.

`comment: false` follows the decision in #366 to drop the repo's own PR comment
as noise on the conversation — Codecov's comment is the same noise from a
different sender. One line to flip.

Pinned to `codecov-action@v7`: tokenless uploads for public repos landed in v5,
and v6.0.1 fixed a template injection the v5 line never received.

Not included, deliberately: no historical backfill.

Local verification: 180 unit tests pass, flake8 clean, both YAML files parse,
the script produces valid Cobertura with repo-relative paths.

### DIFF
--- a/.github/scripts/coverage-summary.sh
+++ b/.github/scripts/coverage-summary.sh
@@ -7,8 +7,8 @@
 # Writes combined/summary.md -- a table of one number per tier, with the
 # per-file tables folded away underneath it -- and prints it. htmlcov/ is
 # written from the total, combined/<name> holds each tier's combined data,
-# and combined/xml/<name>.xml is the same tier as Cobertura. Nothing here
-# is CI-specific, so the same report can be produced locally.
+# and combined/xml/<name>.xml the same tier as Cobertura. Nothing here is
+# CI-specific, so the same report can be produced locally.
 set -euo pipefail
 
 mkdir -p combined/xml
--- a/.github/scripts/coverage-summary.sh
+++ b/.github/scripts/coverage-summary.sh
@@ -7,8 +7,8 @@
 # Writes combined/summary.md -- a table of one number per tier, with the
 # per-file tables folded away underneath it -- and prints it. htmlcov/ is
 # written from the total, and combined/<name> holds each tier's combined
-# data. Nothing here is CI-specific, so the same report can be produced
-# locally.
+# data alongside combined/<name>.xml for tools that want Cobertura.
+# Nothing here is CI-specific, so the same report can be produced locally.
 set -euo pipefail
 
 mkdir -p combined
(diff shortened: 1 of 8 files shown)


.github/scripts/coverage-summary.sh
```diff
@@ -37,6 +37,9 @@ report() {
         coverage report --data-file="combined/$name" --format=markdown
         printf '\n'
     } >> "$detail"
+    # Not `combined/$name.xml`: that matches `combined/$name.*`, so the next
```
sibson: meaningless

.github/workflows/ci.yml
```diff
@@ -203,3 +207,27 @@ jobs:
           name: coverage-html
           path: htmlcov/
           if-no-files-found: ignore
+
+      # Codecov rejects the documented tokenless upload for this repo, so the
```
sibson: so what!

codecov.yml
```diff
@@ -0,0 +1,17 @@
+# Coverage is reported, never gated: no status check may fail a build.
+# See DEVELOP.rst.
```
sibson: see DEVELOP.rst, no thanks

DEVELOP.rst
```diff
@@ -55,6 +55,19 @@ The numbers are at the top of the run's job summary, one line per tier,
 with the per-file tables folded away underneath and ``coverage-html`` an
 artifact on the same run.
 
+The browsable report is https://app.codecov.io/gh/sibson/vncdotool --
+line-by-line annotation, history, and the ``unit`` and ``fleet`` flags
+graphed separately. CI uploads one Cobertura file per tier there.
+
+Codecov documents tokenless uploads for public repos, but rejects them
```
sibson: who cares, we figured it out and configured correctly now.

setup.cfg
```diff
@@ -49,6 +49,9 @@ parallel = True
 patch = subprocess
 concurrency = thread
 source = vncdotool
+# Without this the XML names files `api.py`, relative to the source dir,
```
sibson: ok, but like that is the who point of the option

PR #377, opened 2026-08-18
Design: pixel format layer for pluggable decoders

sibson:
Phase 1 of the decoder work needs a design; `specs/pixel-format.md` is it, and writing it against measured servers changed two decisions in `decoder-architecture.md`.

**Decoders no longer convert to a canonical layout.** They return the bytes they were sent, tagged with the Pillow raw mode that unpacks them, and `client.py` materializes each rectangle with one `Image.frombytes`. Canonical output materializes every rectangle twice, and forces JPEG, TPIXEL and colour-map decoders to produce a layout that then has to be undone. The payoff is that a decoder never interprets a pixel — a background colour or palette entry is an opaque byte string — so the channel-extraction bugs cannot be written. **R3 is restated over the framebuffer** as a result, and `decoder-goldens.md` says where it is checked.

**The negotiation default becomes sending no `SetPixelFormat` at all.** Every server anyone runs — TigerVNC, x11vnc, UltraVNC, Apple Screen Sharing — announces BGRX at 32bpp depth 24 LE, and today's client already reads all four without renegotiating (measured in #394). rfbproto permits the omission explicitly, and #90/#275 cannot bite a client that never asked. The Apple 3.889 branch goes on the same evidence.

Protocol facts checked against a local rfbproto clone rather than inferred: TPIXEL is *not* CPIXEL (depth exactly 24, 8-bit channels, fixed RGB order — correcting a "believed to follow the same rule" note), CPIXEL's depth ≤ 16 tie-break which RFC 6143 omits, and the rule that a client must not have an outstanding `FramebufferUpdateRequest` when it sends `SetPixelFormat` — which is why `--pixel-format` is connect-time only.

Docs only; no code. The branch was rebuilt on current main — its previous commit predated #386, which landed the same design doc, so rebasing would have conflicted against content already merged. Depends on #394 for the fixture name it references.

### DIFF
--- a/specs/pixel-format.md
+++ b/specs/pixel-format.md
@@ -73,7 +73,7 @@ The tag is a `PixelFormat` and not a Pillow mode string so that decoders speak
 only RFB — they implement a specification written in shifts and maxima, and
 naming `BGRX` there would put the rendering library inside code that has no
 other reason to know it exists. Pillow stays in `client.py` and in `raw_mode`,
-which is also the single place to change if the four unsupported layouts below
+which is also the single place to change if the unsupported layouts below
 ever need a converter. A decoder unit test asserts bytes and a `PixelFormat`,
 with no Pillow in it (R2).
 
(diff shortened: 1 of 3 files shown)


sibson:
@pmhahn FYI, exploring fleshing out the decoder support.  Most the the work in the last week has been about establishing repeatable testing environments to support developing this.  Feedback welcome.

specs/pixel-format.md
```diff
@@ -0,0 +1,264 @@
+# Pixel Format — Design
+
+Status: draft, under review. The first slice of
+[decoder-architecture.md](decoder-architecture.md) Phase 1, and the `rgb565`
+entry point [decoder-goldens.md](decoder-goldens.md) names in its phasing.
+
+## Problem
+
+`client.PF2IM` (`client.py:68`) is a five-entry dict keyed on the whole
```
sibson, line 9: won't this line level link drift?

sibson: 🤖 Claude Code: It does, and two already had — `cpixel()` was cited at `rfb.py:800` and lives at 820; `PF2IM` appears as both `client.py:67` and `:68`. All eleven line-level references across the three specs are now symbol names instead. #398, since this merged first.

PR #380, opened 2026-08-19
Fix Windows console-script resolution in functional test suite

sibson:
## Summary

#378 resolved `vncdo`/`vncdo-replay`/`vnclog` as siblings of `sys.executable`, which holds for venvs on every platform and for macOS's framework build. On `actions/setup-python`'s non-venv Windows interpreter, pip installs console-script `.exe` shims into a sibling `Scripts\` directory instead — so every Windows `os-servers.yml` run since #378 merged has failed `assert_cli_installed()` at import time. This was invisible because that job's Windows leg runs with `continue-on-error: true`.

`_resolve_script()` now checks `Scripts\` alongside the interpreter dir on Windows, falling back to the original single-directory behavior elsewhere.

## Test plan

- [x] `flake8 --count --statistics vncdotool tests` clean
- [x] `.venv/bin/python -m unittest discover tests/unit` — 181 passed
- [x] Verified macOS resolution unchanged (built worktree `.venv`, imported module, paths correct)
- [x] Simulated the Windows `Scripts\` layout standalone and confirmed the resolver finds it

### DIFF
--- a/tests/functional/vncservers.py
+++ b/tests/functional/vncservers.py
@@ -41,11 +41,26 @@ READY_DEADLINE = 180.0
 # Named by full path rather than resolved through PATH: another working tree's
 # .venv can come first there, and the suite would report its behaviour as this
 # branch's.
-_SCRIPTS = Path(sys.executable).parent
+_INTERPRETER_DIR = Path(sys.executable).parent
+# venvs put console scripts beside the interpreter on every platform; a
+# non-venv Windows install (e.g. actions/setup-python on windows-latest) puts
+# them in a sibling Scripts\ instead.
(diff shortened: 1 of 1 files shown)


sibson:
🤖 Claude Code: Closing without merging — root cause fixed upstream by #384.

#384 switched the Windows os-servers.yml job from a bare actions/setup-python interpreter to `uv sync`/`uv run`, which creates a real venv. On Windows, uv puts python.exe and the installed console-script .exe shims in the same .venv\Scripts\ directory, so the original single-directory `Path(sys.executable).parent` lookup from #378 already resolves correctly.

Verified against CI: on main at #384 (68b181a), without this branch's fix, the Windows job's "Wait for the server to serve a screen" step — the exact step that raised AssertionError on every run since #378 — passes green (run 32313077322).

PR #382, opened 2026-08-19
build: move packaging to pyproject.toml with hatchling

sibson:
Packaging collapses into one pyproject.toml on hatchling: setup.py, setup.cfg, MANIFEST.in, tox.ini and both requirements files go, and dependencies stop being declared twice. flake8 config lands in `.flake8` since flake8 cannot read pyproject; coverage config moves into pyproject.

Version is now static in `[project]` and bumped with `uv version`, and `vncdotool.__version__` reads the installed metadata, so nothing rewrites a Python source file. That also fixes `make release` reading both version numbers by running bare `python`, which need not exist — an empty `VERSION` would have tagged the release `v`.

Worth a look: the rpm spec now uses the pyproject macros (untested, and already stale at 1.2.0), and PyPI's long description is README.rst alone, where CHANGELOG.rst used to be appended.

Tested: `uv build` with the wheel and sdist inspected, fresh install, 181 unit tests, flake8 clean.


docs/install.rst
```diff
@@ -13,8 +13,7 @@ and if you aren't already using it, and you should be, use a virtualenv_.::
 
     pip install virtualenv
     virtualenv venv-vncdotool
```
sibson: needs to be updated to uv?

Makefile
```diff
@@ -1,7 +1,9 @@
 #!/usr/bin/make -f
 .DEFAULT: help
 
-REQUIREMENTS_TXT?=requirements-dev.txt
+# Makefile.venv installs the package from pyproject.toml; the dev
```
sibson: no longer exists

Makefile
```diff
@@ -17,20 +19,23 @@ help:
 	@echo "docs:		build documentation"
 	@echo "release:	tag and push current version to trigger PyPI release"
 
-VERSION := $(shell python -c "import vncdotool; print(vncdotool.__version__.split('.dev')[0])")
-NEXT_VERSION := $(shell python -c "v='$(VERSION)'.split('.'); v[-1]=str(int(v[-1])+1); print('.'.join(v)+'.dev0')")
+# --dry-run reports what a bump would write without writing it, so both
```
sibson: that is the definition of dry run

pyproject.toml
```diff
@@ -0,0 +1,93 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "vncdotool"
+version = "1.4.0.dev0"
```
sibson: bump to 2.0 to capture breaking changes, new packaging is just the start

pyproject.toml
```diff
@@ -0,0 +1,93 @@
...
+
+[project.scripts]
+vncdo = "vncdotool.command:vncdo"
+vncdotool = "vncdotool.command:vncdo"
+vnclog = "vncdotool.command:vnclog"
+vncdo-replay = "vncdotool.command:vncdo_replay"
+
+[dependency-groups]
+dev = [
+    # 7.10 for `patch = subprocess`, which measures the CLI the functional
```
sibson: inline comment after the comma for coverage

pyproject.toml
```diff
@@ -0,0 +1,93 @@
...
+    # tests shell out to.
+    "coverage>=7.10",
+    "flake8",
+    "sphinx",
+]
+
+[tool.hatch.build.targets.wheel]
+packages = ["vncdotool"]
+
+# Hatchling's default sdist is everything git tracks; this keeps it to what
```
sibson: narration

pyproject.toml
```diff
@@ -0,0 +1,93 @@
...
+    "tests",
+    "docs",
+    "*.rst",
+    "*.txt",
+    "pyproject.toml",
+]
+
+[tool.coverage.run]
+branch = true
+# Most of the data arrives from `vncdo` subprocesses, each writing its own
```
sibson: wat?

pyproject.toml
```diff
@@ -0,0 +1,93 @@
...
+# file; see DEVELOP.rst.
+parallel = true
+patch = ["subprocess"]
+concurrency = ["thread"]
+source = ["vncdotool"]
+relative_files = true
+
+[tool.coverage.paths]
+# CI installs editable, so both spellings are the same checkout.
```
sibson: Do we really need both here?

docs/install.rst
sibson: 🤖 Claude Code: Done. The source-checkout block is `uv sync` now; the virtualenv link target went with it. Left `pip install vncdotool` at the top and the Windows walkthrough alone — installing the published package has not changed.

Makefile
sibson: 🤖 Claude Code: Dropped the comment. `REQUIREMENTS_TXT=` and `venv-dev` are transitional — Makefile.venv is still here in this PR and #383 deletes it along with both of these.

Makefile
sibson: 🤖 Claude Code: Deleted.

pyproject.toml
sibson: 🤖 Claude Code: Now `2.0.0.dev0`, and the CHANGELOG UNRELEASED heading follows it. `make release` then produces 2.0.0 and bumps to 2.0.1.dev0.

pyproject.toml
sibson: 🤖 Claude Code: Inlined.

pyproject.toml
sibson: 🤖 Claude Code: Deleted. The include list stands on its own.

pyproject.toml
sibson: 🤖 Claude Code: Carried over verbatim from setup.cfg, and it was doing two jobs badly. Cut to what the setting needs: `# Each `\vncdo' subprocess writes its own data file.` directly above `parallel`. DEVELOP.rst already covers the rest.

pyproject.toml
sibson: 🤖 Claude Code: No — dropped the whole `[tool.coverage.paths]` section. Every environment that measures this package installs it editable, so the data only ever carries the `vncdotool` spelling. Verified combine + report still work.

docs/install.rst
```diff
@@ -9,13 +9,12 @@ vncdotool relies on a number of libraries, the two major ones are Pillow_, the P
...
 If you are having issues getting things to work you can try using a stable set of libraries
-and if you aren't already using it, and you should be, use a virtualenv_.::
+and if you aren't already using it, and you should be, use uv_.::
 
-    pip install virtualenv
-    virtualenv venv-vncdotool
-    # XXX requirements.txt from vncdotool source tree
-    pip install -r requirements.txt
-    pip install -e .
+    uv sync
```
sibson: You should have pushed back on this, we should follow python defaults of pip.  Never mind the source checkout, folks doing that should be able to figure it out themselves.

sibson: 🤖 Claude Code: Reverted — the block is virtualenv + `pip install vncdotool` again, and the uv link target is gone. You are right that I should have pushed back; the only thing actually broken there was the dead `requirements.txt` line, and swapping the installer was scope I invented. Nothing else in docs/ mentions uv.

sibson: 🤖 Claude Code: Dropped the virtualenv block entirely — `pip install vncdotool` is the whole instruction now, and how to isolate it is the user's call. The virtualenv link target went with it. The paragraph about Pillow/Twisted versions stays, minus the recipe.

PR #390, opened 2026-08-19
ci: pull the test server fleet from GHCR, and stop waiting on SIGTERM-deaf containers

sibson:
Two things making the `servers` job cheaper, plus a note.

**Pull the fleet instead of rebuilding it.** A hosted runner starts with an empty Docker cache, so every run rebuilt all five images (~50s). A new `fleet-images.yml` publishes them to GHCR from `main` whenever `tests/servers` changes, tagged with that directory's git tree hash, and the `servers` job pulls that tag. A pull miss falls back to building, so a PR that edits the fleet (its images don't exist until it lands) and any fork PR behave exactly as before. Local `make servers-up` sets neither override and still builds from source.

**Teardown.** `vncev` and `libvncserver-example` run their upstream binary as PID 1 with no signal handling, so they sat out the full 10s stop grace period; the other three exit in under a second via their entrypoint traps. Timeout drops to 2s.

**One thing needed after merge:** the GHCR packages are created private, and the anonymous pull needs them public — otherwise the pull just fails and every run quietly pays for a build, i.e. today's behaviour. Four packages, `vncdotool-test-{tigervnc,x11vnc,vncev,libvncserver-example}`.

I also tried `type=gha` layer caching, which this repo's follow-up list has proposed twice. It measured slower than plain `docker compose --build` in all four configurations (68–106s against ~50s), so none of it is here — just a note under that item recording why, so it isn't attempted a third time.

Tests: CI green on this branch; the build-fallback path is what ran, since nothing is published yet.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,9 +135,13 @@ jobs:
           path: screenshots/
           if-no-files-found: ignore
 
+      # vncev and libvncserver-example run their upstream binary as PID 1
+      # with no SIGTERM handling, so the default 10s stop grace period is
+      # pure dead time -- the tigervnc/x11vnc entrypoints' traps already
+      # exit in well under a second. Nothing here needs a clean shutdown.
       - name: Stop the VNC test servers
         if: always()
-        run: docker compose -f tests/servers/docker-compose.yml down
+        run: docker compose -f tests/servers/docker-compose.yml down --timeout 2
 
   coverage:
     name: Coverage
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,6 +77,7 @@ jobs:
     timeout-minutes: 20
     env:
       VNCDOTOOL_SCREENSHOT_DIR: screenshots
+      FLEET_IMAGE_PREFIX: ghcr.io/${{ github.repository_owner }}/vncdotool-test
     steps:
       - name: Check out source
         uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -89,8 +90,18 @@ jobs:
       - name: Install dependencies
         run: uv sync
 
-      - name: Build and start the VNC test servers
-        run: docker compose -f tests/servers/docker-compose.yml up -d --build --wait
+      # Images come from .github/workflows/fleet-images.yml. A pull miss is
```
sibson: overkill

.github/workflows/ci.yml
```diff
@@ -135,9 +146,13 @@ jobs:
           path: screenshots/
           if-no-files-found: ignore
 
+      # vncev and libvncserver-example run their upstream binary as PID 1
```
sibson: belongs in commit message not comment

.github/workflows/fleet-images.yml
```diff
@@ -0,0 +1,65 @@
+name: Fleet images
+
+# The tag is deliberately coarser than the build inputs: anything under
+# tests/servers rebuilds, including scripts that reach no image. Narrowing
```
sibson: over explaining

tests/servers/docker-compose.yml
```diff
@@ -11,6 +11,9 @@
 # Readiness is the HEALTHCHECK baked into the image, which makes
 # `up --wait` a reliable barrier. Every service publishes its RFB port to
 # localhost only.
+#
+# CI sets FLEET_IMAGE_PREFIX and FLEET_TAG to pull these prebuilt rather
```
sibson: explains design again

.github/workflows/ci.yml
sibson: 🤖 Claude Code: Deleted. `||` already says the fallback is intentional.

.github/workflows/ci.yml
sibson: 🤖 Claude Code: Deleted. Already in 3e54d60's body.

.github/workflows/fleet-images.yml
sibson: 🤖 Claude Code: Header deleted. The tag rationale is in the commit body; the apt-pinning consequence had no other home, so it moved to the spec item rather than staying here.

tests/servers/docker-compose.yml
sibson: 🤖 Claude Code: Deleted. Third description of the same design; the `:-` defaults show the rest.

PR #391, opened 2026-08-19
ci: promote os-servers.yml from report-only to blocking

sibson:
Drops `continue-on-error: true` from the Tier 2 Windows/macOS job now that it's proven stable: macOS/Screen Sharing is 32/32 green across its whole history, and Windows/UltraVNC's only failures trace to the mid-uv-migration branches missing `uv run` (fixed by #384) -- clean before and after.

Dropped the earlier idea of adding windows-latest/macos-latest to the unit test matrix -- no value there, the protocol classes are driven against a mocked transport with nothing OS-specific to catch.

## Test plan
- [x] Reviewed os-servers.yml run history (32/32 macOS, Windows clean pre- and post-#384)
- [ ] CI green on this PR

### DIFF
--- a/specs/server-compatibility-plan.md
+++ b/specs/server-compatibility-plan.md
@@ -256,8 +256,8 @@ spin them up when something relevant changes:
   VM, and a quiet repository consumes nothing.
 - Accepted trade-off: drift in the runner images or the Chocolatey
   package surfaces on the next change-triggered run rather than the night
-  it happens, and now blocks the PR that triggered it like any other
-  CI failure.
+  it happens, and blocks the PR that triggered it like any other CI
+  failure.
 - Each successful run can upload a wire capture as an artifact for
   offline debugging.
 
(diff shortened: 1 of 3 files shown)


specs/server-compatibility-plan.md
```diff
@@ -423,10 +422,16 @@ service-mode setup steps and move passwords into repository secrets.
...
 `.github/workflows/os-servers.yml` (renamed from `spike-os-servers.yml`),
-change-triggered and path-filtered per the policy above, report-only via
-`continue-on-error: true`. Credentials come from `VNC_OS_SERVER_USERNAME`/
-`VNC_OS_SERVER_PASSWORD` repository secrets, falling back to the spike
-values when unset.
+change-triggered and path-filtered per the policy above. Credentials come
+from `VNC_OS_SERVER_USERNAME`/`VNC_OS_SERVER_PASSWORD` repository secrets,
+falling back to the spike values when unset.
+
+**Promoted to blocking (2026-08-19):** macOS was 32/32 green across the
```
sibson: this is historical commentery

sibson: 🤖 Claude Code: Fair — dropped that paragraph, folded the current state into the existing Graduated line. The stability evidence stays in the commit message (3df79d1) instead.

PR #394, opened 2026-08-20
Name the golden fixture for its pixel format, and log the format we request

sibson:
The golden fixture was called `tigervnc-raw-rgb888`, but the ServerInit it recorded is 32bpp depth-24 little-endian with shifts 16/8/0 — BGRX on the wire. Renamed `tigervnc-raw-bgrx8888`, in a vocabulary the coming `--pixel-format` flag will share: channel order as the bytes arrive, per-channel widths, `x` for a pad byte.

`rfb.py` logged the pixel format a server announced but never the one we ask it for, which is half a bug report. `vncdo -v` now prints both.

Those two together measured something worth having: every server anyone runs — TigerVNC, x11vnc, UltraVNC, Apple Screen Sharing — announces BGRX at 32bpp depth 24 LE, and none of them renegotiates today. The numbers are recorded in `specs/pixel-format.md` (#377) and the probe test that produced them is removed in the last commit: it spent a connection per server per run to re-derive a value that only changes when a runner image does, and asserted nothing about it.

Also moves the rfbproto clone into `docs/rfbproto` per review, git-ignored.

Tested: unit suite green; fleet functional suite green locally; the OS-server jobs ran green on both Windows and macOS while the probe was in place.


specs/decoder-goldens.md
```diff
@@ -138,7 +138,7 @@ label.
 
 A fixture is a whole capture session, not a single update:
 
-    tests/unit/fixtures/goldens/tigervnc-raw-rgb888/
+    tests/unit/fixtures/goldens/tigervnc-raw-native/
```
sibson: native could change or be anything, let's call it what it was at capture or omit

tests/functional/vncservers.py
```diff
@@ -363,6 +375,21 @@ def test_mousemove(self) -> None:
         """A pointer event is accepted without the server dropping the session."""
         self.run_vncdo_ok("move", "10", "10")
 
+    def test_pixel_format(self) -> None:
```
sibson: should this be built into client logging by default rather than awkwardly in a test?

DEVELOP.rst
```diff
@@ -101,7 +101,10 @@ what vncdotool implements, it is not.
 
 rfbproto is a living document with no releases or version numbers, so link
 a commit permalink rather than ``master`` when a comment depends on its
-wording. Neither document describes how any particular server actually
+wording. It is one grep-able reStructuredText file, so keep a clone beside
+this one and read it locally rather than fetching sections over the web::
+
+    git clone https://github.com/rfbproto/rfbproto.git ../rfbproto Neither document describes how any particular server actually
```
sibson: let's put under our docs directory ratyher than polluting outside our tree

specs/decoder-goldens.md
sibson: 🤖 Claude Code: Renamed to `tigervnc-raw-bgrx8888` — the format it was captured at (32bpp, depth 24, LE, shifts 16/8/0). Agreed on "native": it names how the format was chosen, so the same directory would quietly mean something else after a server or image change. Naming vocabulary is shared with the `--pixel-format` flag: channel order as the bytes arrive, per-channel widths, `x` for a pad byte.

tests/functional/vncservers.py
sibson: 🤖 Claude Code: Keeping it behind `-v`, per your call. The client does log it — `rfb.py` logs the native format at ServerInit, and this PR adds the format we request — but `vncdo` calls `logging.basicConfig()` with no level, so INFO only surfaces at `-v`. Putting a pixel format on every default run would be noise.

So the test is not where the logging lives; it is what makes CI run with `-v` and record the line per server, which is how UltraVNC and Screen Sharing got measured. Docstring trimmed in the sweep.

Separately: a failed connection currently prints `CRITICAL:root:Connection was refused...`, which is logging leaking into the CLI surface. Filed as its own task, not folded in here.

DEVELOP.rst
sibson: 🤖 Claude Code: Moved — `git clone https://github.com/rfbproto/rfbproto.git docs/rfbproto`, with `docs/rfbproto/` added to `.gitignore`. Also repaired that paragraph: the earlier edit had glued the following sentence onto the clone command line.

PR #395, opened 2026-08-20
Report vncdo failures as CLI errors, not CRITICAL log records

sibson:
`vncdo` reported a refused connection as `CRITICAL:root:Connection was refused by other side: 61: Connection refused.` — the root logging handler's default format applied to what is an ordinary CLI failure.

`VNCDoCLIFactory.error` now prints the message to stderr the way the usage-error path in `build_command_list` already does, and the failure's traceback drops to DEBUG, where `-vv` picks it up. `setup_logging` gains a `format` so the remaining `-v`/`-vv` diagnostics lose the `:root:` logger name too.

Exit statuses are untouched — verified 10 for a refused connection.

Worth a look: the `setUp` change in `TestVNCDoCLIFactory` captures stderr for the whole class, otherwise the pre-existing `error()` tests spray the new message into the test run's own output.

Tested: `make test` (211 pass), `flake8 --count --statistics vncdotool tests` clean, and `vncdo -s 127.0.0.1::1 move 10 10` by hand with and without `-vv`.

### DIFF
--- a/tests/unit/test_command.py
+++ b/tests/unit/test_command.py
@@ -1,3 +1,5 @@
+import io
+import logging
 import os
 import socket
 import tempfile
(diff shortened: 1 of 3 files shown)


tests/unit/test_command.py
```diff
@@ -336,6 +338,10 @@ class TestVNCDoCLIFactory(unittest.TestCase):
 
     def setUp(self) -> None:
         self.factory = command.VNCDoCLIFactory()
+        self.stderr = io.StringIO()
```
sibson, line 341: decorator?

PR #396, opened 2026-08-20
CI: fix the fleet publish race, and cut the macOS job from 2m37 to 24s

sibson:
Two CI costs, and the safety rework that came out of fixing the second.

**The fleet publish race.** `fleet-images.yml` published on the same `main` push that `ci.yml` tested, so the run that introduced a `tests/servers` change always lost the pull and rebuilt (four `manifest unknown`, run 32367992392). Publishing moves into the `servers` job that discovers the miss, and that workflow is deleted — nothing can pull a tag before its own build. It runs last and without `if: always()`, so a fleet only reaches GHCR after passing the suite it exists to serve, and only from `main`.

**The macOS job: 2m37 → 24s.** The job's own `log show` artifacts showed the ARD connection fast-user-switching into a dedicated account's first login — `UserAccountUpdater` 28s and 33s across two runs, Setup Assistant behind it, `loginwindow` never reaching `LoginComplete`. Authenticating as the console owner skips all of it. That account's password can't be *set* (secure token: `sysadminctl` fails while exiting 0, `dscl` returns `eDSAuthFailed`) but it doesn't need to be — these images enable auto-login, so macOS keeps the password in `/etc/kcpassword` XORed against a published key, and the account logging itself in at boot proves the file matches. The setup decodes it, masks it, and gates it on `dscl -authonly`. No fallback: an image that stops auto-logging in should be a red build, not a silently slower one.

**Setup moved out of `tests/servers`.** Those scripts turned any machine that ran them into an unattended remote-control target, and deleting a checkout didn't undo it. Setup now lives in `.github/actions/os-server` as scripts the action drives, each calling `require-disposable-host.sh` immediately before the first thing that changes the machine. There is no longer any path that creates or repasswords an account — Screen Sharing is reachable only as whoever already owns the console — so on a developer machine, which has no `/etc/kcpassword`, the script stops before touching anything.

Reviewer notes: green except CodeQL, which flags `decode-kcpassword.py:52` writing the decoded value to stdout. That alert is correct — a password crosses a process boundary — and is the deliberate design: stdout is the only channel a `sudo` child has, the value is masked before anything can echo it, and the runner is ephemeral. It needs dismissing with a reason, or the decode rewriting in shell where CodeQL has no query for it. Also worth a look: `ci.yml` now takes `packages: write` on one job, and credentials moved out of the workflow env into the action, reaching the tests via `$GITHUB_ENV`.


.github/actions/os-server/action.yml
```diff
@@ -0,0 +1,57 @@
+name: Set up an OS-hosted VNC server
+description: >
+  Turn the runner into a live VNC server for tests/functional/test_os_servers.py:
+  UltraVNC as a Windows service, or Apple Screen Sharing on macOS.
+
+# The work is in the scripts beside this file. Each calls
```
sibson: we're describing the code

.github/actions/os-server/action.yml
```diff
@@ -0,0 +1,57 @@
...
+# require-disposable-host.sh before the first thing that changes the machine,
+# so running one directly is refused the same way.
+#
+# What each server needs, and what it does and doesn't prove, is documented
```
sibson: decribing the code

.github/actions/os-server/action.yml
```diff
@@ -0,0 +1,57 @@
...
+# next to the diagnostic scripts in tests/servers/ultravnc/README.md and
+# tests/servers/screen-sharing/README.md.
+
+inputs:
+  server:
+    description: Which OS-hosted server to set up -- ultravnc or screen-sharing.
```
sibson: Would macos be better than screen-sharing

.github/actions/os-server/require-disposable-host.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+# Refuse to go on unless this machine is a disposable GitHub-hosted runner.
```
sibson: decdribes the code

.github/actions/os-server/require-disposable-host.sh
```diff
@@ -0,0 +1,38 @@
...
+#
+# The setup scripts next to this one leave the machine they run on reachable
+# for unattended remote control, and deleting a checkout undoes none of it:
+# macOS keeps the Remote Management setting, Windows keeps the UltraVNC
+# service.
+#
+# RUNNER_ENVIRONMENT is the load-bearing one -- it separates a hosted runner
```
sibson: should be local to the test if it's even needed

.github/actions/os-server/require-disposable-host.sh
```diff
@@ -0,0 +1,38 @@
...
+# from a self-hosted one, which is somebody's real machine. The rest catch a
+# local `act` or a step copied out of the workflow.
+set -uo pipefail
+
+# For exercising the setup against a VM. Deliberately a phrase rather than a
+# 1, so that nothing sets it by accident and no copy-pasteable command in the
+# docs contains it.
+if [ "${VNCDOTOOL_OS_SERVER_DISPOSABLE_HOST:-}" = "yes-destroy-this-machine" ]; then
```
sibson: this is super confusing, how does this envvar get set? Why just exit 0

.github/actions/os-server/require-disposable-host.sh
```diff
@@ -0,0 +1,38 @@
...
+# The setup scripts next to this one leave the machine they run on reachable
+# for unattended remote control, and deleting a checkout undoes none of it:
+# macOS keeps the Remote Management setting, Windows keeps the UltraVNC
+# service.
+#
+# RUNNER_ENVIRONMENT is the load-bearing one -- it separates a hosted runner
+# from a self-hosted one, which is somebody's real machine. The rest catch a
+# local `act` or a step copied out of the workflow.
+set -uo pipefail
+
```
sibson: why not make this a function in the main script?

.github/actions/os-server/macos.sh
```diff
@@ -0,0 +1,119 @@
...
+
+HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+KICKSTART=/System/Library/CoreServices/RemoteManagement/ARDAgent.app/Contents/Resources/kickstart
+
+PORT="${PORT:-5900}"
+WAIT_SECONDS="${WAIT_SECONDS:-60}"
+CAFFEINATE_SECONDS="${CAFFEINATE_SECONDS:-1200}"
+
+KCPASSWORD=/etc/kcpassword
+# Auto-login stores the password obfuscated rather than hashed, because
```
sibson, line 14: verbose

.github/actions/os-server/macos.sh
```diff
@@ -0,0 +1,119 @@
...
+# loginwindow has to replay it: XORed against this key, repeated, and
+# NUL-padded to a multiple of its length.
+KCPASSWORD_KEY=(125 137 82 35 210 188 221 234 163 185 31)
+
+# Reads the file's bytes as decimals on stdin and prints the password.
```
sibson, line 29: over verbose

.github/actions/os-server/screen-sharing.sh
```diff
@@ -0,0 +1,119 @@
...
+    done
+
+    if [ -z "$password" ]; then
+        echo "$KCPASSWORD decoded to nothing" >&2
+        return 1
+    fi
+    printf '%s' "$password"
+}
+
+# The console owner's password is read rather than set because it cannot be
```
sibson: why are we talking about setting? That is a design artifact

.github/actions/os-server/ultravnc.ps1
```diff
@@ -0,0 +1,127 @@
+<#
+.SYNOPSIS
+    Install, configure and start UltraVNC as a Windows service.
+
+.DESCRIPTION
+    Gives tests/functional/test_os_servers.py a live Windows server on
```
sibson: unessesary contenxt

.github/actions/os-server/screen-sharing.sh
```diff
@@ -0,0 +1,119 @@
...
+    echo "recovered an empty password without an error" >&2
+    exit 1
+fi
+echo "::add-mask::$PASSWORD"
+echo "--- authenticating as the console owner $USERNAME, no user switch"
+
+# Everything above this line reads. Everything below it changes the machine,
+# which is why the check sits here rather than at the top: Remote Management
+# gets switched on, and a checkout deletion does not switch it back off.
+bash "$HERE/require-disposable-host.sh"
```
sibson: seems wonky we're calling bash from a ps1 script

.github/workflows/ci.yml
```diff
@@ -75,6 +75,11 @@ jobs:
     name: Test server functional tests
     runs-on: ubuntu-latest
     timeout-minutes: 20
+    # packages: write so the job that discovers a missing fleet image is also
```
sibson: unessesary

.github/workflows/ci.yml
```diff
@@ -146,6 +155,29 @@ jobs:
         if: always()
         run: docker compose -f tests/servers/docker-compose.yml down --timeout 2
 
+      # Publishing runs last and without `if: always()`, so a fleet only
```
sibson: overkill describes code

.github/workflows/os-servers.yml
```diff
@@ -8,10 +8,10 @@ name: OS-hosted VNC servers
 # change-triggered (path-filtered, no schedule).
 #
 # What each server needs, and what it does and doesn't prove, is documented
-# next to the setup scripts in tests/servers/ultravnc/README.md and
-# tests/servers/screen-sharing/README.md. This file only wires those scripts
-# and tests/functional/test_os_servers.py together: the two OSes differ by
-# which script sets the server up and which shell runs it, so they are one
+# in tests/servers/ultravnc/README.md and tests/servers/screen-sharing/README.md.
```
sibson, line 11: describes code flow, why?

.github/actions/os-server/macos.sh
```diff
@@ -0,0 +1,120 @@
...
+    # decode_kcpassword already fails on an empty password, so this catches
+    # something in the pipeline before it exiting 0 with nothing to show.
+    echo "recovered an empty password without an error" >&2
+    exit 1
+fi
+echo "::add-mask::$PASSWORD"
+echo "--- authenticating as the console owner $USERNAME, no user switch"
+
+# Everything above this line only reads.
+require_disposable_host
```
sibson: we should just do this first to avoid unnesseary decode of password and exit early

.github/actions/os-server/screen-sharing.sh
```diff
@@ -0,0 +1,120 @@
...
+# tests/functional/vncservers.py reads these.
+if [ -n "${GITHUB_ENV:-}" ]; then
+    {
+        echo "VNCDOTOOL_OS_SERVER_USERNAME=$USERNAME"
+        echo "VNCDOTOOL_OS_SERVER_PASSWORD=$PASSWORD"
+    } >>"$GITHUB_ENV"
+fi
+
+# A password that does not authenticate otherwise surfaces as three minutes of
+# readiness retries against a server that was never the problem.
```
sibson: we know this is the right password because we docoded it., check is surperfluous

.github/actions/os-server/ultravnc.ps1
```diff
@@ -0,0 +1,115 @@
...
+# under C:\Program Files instead: a recursive scan of that tree takes over
+# five minutes on the loaded runner image.
+$InstallDir = 'C:\Program Files\uvnc bvba\UltraVNC'
+$WinVnc = Join-Path $InstallDir 'winvnc.exe'
+# Some UltraVNC builds read the ini from ProgramData rather than from the
+# install directory, so it gets written to both.
+$ProgramDataDir = 'C:\ProgramData\uvnc bvba\UltraVNC'
+
+Write-Host '--- installing UltraVNC'
+# The community feed intermittently answers with something that isn't valid
```
sibson: less verbose comment, in the retry block

.github/actions/os-server/ultravnc.ps1
```diff
@@ -0,0 +1,115 @@
...
+    }
+    Write-Host "::warning::UltraVNC install attempt $attempt did not produce $WinVnc"
+    Start-Sleep -Seconds (5 * $attempt)
+}
+if (-not $installed) {
+    throw "winvnc.exe not found at $WinVnc after 3 install attempts"
+}
+
+Write-Host '--- writing ultravnc.ini'
+# The helper writes the hex to a file rather than to stdout, so the value
```
sibson: explains code

.github/actions/os-server/ultravnc.ps1
```diff
@@ -0,0 +1,115 @@
...
+"@
+
+$iniPath = Join-Path $InstallDir 'ultravnc.ini'
+Set-Content -Path $iniPath -Value $ini -Encoding ASCII
+New-Item -ItemType Directory -Force -Path $ProgramDataDir | Out-Null
+Copy-Item $iniPath (Join-Path $ProgramDataDir 'ultravnc.ini') -Force
+Write-Host "wrote $iniPath (port $Port, password set)"
+
+Write-Host '--- installing and starting the UltraVNC service'
+# Out-Host, or winvnc's own console output would end up in the pipeline
```
sibson: winvnc?

PR #397, opened 2026-08-21
Add pixelformat.py: negotiated format to Pillow raw mode, CPIXEL widths

sibson:
Phase 1 of `specs/pixel-format.md` (#377). New module and its tests only — no call sites, `PF2IM` and `setImageMode` still in place, nothing user-visible, so no CHANGELOG entry.

`raw_mode()` computes the Pillow mode from the format's fields instead of matching a whole dataclass against a table. That difference is the point: libvncserver-example announces the layout `PF2IM` calls `RGBX` but declares depth 32, so it misses the table and we renegotiate to a format byte-identical to the one already arriving. `raw_mode` ignores `depth`; `cpixel_bytes` obeys it, since the encoder used the same declaration.

Every mode string is a claim about Pillow's behaviour, so the tests pack a primary using the format's own shifts, decode through the resolved mode, and assert the pixel that comes back — asserting *which string we picked* would prove nothing. Two things fell out of testing that way: Pillow has `RGB;4B` and no mirror of it, so 444-with-red-high has no mode at all; and `cpixel_offset` is a byte offset while CPIXEL's placement rule is about value bits, so big-endian reverses which end the three bytes sit at. The first cut returned the same offset either way, which slices off a channel and reads the pad byte. Both are fixed here and corrected in the spec on #377.

Tested: 228 unit tests green, flake8 clean.

### DIFF
--- a/tests/unit/test_pixelformat.py
+++ b/tests/unit/test_pixelformat.py
@@ -1,0 +1,179 @@
+from unittest import TestCase
+
+from PIL import Image
+
+from vncdotool import rfb
+from vncdotool.pixelformat import UnsupportedPixelFormat, cpixel_bytes, cpixel_offset, raw_mode
+
+# 32bpp, 8-bit channels, one PixelFormat per Pillow raw mode, plus a
+# depth-only variant and a bigendian variant of bgrx8888.
+BGRX8888 = rfb.PixelFormat(32, 24, False, True, 255, 255, 255, 16, 8, 0)
+BGRX8888_DEPTH32 = rfb.PixelFormat(32, 32, False, True, 255, 255, 255, 16, 8, 0)
+BGRX8888_BIGENDIAN = rfb.PixelFormat(32, 24, True, True, 255, 255, 255, 16, 8, 0)
+RGBX8888 = rfb.PixelFormat(32, 24, False, True, 255, 255, 255, 0, 8, 16)
+XRGB8888 = rfb.PixelFormat(32, 24, False, True, 255, 255, 255, 8, 16, 24)
+XBGR8888 = rfb.PixelFormat(32, 24, False, True, 255, 255, 255, 24, 16, 8)
+
+RGB24 = rfb.PixelFormat(24, 24, False, True, 255, 255, 255, 0, 8, 16)
+BGR24 = rfb.PixelFormat(24, 24, False, True, 255, 255, 255, 16, 8, 0)
+
+BGR565 = rfb.PixelFormat(16, 16, False, True, 31, 63, 31, 11, 5, 0)
+RGB565 = rfb.PixelFormat(16, 16, False, True, 31, 63, 31, 0, 5, 11)
+BGR555 = rfb.PixelFormat(16, 15, False, True, 31, 31, 31, 10, 5, 0)
+RGB555 = rfb.PixelFormat(16, 15, False, True, 31, 31, 31, 0, 5, 10)
+RGB444 = rfb.PixelFormat(16, 12, False, True, 15, 15, 15, 0, 4, 8)
+
+
+def pack(pixel_format: rfb.PixelFormat, red: int, green: int, blue: int) -> bytes:
+    """Encode (red, green, blue) as PIXEL bytes for pixel_format's own layout."""
+    value = (
+        (red << pixel_format.redshift)
+        | (green << pixel_format.greenshift)
+        | (blue << pixel_format.blueshift)
+    )
+    return value.to_bytes(pixel_format.bpp // 8, "big" if pixel_format.bigendian else "little")
+
+
+class TestRawMode(TestCase):
+
+    def test_resolves_every_format_pillow_can_read(self):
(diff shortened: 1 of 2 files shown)


tests/unit/test_pixelformat.py
```diff
@@ -0,0 +1,178 @@
...
+        (red << pixel_format.redshift)
+        | (green << pixel_format.greenshift)
+        | (blue << pixel_format.blueshift)
+    )
+    return value.to_bytes(pixel_format.bpp // 8, "big" if pixel_format.bigendian else "little")
+
+
+class TestRawMode(TestCase):
+
+    def test_resolves_every_format_pillow_can_read(self):
```
sibson, line 73: should this be a generative test?

tests/unit/test_pixelformat.py
```diff
@@ -0,0 +1,178 @@
...
+            BGR565,
+            RGB565,
+            BGR555,
+            RGB555,
+            RGB444,
+        ]
+        for pixel_format in formats:
+            mode = raw_mode(pixel_format)
+            with self.subTest(pixel_format=pixel_format, mode=mode):
+                red = pack(pixel_format, pixel_format.redmax, 0, 0)
```
sibson: should this be a global constant?

tests/unit/test_pixelformat.py
sibson: 🤖 Claude Code: Yes — done in 33dfdf9. `every_layout()` permutes the shift slots across channels for each width set, both endiannesses, so the check covers every arrangement rather than the twelve I thought to list. The property is: a mode that resolves must decode the primaries it claims, a layout Pillow cannot read must raise, no third outcome.

It failed immediately, which is the argument for it. `raw_mode` picked a 16bpp mode by asking whether red or blue sat higher — that says nothing about the channel between them, so 565 with red at shift 5 and green at 0 overlaps at bit 5, resolved to `RGB;16`, and decoded red as green. Now matched as whole (widths, shifts) triples against the arrangements Pillow actually has.

Kept the explicit table above it: that one pins the exact mode strings, and generating its expectations would compute them the same way the code does.

tests/unit/test_pixelformat.py
sibson: 🤖 Claude Code: Gone rather than hoisted — the generative rewrite in 33dfdf9 removed that list entirely, since `every_layout()` now produces the cases. The module-level constants that remain are the ones named tests refer to individually (`BGRX8888`, `BGR565`, and so on).

tests/unit/test_pixelformat.py
```diff
@@ -0,0 +1,198 @@
...
+        resolved = set()
+
+        for pixel_format in every_layout():
+            with self.subTest(pixel_format=pixel_format):
+                try:
+                    mode = raw_mode(pixel_format)
+                except UnsupportedPixelFormat:
+                    continue
+                resolved.add(mode)
+                for channel, expected in enumerate(((255, 0, 0), (0, 255, 0), (0, 0, 255))):
```
sibson: generative?

tests/unit/test_pixelformat.py
```diff
@@ -0,0 +1,198 @@
...
+        low = rfb.PixelFormat(32, 24, True, True, 255, 255, 255, 16, 8, 0)
+        high = rfb.PixelFormat(32, 24, True, True, 255, 255, 255, 24, 16, 8)
+
+        self.assertEqual(cpixel_offset(low), low.bypp - 3)
+        self.assertEqual(cpixel_offset(high), 0)
+
+    def test_the_three_bytes_carry_every_colour_bit(self):
+        """Slicing at the reported offset must not take the pad byte."""
+        for pixel_format in (BGRX8888, XBGR8888, BGRX8888_BIGENDIAN):
+            with self.subTest(pixel_format=pixel_format):
```
sibson, line 195: generative?

tests/unit/test_pixelformat.py
sibson: 🤖 Claude Code: The format loop was already generative here; the primaries were not. It ran only pure red, green and blue, where every bit of a channel is set and a wrong shift can still look right.

122fd8f generates the colour values too — black, white, and max/mid/1 per channel. Each decoded channel is bounded against `value * 255 / maximum` within 1, rather than compared to a reproduction of Pillow's rounding, which would pass by construction whatever Pillow did.

tests/unit/test_pixelformat.py
sibson: 🤖 Claude Code: Done in 122fd8f. It now walks every generated layout, skips the ones that are not CPIXEL-eligible, and asserts the bytes *outside* the reported window are pad — a stronger property than the three hand-picked formats, and it is exactly what the big-endian bug broke: a big-endian pixel is [pad, R, G, B], so slicing at offset 0 keeps the pad and drops blue.

It also counts the eligible formats and asserts the count is non-zero, so a `cpixel_bytes` that never returns 3 cannot pass by skipping everything.

PR #399, opened 2026-08-21
Phase 1: resolve pixel formats through pixelformat.raw_mode, add --pixel-format

sibson:
Phase 1 of `specs/decoder-architecture.md`, and the first of a stack of four — Phase 2 (pump, Raw, CopyRect) branches off this one.

`client.PF2IM` tabulated five layouts, so a server announcing anything else was asked to switch to RGB32 even when Pillow could read what it had already sent. `pixelformat.raw_mode` (landed in #397, unused until now) computes the mode instead, and the RGB32 / BGR16 fallback now fires only for a format that genuinely has no raw mode. `vncdo --pixel-format` makes a second format reachable, which is what Phase 1 exists for.

What a reviewer would not guess from the diff:

- The requested format is stamped onto each protocol by `VNCDoToolFactory.buildProtocol` rather than read back off `self.factory`. Every unit test builds a client with a `Mock` factory, so a new factory attribute read at connect time returns a `Mock` and sends it to `struct.pack` — 17 tests died that way before the hop moved. This way a hand-built client has a real default.
- `make goldens` now captures the catalogue twice, and `test_goldens` gains the cross-format check R3 asks for: fixtures differing only in the format they requested must decode to the same framebuffer, within the coarser format's quantization, which it reads off each fixture's own ServerInit rather than its `conditions.json`.

Tested: `make test` (234 unit tests) and `flake8` clean. **The cross-format check pairs nothing today** — capturing the rgb565 fixture needs the fleet, and Docker is unreachable where this was built, so it asserts only once someone runs `make goldens`. That leaves R3's two-format proof outstanding; everything else in Phase 1 is here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UmXMPeqmAPpKexubNj2dny)_

### DIFF
--- a/tests/functional/test_pixel_format.py
+++ b/tests/functional/test_pixel_format.py
@@ -1,0 +1,69 @@
+"""R3, live: the framebuffer does not depend on the negotiated pixel format.
+
+The offline half of this lives in tests/unit/test_goldens.py and compares
+committed captures. This half needs no fixture: it drives one server at
+each format in turn and compares what comes back.
+"""
+from __future__ import annotations
+
+import tempfile
+from pathlib import Path
+from unittest import TestCase
+
+from PIL import Image
+
+from tests.goldens import scenes
+from vncdotool import pixelformat
+
+from .vncservers import HOST, TIGERVNC, port_open, run_vncdo
+
+SCENES_DIR = Path(__file__).resolve().parents[1] / "goldens" / "scenes"
+
+# 32bpp throughout, so the only difference is the order of the channels and
+# a mismatched raw mode shows up as swapped colour rather than as noise a
+# tolerance could hide. A reduced format cannot be compared this way: the
+# scene player's key patch does not survive quantization, so the driver
+# cannot tell which scene it is looking at.
+FORMATS = ("bgrx8888", "rgbx8888")
+
+
+class TestPixelFormatIndependence(TestCase):
+    def setUp(self) -> None:
+        if not port_open(HOST, TIGERVNC.port):
+            self.fail(f"{TIGERVNC.name} is not listening on {TIGERVNC.port}; {TIGERVNC.how_to_start}")
+
+    def _capture(self, pixel_format: str, key: str) -> Image.Image:
+        with tempfile.TemporaryDirectory() as tmp:
+            path = Path(tmp) / "screen.png"
+            result = run_vncdo(
+                TIGERVNC, "--pixel-format", pixel_format,
+                "key", key, "pause", "0.3", "capture", str(path),
+            )
+            if result.returncode != 0:
+                self.fail(f"vncdo --pixel-format {pixel_format} failed ({result.returncode}): {result.stderr}")
+            return Image.open(path).convert("RGB").copy()
+
+    def test_every_format_decodes_the_scene_the_same_way(self) -> None:
+        for key in ("0", "s", "d"):
+            oracle = Image.open(SCENES_DIR / f"{key}.png").convert("RGB")
+            for pixel_format in FORMATS:
+                with self.subTest(scene=key, pixel_format=pixel_format):
+                    screen = self._capture(pixel_format, key)
+                    self.assertEqual(screen.size, oracle.size)
+                    self.assertEqual(
+                        screen.tobytes(), oracle.tobytes(),
+                        f"scene {key} at {pixel_format} does not match the image the server was shown",
+                    )
+
+    def test_the_server_is_asked_for_every_format_we_offer(self) -> None:
+        """A format the server refuses would otherwise pass as a silent no-op,
+        since it keeps sending its own and the client keeps decoding it.
+        """
+        for name in sorted(pixelformat.PIXEL_FORMATS):
+            with self.subTest(pixel_format=name):
+                result = run_vncdo(TIGERVNC, "-v", "--pixel-format", name, "pause", "0")
+                self.assertEqual(result.returncode, 0, result.stderr)
+                self.assertIn("Requesting", result.stderr)
+
+    def test_the_scene_key_patch_survives_a_reordered_format(self) -> None:
+        self.assertEqual(scenes.read_patch(self._capture("rgbx8888", "g")), "g")
--- a/tests/functional/test_pixel_format.py
+++ b/tests/functional/test_pixel_format.py
@@ -1,18 +1,19 @@
-"""R3, live: the framebuffer does not depend on the negotiated pixel format.
+"""The framebuffer a client sees does not depend on the pixel format it
+negotiated, checked against a live server.
 
-The offline half of this lives in tests/unit/test_goldens.py and compares
-committed captures. This half needs no fixture: it drives one server at
-each format in turn and compares what comes back.
+The offline half is in tests/unit/test_goldens.py and compares committed
+captures. This half needs no fixture: it drives one server at each format
+in turn and compares what comes back.
 """
 from __future__ import annotations
 
 import tempfile
+import unittest
 from pathlib import Path
 from unittest import TestCase
 
 from PIL import Image
 
-from tests.goldens import scenes
 from vncdotool import pixelformat
 
 from .vncservers import HOST, TIGERVNC, port_open, run_vncdo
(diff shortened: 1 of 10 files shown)


tests/goldens/capture.py
```diff
@@ -50,8 +51,20 @@ def _start_vnclog(archive: Path) -> subprocess.Popen:
 
 def main() -> int:
     parser = argparse.ArgumentParser(description=__doc__)
-    parser.add_argument("--name", default="tigervnc-raw-bgrx8888", help="fixture directory name")
+    parser.add_argument(
+        "--pixel-format",
+        choices=sorted(pixelformat.PIXEL_FORMATS),
+        default="bgrx8888",
```
sibson: default should be None to allow server native?

tests/unit/test_goldens.py
```diff
@@ -50,10 +50,62 @@ def first_difference(actual: Image.Image, expected: Image.Image, tolerance: int)
...
+        groups: dict[str, List[Path]] = {}
+        for fixture in fixtures():
+            groups.setdefault(fixture.name.rsplit("-", 1)[0], []).append(fixture)
+
+        for group, members in sorted(groups.items()):
+            if len(members) < 2:
+                continue
+            reference, *others = members
+            expected = dict(replay(reference))
+            for other in others:
```
sibson: generative?

vncdotool/client.py
```diff
@@ -62,25 +62,19 @@ class ProtocolError(VNCDoException):
...
 RGB24 = rfb.PixelFormat(24, 24, False, True, 255, 255, 255, 0, 8, 16)
-BGR16 = rfb.PixelFormat(16, 16, False, True, 31, 63, 31, 11, 5, 0)
-PF2IM = {
-    RGB24: "RGB",
-    RGB32: "RGBX",
-    BGR16: "BGR;16",
-    rfb.PixelFormat(24, 24, False, True, 255, 255, 255, 16, 8, 0): "BGR",
-    rfb.PixelFormat(32, 24, False, True, 255, 255, 255, 16, 8, 0): "BGRX",
-}
+BGR16 = pixelformat.PIXEL_FORMATS["rgb565"]
```
sibson: this smells wrong

sibson:
Reviewed this three ways — protocol correctness against RFC 6143/rfbproto, whether the tests can actually fail, and CLAUDE.md conventions. Four findings, all fixed in the commits since the PR opened.

**The cross-format check could not have worked.** A fixture holds only the server's half of the session, so the `SetPixelFormat` the capturing client sent isn't in it to replay: any capture taken at a non-native format would have been re-read at the server's announced pixel width and desynced at its first rectangle. `quantization()` had the same root cause and always returned the native format's tolerance. Each capture now records the format it asked for, and the replay negotiates it.

**And it could have passed while asserting nothing** — three ways, now all failures: a fixture with no steps (exactly what a desynced capture distills to), a pair covering different scenes, and a pair captured at the same format.

**The second capture is now rgbx8888, not rgb565.** A reduced format can't be captured at all today: the scene player's key patch is matched byte-for-byte and `expect` demands an exact histogram, and separately the scene keys sit close enough in value that 5-bit red can't distinguish `c` from `d`. Making them quantization-tolerant means re-encoding the key patch, which invalidates the committed fixture and needs a re-capture against the fleet. Reordered channels prove format independence without any of it. `tests/functional/test_pixel_format.py` proves the same thing live against the fleet, where CI can run it and no fixture is involved.

**One real bug, found by the protocol pass:** `raw_mode` on the *requested* format ran outside the `try`, so an unsupported one escaped `setImageMode` through `dataReceived` — Twisted dropped the connection but the factory deferred never fired, leaving `api.connect` to wait out its timeout undiagnosed. It now fails the connection, and does so before asking the server for a format we'd be unable to read.

Two things noted and deliberately not addressed here:

- `cpixel_bytes`/`cpixel_offset` still have no production caller — `rfb.py`'s ZRLE `cpixel()` hardcodes three little-significant-first bytes and is wrong for any non-32bpp format. Unreachable from `vncdo`, and `specs/pixel-format.md` defers it to Phase 5.
- The fallback now fires strictly less often, so a server whose native format is 15/16bpp is accepted at that depth instead of being upgraded to RGB32 — quantized screenshots where there previously were none. That is what `specs/pixel-format.md` specifies, but it is a user-visible change worth knowing about. Relatedly, the code keeps the `3.889` Apple Remote Desktop branch that the spec says was dropped; harmless, but the two disagree.

---

tests/functional/test_pixel_format.py
```diff
@@ -0,0 +1,69 @@
+"""R3, live: the framebuffer does not depend on the negotiated pixel format.
```
sibson: r3 what is this nonesense

tests/functional/test_pixel_format.py
```diff
@@ -0,0 +1,69 @@
...
+            result = run_vncdo(
+                TIGERVNC, "--pixel-format", pixel_format,
+                "key", key, "pause", "0.3", "capture", str(path),
+            )
+            if result.returncode != 0:
+                self.fail(f"vncdo --pixel-format {pixel_format} failed ({result.returncode}): {result.stderr}")
+            return Image.open(path).convert("RGB").copy()
+
+    def test_every_format_decodes_the_scene_the_same_way(self) -> None:
+        for key in ("0", "s", "d"):
```
sibson: generative?

tests/functional/test_pixel_format.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def test_the_server_is_asked_for_every_format_we_offer(self) -> None:
+        """A format the server refuses would otherwise pass as a silent no-op,
+        since it keeps sending its own and the client keeps decoding it.
+        """
+        for name in sorted(pixelformat.PIXEL_FORMATS):
+            with self.subTest(pixel_format=name):
+                result = run_vncdo(TIGERVNC, "-v", "--pixel-format", name, "pause", "0")
+                self.assertEqual(result.returncode, 0, result.stderr)
+                self.assertIn("Requesting", result.stderr)
```
sibson: need stronger assert server supports?

tests/goldens/capture.py
```diff
@@ -50,8 +51,20 @@ def _start_vnclog(archive: Path) -> subprocess.Popen:
...
+        help="format the capturing client asks the server for [%(default)s]",
+    )
+    parser.add_argument(
+        "--name",
+        help="fixture directory name [tigervnc-raw-PIXEL_FORMAT]",
+    )
     args = parser.parse_args()
+    # The name carries the format because the cross-format check pairs
```
sibson: do we really need to state this

tests/unit/test_client.py
```diff
@@ -360,7 +362,44 @@ def test_setImageMode_does_not_warn(self):
...
+
+        self.client.setPixelFormat.assert_called_once_with(client.BGR16)
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", FutureWarning)
+            assert self.client.image_mode == "BGR;16"
+
+    def test_setImageMode_falls_back_when_the_server_format_cannot_be_unpacked(self):
+        self.client._version_server = (3, 8)
+        self.client.pixel_format = COLOUR_MAPPED
+        self.client.setPixelFormat = mock.Mock()  # type: ignore[assignment]
```
sibson: why ignore here?  Is this something we can do better with Mock()

tests/unit/test_goldens.py
```diff
@@ -23,7 +23,11 @@ def fixtures() -> List[Path]:
     return sorted(path for path in FIXTURE_ROOT.iterdir() if (path / "conditions.json").exists())
 
 
-def make_client() -> client.VNCDoToolClient:
+def conditions(fixture: Path) -> dict:
```
sibson: should we have a fixture object?

tests/unit/test_goldens.py
```diff
@@ -34,6 +38,13 @@ def make_client() -> client.VNCDoToolClient:
     cli.factory.pseudodesktop = False
     cli.factory.last_rect = False
     cli.factory.qemu_extended_key = False
+    if fixture is not None:
+        # A capture holds only the server's half, so the SetPixelFormat the
```
sibson: do we need to explain this?

tests/unit/test_goldens.py
```diff
@@ -50,10 +61,90 @@ def first_difference(actual: Image.Image, expected: Image.Image, tolerance: int)
...
+    if not all(maxima):
+        raise AssertionError(f"{fixture.name}: no per-channel maxima in {fmt}")
+    return max(255 // maximum for maximum in maxima)
+
+
 class TestGoldens(unittest.TestCase):
     def test_at_least_one_fixture_is_committed(self) -> None:
         self.assertTrue(fixtures(), f"no golden fixtures under {FIXTURE_ROOT}; capture one with `make goldens`")
 
+    def test_every_fixture_holds_steps(self) -> None:
```
sibson: should this not be an invariant that golden generation enforces rather thna we test?

tests/unit/test_goldens.py
```diff
@@ -50,10 +61,90 @@ def first_difference(actual: Image.Image, expected: Image.Image, tolerance: int)
...
+        """A capture that desynced distills to init and nothing else, and
+        every check over its steps would then pass by iterating nothing.
+        """
+        for fixture in fixtures():
+            with self.subTest(fixture=fixture.name):
+                self.assertTrue(sorted(fixture.glob("step-*.bin.gz")), "fixture holds no steps")
+
+    def test_paired_fixtures_were_captured_at_different_formats(self) -> None:
```
sibson: isn't this just double checking something we did, we shouldn't be creating this case.

tests/unit/test_goldens.py
```diff
@@ -50,10 +61,90 @@ def first_difference(actual: Image.Image, expected: Image.Image, tolerance: int)
...
+            with self.subTest(group=group):
+                self.assertCountEqual(formats, set(formats), f"{group}: duplicate formats in {formats}")
+
+    def test_a_scene_decodes_the_same_at_every_captured_format(self) -> None:
+        """R3: the framebuffer does not depend on the format the server negotiated.
+
+        Fixtures of one server and encoding differ only in the format their
+        capture asked for, which is the last dash-separated part of the name.
+        """
+        groups: dict[str, List[Path]] = {}
```
sibson: generative tests?

vncdotool/client.py
```diff
@@ -317,16 +311,29 @@ def image_mode(self) -> str:
 
     def setImageMode(self) -> None:
         """Check support for PixelFormats announced by server or select client supported alternative."""
+        pixel_format = self.requested_pixel_format
+        if pixel_format is None:
+            try:
+                self._image_mode = pixelformat.raw_mode(self.pixel_format)
```
sibson, line 312: is self.pixel_format guarenteed to have a value?

tests/goldens/capture.py
sibson: Agreed, and it mattered more than it looks: `bgrx8888` happens to be tigervnc's native format, so the old default sent a `SetPixelFormat` that changed nothing and the no-flag path was never exercised. Default is now `None` — no request sent, fixture named `tigervnc-raw-native`. `make goldens` still names both formats explicitly so the recorded label matches what was actually negotiated.

---

tests/goldens/capture.py
sibson: No. Gone.

---

vncdotool/client.py
sibson: It does. Two of the three contradicted what they pointed at — `RGB32` was `rgbx8888`, and `BGR16` was `rgb565` whose Pillow raw mode is `BGR;16` — so a reader had three names for one thing and nothing saying which was authoritative. All three deleted; call sites and tests use `PIXEL_FORMATS[...]`, and `RGB24` (24bpp, no registry entry) is now a constant in the test that needs it.

---

vncdotool/client.py
sibson: Yes, always. `RFBClient.__init__` sets it to `PixelFormat()` and `_handleServerInit` overwrites it from the wire before calling `vncConnectionMade`, which is the only caller of `setImageMode`. It is never `None`, so there is nothing to guard.

---

tests/functional/test_pixel_format.py
sibson: Fair — "R3" is a requirement ID from `specs/decoder-architecture.md` and means nothing in a test file. The docstring now says what it checks in plain English: the framebuffer a client sees does not depend on the pixel format it negotiated.

---

tests/functional/test_pixel_format.py
sibson: Right — grepping stderr for `Requesting` proves we asked, not that the server agreed.

Nothing on the wire acknowledges `SetPixelFormat`, so there is no handshake to assert on: the server simply starts sending in that format or it doesn't. That makes rendering the scene the only thing that can prove the request was honoured — a server that ignored it would keep sending its native format while the client reads at the width it asked for, so the stream desyncs and the picture is wrong rather than subtly off.

So the test now drives all three registered formats, captures the scene, and compares against the image the server was shown within that format's own quantization (`255 // channel max`, so 8 for rgb565's 5-bit red and blue). That also gets `rgb565` under live test for the first time — the quantization problem only blocks the *capture* pipeline, which has to read the scene key back out of the framebuffer; pressing the key ourselves and comparing to the oracle doesn't.

---

tests/unit/test_goldens.py
sibson: Yes — moved into `capture.py`, which now refuses to write a fixture with no steps at all. A stepless fixture is what a desynced capture distills to, so the capture is where it should die, not two layers downstream in a test that watches for a state we should never create.

---

tests/unit/test_goldens.py
sibson: Agreed, same treatment: `capture.py` now refuses to write a fixture whose group already holds one captured at the same format, so the case can't be created rather than being detected afterwards.

One limit worth knowing: the guard compares recorded labels, so a `native` capture and an explicit one at the same format would both be allowed and would be the same format in fact. `make goldens` names both formats explicitly for that reason.

---

tests/unit/test_goldens.py
sibson: Done — I read this as "generate the test methods", matching the `load_tests` hook already building one case per fixture.

The cross-format check was a `subTest` loop over pairs; it is now one generated `TestCase` per pair, so a failure's id reads `TestCrossFormat_tigervnc_raw_bgrx8888_vs_tigervnc_raw_rgbx8888` instead of a subTest label buried in the output. Same treatment in `tests/functional/test_pixel_format.py`: one case per format.

Reading this closely also turned up a real bug of mine — `GoldenReplay` built its client with `make_client()` rather than `make_client(fixture)`, so the per-fixture oracle replay decoded at the format the server *announced* instead of the one the capture negotiated. Correct for the one committed fixture, wrong for any capture at another format, which is exactly the thing this PR exists to make possible.

---

tests/unit/test_client.py
sibson: The ignore was there because mypy doesn't accept assigning to a bound method. `mock.patch.object` is the better tool and needs none — it also restores the method afterwards, which plain assignment never did.

It can't go in `setUp` though: `test_setImageMode_keeps_bypp_in_step_with_the_negotiated_format` deliberately runs the *real* `setPixelFormat`, since `bypp` is a read-through of the `pixel_format` it assigns and a mock would hide any drift. So it's an opt-in `self.patch_setPixelFormat()` helper — one line in the tests that want it, nothing in the one that doesn't, and six `type: ignore` comments gone.

---

tests/unit/test_goldens.py
sibson: No — deleted. The code reads the recorded format and hands it to the client it is building; the reason it has to is now in the commit message where it belongs.

---

tests/functional/test_pixel_format.py
sibson: Done — one generated `TestCase` per format rather than a loop over a hardcoded tuple, so a failing id names the format.

---

tests/unit/test_goldens.py
sibson: Yes — there is one now. `conditions()`, `replay()`, `quantization()` and `make_client()` each took a path and re-read the same directory, and every caller had to remember that replaying a fixture means negotiating the format its conditions record. A `Fixture` knows its own group, tolerance and steps, and hands back a client already fed the init bytes; the two standalone helpers that duplicated the methods are gone.

---

PR #402, opened 2026-08-21
Phase 2: decode Raw and CopyRect through a generator pump

sibson:
Phase 2 of `specs/decoder-architecture.md`, now based on main since #399 merged. #405 and #406 sit above it.

A decoder becomes a generator that yields the byte count it wants and receives those bytes back. Raw and CopyRect move to `vncdotool/decoders/` behind a registry; the rest stay on the continuation-passing path until their phases. `updateRectangle` gains the `PixelFormat` its bytes are in — the contract change #385 warned about — and `fillRectangle` is no longer called for the migrated encodings.

**N1 is not met and needs a decision.** `make bench` replays a committed Raw fixture: 3.8ms on main, 4.7ms here, median of 60 runs. Two rounds of profiling; the two obvious hypotheses were wrong and were reverted. What is left is per-rectangle cost the design implies, and Raw is its worst case — already one paste per rectangle, so the single-paste win that pays for the pump elsewhere buys it nothing. Accept ~20% on Raw, or spend a fast path on it; I did not want to invent API to beat a benchmark without you.

The abort fix is the other thing worth reading closely: a failed decode used to leave the parked handler armed, so the parser re-entered a terminated generator and painted from an untouched buffer after the session was declared bad.

Tested: 271 unit tests, `flake8` clean, fleet and OS-server jobs green.

Issue #385, opened 2026-08-19
Subclass compatibility survey: RFBClient.fillRectangle / updateRectangle / image_mode


sibson:
🤖 Claude Code: Part of the pluggable-decoders refactor (see `specs/decoder-architecture.md`, Phase -1).

`vncdotool/rfb.py` is being split so each encoding decodes into a canonical
pixel format independently of the client, instead of the current
continuation-passing state machine spread across `RFBClient`. Two callback
contracts change once that lands:

- `updateRectangle` will receive canonical-format bytes instead of
  server-format ones, and one call per rectangle instead of one per tile.
- `fillRectangle` stops being called — decoders will fill a rect buffer
  directly instead.

`self.image_mode` also stops reflecting anything you can rely on by reading
it directly; it becomes a property.

`api.connect(..., factory_class=...)` and `RFBFactory.protocol` are a
supported extension point, so subclasses of `RFBClient` (or
`VNCDoToolClient`) overriding either hook, or reading `image_mode`, may
exist — we have no evidence either way. If that's you: **comment here**.
We'll design compatibility around real usage rather than break it silently.

Until then, subclassing either hook or `VNCDoToolClient.setImageMode` /
reading `image_mode` will start emitting a `FutureWarning` pointing at this
issue, with no behavior change yet.

sibson:
🤖 Claude Code:

The `fillRectangle`/`updateRectangle` contracts have now actually changed on `main` for 2.0, so the `FutureWarning` a subclass gets has been reworded from "will change in a future release" to naming the completed 2.0 change (#493). This issue stays open through the 2.0 release: if the change broke a subclass of yours, that is exactly what we want to hear about here, and there is still time to design compatibility around a real case rather than a hypothetical one.

PR #405, opened 2026-08-21
Phase 3: --encodings, and RRE and CoRRE on the registry

sibson:
Phase 3 of `specs/decoder-architecture.md`, stacked on #402 — this diff is the last two commits.

`--encodings` is the point of the phase: the client offered `[RAW]` plus the pseudo-encodings whatever the server supported, so every other encoding was compiled and never run. RRE and CoRRE become one generator each, CoRRE inheriting RRE's with a narrower coordinate struct, which is the entire difference between them.

**Making them selectable made two dead bugs live, so they are fixed here rather than carried across:** CoRRE raised on a format string missing its `f` prefix and stopped after the first subrectangle, and `copyRectangle` was a docstring and nothing else. `tests/unit/test_decoder_bugs.py` is gone — its `@expectedFailure` reproductions pass now and move into `test_decoders.py`.

Two choices a reviewer should check: subrectangle coordinates are rectangle-local, so there is a test with the rectangle off the origin, where a screen-coordinate decoder would still pass; and a subrectangle count larger than the rectangle has pixels is refused, which is our bound, not the spec's.

Tested: 283 unit tests, `flake8` clean. Decoder bytes come from RFC 6143 §7.7.3 and rfbproto, not from our decoder. `tests/functional/test_encodings.py` renders each encoding against the fleet and separately asserts tigervnc really emits RRE rather than answering with Raw.

Not done: the encoding probe filling in the UltraVNC and Screen Sharing columns of the support table — needs the OS fleet, worth its own change.


sibson:
Reviewed for protocol correctness against rfbproto, and separately for tests and conventions. The RRE and CoRRE wire handling came back clean — field order, coordinate widths, the unsigned count, and CoRRE-as-RRE-with-narrower-coordinates all check out against the spec text, and the subrectangle-count guard cannot reject a conforming stream. What the review did find was in `copyRectangle`, which this PR implements for the first time. All fixed in the commit above.

- **A copy whose source ran off the framebuffer pasted black.** Pillow zero-fills a crop that leaves the image, so the framebuffer was quietly corrupted and a screenshot would have lied about it — while an RRE subrectangle one pixel past its rectangle ends the session. Inconsistent, and the silent half was the wrong half. The source is now bounded in the decoder.
- **A copy landing past the announced size was clipped.** `updateRectangle` grows the canvas for servers that paint outside what they announced (the VMware ESXi case its comment names); a copy out there is the same case and now grows the same way.
- **`drawCursor` was never called**, so under `--localcursor` a copy over the pointer erased it until something else repainted, and a copy whose source held the composited pointer stamped a permanent ghost at the destination.

Also corrected: the `[BREAKING]` changelog entry said `fillRectangle` is no longer called for Raw and CopyRect. RRE and CoRRE composite into a rect buffer too, so a subclass overriding only `fillRectangle` — exactly the case #385 warned about — now renders nothing for those either. And a comment claimed encodings still on `rfb.py`'s path *cannot* be offered; they decode fine, they are simply out of scope until their phase.

Two notes rather than changes. `SUPPORTED_ENCODINGS` turns out to have no reader anywhere in the tree — deriving it from the registry is therefore inert today, correct for the phase but not load-bearing. And the strictness this inherits from #402 becomes reachable here: a server emitting a malformed RRE subrectangle now loses its connection where it previously rendered slightly wrong. That is the R6 trade the spec asks for, but it is worth being a deliberate choice rather than an inherited one.

---

vncdotool/client.py
```diff
@@ -424,6 +425,25 @@ def updateRectangle(
 
         self.drawCursor()
 
+    def copyRectangle(
+        self, srcx: int, srcy: int, x: int, y: int, width: int, height: int
+    ) -> None:
+        if self.screen is None:
+            return
+        region = self.screen.crop((srcx, srcy, srcx + width, srcy + height))
+        # Pillow pastes what fits and drops the rest, so a copy landing
```
sibson: I think Pillow behavoiur is correct.  Why do we need to grow?

vncdotool/decoders/rre.py
```diff
@@ -0,0 +1,50 @@
...
+from __future__ import annotations
+
+from struct import Struct, unpack
+from typing import TYPE_CHECKING, ClassVar, Iterator
+
+from ..const import Encoding
+from .base import DecodeError, PixelDecoder
+
+# rfb.py imports this package, so importing from it at runtime is a cycle.
+if TYPE_CHECKING:  # pragma: no cover
```
sibson: Keep an eye out for a rebase that will solve this

vncdotool/decoders/rre.py
```diff
@@ -0,0 +1,50 @@
...
+        # The count is a U32 and the specs do not bound it, but a
+        # subrectangle covers at least one pixel and this encoding exists to
+        # be smaller than Raw, so more of them than the rectangle has pixels
+        # is malformed rather than merely wasteful.
+        if count > target.width * target.height:
+            raise DecodeError(
+                f"{count} subrectangles in a {target.width}x{target.height} rectangle"
+            )
+
+        size = bypp + self.COORDS.size
```
sibson: would bytes be a better name than size?

tests/functional/test_encodings.py
```diff
@@ -0,0 +1,67 @@
...
+
+from PIL import Image
+
+from vncdotool import decoders
+
+from .vncservers import HOST, TIGERVNC, port_open, run_vncdo
+
+SCENES_DIR = Path(__file__).resolve().parents[1] / "goldens" / "scenes"
+
+# tigervnc answers a request for CoRRE with Raw (see the fleet encoding
```
sibson: run a spike to verify this, link to PR that verifies rather than hold unsubstanciated claim

tests/functional/test_encodings.py
```diff
@@ -0,0 +1,67 @@
...
+            result = run_vncdo(
+                TIGERVNC, "-v", "--encodings", encodings,
+                "key", key, "pause", "0.3", "capture", str(path),
+            )
+            if result.returncode != 0:
+                self.fail(f"vncdo --encodings {encodings} failed ({result.returncode}): {result.stderr}")
+            return Image.open(path).convert("RGB").copy(), result.stderr
+
+    def test_every_encoding_renders_the_scene_it_was_shown(self) -> None:
+        for name in sorted(decoders.ENCODING_NAMES):
```
sibson: generative?

tests/functional/test_encodings.py
```diff
@@ -0,0 +1,93 @@
...
+    """One encoding against the images the server was shown.
+
+    Not a TestCase itself, so the loader collects it only through the
+    subclasses load_tests builds.
+    """
+
+    encoding: str
+
+    def test_renders_every_scene(self) -> None:
+        for key in SCENES:
```
sibson: generative?  Is a test fails can we post the screens as pr comments?

vncdotool/client.py
sibson: 🤖 Claude Code: Agreed, dropped in 7113625 — `copyRectangle` is crop/paste/drawCursor again. It was strictly dead code: the commit that added the grow only tested the source-bounds refusal, and no server was ever observed copying past the announced size (unlike `updateRectangle`, whose growth has the VMware ESXi case behind it).

vncdotool/decoders/rre.py
sibson: 🤖 Claude Code: That rebase landed. #415 moved `PixelFormat` below `rfb.py` and split `DecodeError` into `decoders/errors.py`, so the `TYPE_CHECKING` guard and the quoted annotations are gone — `rre.py` now imports `PixelFormat`, `RectBuffer` and `DecodeError` for real.

vncdotool/decoders/rre.py
sibson: 🤖 Claude Code: Renamed to `nbytes` in 6add191. Not `bytes` — it would shadow the builtin in a module that unpacks byte strings a few lines below.

tests/functional/test_encodings.py
sibson: 🤖 Claude Code: Spike ran, claim held: #417 probed the fleet tigervnc 1.12.0 at the socket — a CoRRE-only offer came back Raw every time, an RRE-only offer came back RRE, and x11vnc answered the same CoRRE probe with CoRRE. Upstream `EncodeManager::supported()` accepts Raw, RRE, Hextile, ZRLE and Tight only. The comment now cites that measurement (71a0f58); #417 carries the spec-table half.

tests/functional/test_encodings.py
sibson: 🤖 Claude Code: Done in 6add191 — `load_tests` builds a case per encoding, following `tests/unit/test_goldens.py` and `tests/functional/test_pixel_format.py`. 95c35d2 then extended it to the cross product with scenes, so ids read `TestRenders_rre_scene_s`. The convention is in CLAUDE.md.

tests/functional/test_encodings.py
sibson: 🤖 Claude Code: Generative half done in 95c35d2 — one case per encoding and scene, so a failure names both (`TestRenders_corre_scene_0`), and CLAUDE.md now says to generate the cross product rather than loop the inner set.

Posting the screens on failure is a separate piece of work and is queued, not done: captured / oracle / diff images published only when a comparison fails, following the workflow #416 added for decode counts. Leaving this thread open until that lands.

PR #406, opened 2026-08-21
Phase 4: Hextile

sibson:
Phase 4 of `specs/decoder-architecture.md`, the last of the stack, on #405 — this diff is the last two commits.

Six mutually recursive `_handleDecodeHextile*` methods threading eight parameters through every hop become a generator with two `for` loops: `rfb.py` loses 193 lines, the decoder is 84. First phase with a user-visible result — `--encodings hextile` now works against the near-universal encoding.

Two things the migration fixes rather than carries across: a subrectangle is checked against its own tile, not just the rectangle, so packed-coordinate mistakes can no longer hide inside the rect buffer; and a first tile that specifies neither background nor foreground is a diagnosed protocol error instead of an exception on an empty colour.

**N2 is half measured.** `tests/functional/test_bandwidth.py` measures bytes on the wire in CI and asserts Hextile beats Raw on the same scene. Render time is not measured: a shared runner cannot time a decode against a 256x192 scene usefully, and I have no fleet locally to capture a Hextile fixture for `make bench`. That capture is the missing step. The ~20% Raw regression from #402 is unchanged here and unmeasured by this phase.

Tested: 295 unit tests, `flake8` clean. Bytes are derived from RFC 6143 §7.7.4, not from the implementation this deletes — packed subrectangle coordinates, colour carry-over across tiles, and a rectangle whose size is not a multiple of 16.


sibson:
Reviewed for protocol correctness against rfbproto and separately for tests and conventions. The correctness pass fuzzed the decoder against an independently written reference encoder — 4,000 valid streams across ten rectangle sizes compared pixel-for-pixel and byte-for-byte, plus 200,000 garbage streams. Tile geometry, the subencoding mask, the packed subrectangle coordinates and the byte accounting all came back clean, and no malformed stream produced anything but a `DecodeError`. It found two real defects, both fixed above.

**Colours must not carry across a raw tile.** rfbproto says a background may not be carried over if the previous tile was raw, and a foreground may not be carried over if the previous tile was raw *or* had `SubrectsColored` set. The decoder carried both — and the test I committed alongside it asserted that as the correct behaviour, so the test was wrong in the same direction as the code. Both corrected; clearing the colours also means the "no colour has been established" check now covers the one path that was silently exempt from it.

**A tile declaring zero subrectangles was refused.** `AnySubrects` with a count of `0` is redundant but well-formed, and it is what a naive encoder emits when it always sets the bit. The foreground check ran before the count was read, so such a tile killed the session — and whether it did depended on whether some unrelated earlier tile happened to set a foreground, while the same tile with `SubrectsColored` also set decoded fine. It was also a regression: the deleted implementation gated on `if subrects:` and passed these through.

Separately, the tests-and-conventions pass caught why the fleet job failed on the first push, which was my bug in the new bandwidth test: `key` and `pause` never issue a `FramebufferUpdateRequest`, and a compliant server sends pixel data only when asked (RFC 6143 §7.5.3). The capture held the handshake and nothing else — identical whatever encoding was negotiated — so the comparison was between two constants. It captures now, and fails loudly if a capture comes back smaller than the scene has pixels rather than comparing two handshakes.

Two test gaps closed while there: a 2×2 grid of tiles cannot distinguish a row-major walk from a column-major one, and nothing covered a raw tile sitting between a tile that sets colours and one that inherits them — which is how the carry-over defect above became visible.

---

tests/unit/test_decoders.py
```diff
@@ -345,6 +372,308 @@ def test_corre_subrect_count_larger_than_rectangle_pixels_is_a_protocol_error(se
...
+        # No BACKGROUND_SPECIFIED bit anywhere, and no earlier tile to
+        # inherit one from.
+        tile = hextile_tile(HextileEncoding(0))
+        self.cli.dataReceived(
+            framebuffer_update([hextile_rect(0, 0, width, height, [tile])])
+        )
+
+        self.cli.vncProtocolError.assert_called_once()
+        self.cli.transport.loseConnection.assert_called_once()
+
```
sibson, line 676: file is getting long each decoder has it's own test file?

sibson: 🤖 Claude Code: Agreed — 679 lines, 25 tests in one class, and ZRLE/Tight/Cursor each add a comparable block. Split spun off as its own change rather than folded in here, since it moves the RRE and CoRRE tests that landed with #405 as well as the Hextile ones: one file per decoder, fixture builders shared, every test moved unchanged, plus the CLAUDE.md convention so the next encoding lands in its own file.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #407, opened 2026-08-21
Narrow Twisted's transport/factory/peer to what runs after connect

sibson:
Twisted types `transport`, `factory` and `peer` wide because they are unset before a connection exists. Every protocol class here touches them only after `connectionMade`, so declaring the narrow type on the subclass states an invariant the code already relies on rather than suppressing anything. Six bare annotations; no ignores, no casts, no logic changes.

**141 → 89 errors for `vncdotool` alone** (303 → 230 across `vncdotool`+tests). Bare annotations add an `__annotations__` entry and nothing else — verified Twisted's own `transport = None` / `peer = None` class attributes are untouched at runtime.

Three narrowings were tried and reverted because they surface real problems rather than resolving noise, and each deserves its own change:

- `RFBClient.factory` — `RFBFactory` never declares `username`, but `RFBClient`'s ARD auth reads `self.factory.username`. A genuine gap in the base class.
- `VNCDoToolClient.factory` — surfaces `connectionLost`/`vncAuthFailed` passing `self` where `clientConnectionLost`/`clientConnectionFailed` declare an `IConnector`. Pre-existing signature mismatch.
- `RFBServer.transport` — no honest single type. It is only mixed into `VNCLoggingServerProxy(ProxyServer, RFBServer)`, and Twisted narrows `ProxyServer.transport` to `_MakeTypesHappy`, which lacks `setTcpNoDelay`; neither it nor `ITCPTransport` is a subtype of the other, so any narrowing conflicts in the MRO.

`setTcpNoDelay` in `client.py` needed nothing — the existing `isinstance(self.transport, ITCPTransport)` guard already narrows per call, and `ITransport` is the honest declared type since `factory_connect` also serves UNIX sockets.

Tested: `mypy vncdotool` 89, `make typecheck` (currently `vncdotool tests`) 230, `make test` 302 pass, `--warn-unused-ignores` still empty, `flake8` clean.

### DIFF
--- a/vncdotool/command.py
+++ b/vncdotool/command.py
@@ -91,6 +91,8 @@ def log_connected(pcol: TClient) -> TClient:
 
 
 class VNCDoCLIClient(VNCDoToolClient):
+    factory: VNCDoCLIFactory
+
     def vncRequestPassword(self) -> None:
         if self.factory.password is None:
             self.factory.password = getpass.getpass("VNC password:")
--- a/vncdotool/loggingproxy.py
+++ b/vncdotool/loggingproxy.py
@@ -220,6 +220,7 @@ class VNCLoggingClientProxy(portforward.ProxyClient):
 
     vnclog: VNCLoggingClient | None = None
     ncaptures = 0
+    peer: VNCLoggingServerProxy
 
     def startLogging(self, peer: VNCLoggingServerProxy) -> None:
         self.vnclog = VNCLoggingClient()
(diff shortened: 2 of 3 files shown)


sibson:
🤖 Claude Code: Rebased onto current main (#403 merged, dropped). Numbers refreshed: vncdotool 141→89, vncdotool+tests 303→230, 302 tests pass. Mergeable.

PR #408, opened 2026-08-21
Add raw QEMU/KVM as a third Tier 2 server

sibson:
## Summary

Adds a third Tier 2 OS-hosted server alongside UltraVNC (Windows) and Screen Sharing (macOS): a raw `qemu-system-x86_64` VNC server started directly on a Linux runner, no container, no guest OS. `specs/server-compatibility-plan.md` records why it's raw rather than Tier 1's Docker-based `qemu` service (a real-world console shape, and hosted `ubuntu-latest` has had `/dev/kvm` since 2023).

- `tests/servers/qemu-kvm/{setup.sh,collect-diagnostics.sh,README.md}` — no `-drive`/`-kernel`, so the firmware's own "no bootable device" framebuffer is what's served; `setup.sh` opens up `/dev/kvm` permissions on hosted runners and fails loudly rather than silently falling back to TCG.
- `tests/functional/vncservers.py` — registers `QEMU_KVM` under `OS_SERVERS_BY_PLATFORM["linux"]`, picked up by the existing `test_os_servers.py` machinery.

## Test plan

- [x] `flake8` clean
- [x] Local proof under TCG (this dev machine has no `/dev/kvm`): `setup.sh`, then the shared connect/keypress/mousemove/capture suite run directly against `QEMU_KVM`, then `collect-diagnostics.sh` — all clean. Captured a real 720x400 firmware text screen, not black.
- [ ] Not yet proven: the actual KVM-accelerated path, and CI wiring into `os-servers.yml`. Deliberately left out of this PR pending a manual `workflow_dispatch` proof run on `ubuntu-latest` — follow-up.

### DIFF
--- a/.github/actions/os-server/action.yml
+++ b/.github/actions/os-server/action.yml
@@ -1,11 +1,12 @@
 name: Set up an OS-hosted VNC server
 description: >
   Turn the runner into a live VNC server for tests/functional/test_server_compat_native.py:
-  UltraVNC as a Windows service, or Apple Screen Sharing on macOS.
+  UltraVNC as a Windows service, Apple Screen Sharing on macOS, or a raw
+  QEMU on Linux.
 
 inputs:
   os:
-    description: macos or windows.
+    description: macos, windows, or linux.
     required: true
   port:
     description: Port the server serves on.
@@ -47,3 +48,11 @@ runs:
         PORT: ${{ inputs.port }}
         WAIT_SECONDS: ${{ inputs.wait-seconds }}
       run: pwsh -File "$env:GITHUB_ACTION_PATH/windows.ps1"
+
+    - name: Start raw QEMU
+      if: inputs.os == 'linux'
+      shell: bash
+      env:
+        VNCDOTOOL_OS_SERVER_PORT: ${{ inputs.port }}
+        VNCDOTOOL_OS_SERVER_WAIT: ${{ inputs.wait-seconds }}
+      run: bash tests/servers/qemu-kvm/setup.sh
--- a/.github/workflows/os-servers.yml
+++ b/.github/workflows/os-servers.yml
@@ -1,14 +1,14 @@
 name: OS-hosted VNC servers
 
 # Live, OS-bound VNC servers (UltraVNC on Windows, Apple Screen Sharing on
-# macOS) that only exist on GitHub-hosted windows-latest / macos-latest VMs --
-# see specs/server-compatibility-plan.md, Tier 2. These are the only place
-# UltraVNC and Screen Sharing compatibility can be exercised live, so this
-# job runs on real changes rather than never, but it is deliberately
-# change-triggered (path-filtered, no schedule).
+# macOS, raw QEMU/KVM on Linux) -- see specs/server-compatibility-plan.md,
+# Tier 2. These are the only place UltraVNC and Screen Sharing compatibility
+# can be exercised live, so this job runs on real changes rather than never,
+# but it is deliberately change-triggered (path-filtered, no schedule).
 #
 # What each server needs, and what it does and doesn't prove, is documented
-# in tests/servers/ultravnc/README.md and tests/servers/screen-sharing/README.md.
+# in tests/servers/ultravnc/README.md, tests/servers/screen-sharing/README.md,
+# and tests/servers/qemu-kvm/README.md.
 
 on:
   push:
@@ -19,6 +19,7 @@ on:
       - 'tests/functional/**'
       - 'tests/servers/ultravnc/**'
       - 'tests/servers/screen-sharing/**'
+      - 'tests/servers/qemu-kvm/**'
       - '.github/workflows/os-servers.yml'
       - '.github/actions/os-server/**'
   pull_request:
@@ -27,6 +28,7 @@ on:
       - 'tests/functional/**'
       - 'tests/servers/ultravnc/**'
       - 'tests/servers/screen-sharing/**'
+      - 'tests/servers/qemu-kvm/**'
       - '.github/workflows/os-servers.yml'
       - '.github/actions/os-server/**'
   workflow_dispatch:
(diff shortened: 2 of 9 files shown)


tests/functional/vncservers.py
```diff
@@ -133,14 +133,36 @@ class VNCServer(NamedTuple):
     how_to_start="the OS server setup runs in CI only, see tests/servers/screen-sharing/README.md",
 )
 
+QEMU_KVM = VNCServer(
+    name="qemu-kvm",
+    port=OS_SERVER_PORT,
```
sibson: should we hav a default port?

tests/functional/vncservers.py
```diff
@@ -133,14 +133,36 @@ class VNCServer(NamedTuple):
...
+    # QEMU's -vnc has no authentication unless started with password=on.
+    # Firmware's default VGA mode, not a geometry we ask for -- see
+    # tests/servers/qemu-kvm/README.md.
+    size=None,
+    timeout=OS_SERVER_TIMEOUT,
```
sibson: default timeout?

tests/functional/vncservers.py
```diff
@@ -133,14 +133,36 @@ class VNCServer(NamedTuple):
     how_to_start="the OS server setup runs in CI only, see tests/servers/screen-sharing/README.md",
 )
 
+QEMU_KVM = VNCServer(
+    name="qemu-kvm",
+    port=OS_SERVER_PORT,
+    # QEMU's -vnc has no authentication unless started with password=on.
```
sibson: explaining what exacly?

tests/functional/vncservers.py
```diff
@@ -133,14 +133,36 @@ class VNCServer(NamedTuple):
...
+    timeout=OS_SERVER_TIMEOUT,
+    how_to_start="set the server up first with tests/servers/qemu-kvm/setup.sh",
+)
+
 OS_SERVERS_BY_PLATFORM: Dict[str, List[VNCServer]] = {
     "win32": [ULTRAVNC],
     "darwin": [SCREEN_SHARING],
 }
 
+# Windows and macOS are safe to key off sys.platform alone: the OS-server
```
sibson: overkill

tests/functional/vncservers.py
```diff
@@ -133,14 +133,36 @@ class VNCServer(NamedTuple):
...
+# isn't -- ubuntu-latest also hosts ci.yml's unrelated Tier 1 Docker-fleet
+# job, which never runs tests/servers/qemu-kvm/setup.sh, so registering
+# QEMU_KVM for every Linux process would fail that job instead of skipping
+# it. setup.sh sets this to opt in only where it actually started the server.
+QEMU_KVM_OPT_IN = "VNCDOTOOL_QEMU_KVM_OS_SERVER"
+
 
 def os_servers(platform: str = sys.platform) -> List[VNCServer]:
-    return OS_SERVERS_BY_PLATFORM.get(platform, [])
+    servers = list(OS_SERVERS_BY_PLATFORM.get(platform, []))
```
sibson: something about this feels like we've made a wrong design decision somewhere

tests/functional/vncservers.py
sibson: OS_SERVER_PORT already defaults to 5900 via env fallback (a few lines up), same as ULTRAVNC/SCREEN_SHARING. Not new to this PR.

tests/functional/vncservers.py
sibson: Same as the port question -- OS_SERVER_TIMEOUT already defaults to 60s via env fallback, shared with ULTRAVNC/SCREEN_SHARING.

tests/functional/vncservers.py
sibson: Fair -- that comment was explaining the absent password=/username= kwargs but landed right above size=None, so it read like it was explaining that instead. Removed in daaaf42; ULTRAVNC/SCREEN_SHARING don't comment their auth kwargs either, so there was nothing here that actually needed saying.

tests/functional/vncservers.py
sibson: Trimmed to two lines in daaaf42 -- the five-line version was restating PR history that belongs in the commit message, not the code.

tests/functional/vncservers.py
sibson: Dug into this before renaming anything. The actual constraint: ubuntu-latest runs both ci.yml's unscoped `discover -s tests/functional` (Tier 1, no server ever started) and would run the future Tier 2 QEMU job -- same sys.platform value, two purposes. win32/darwin don't have this problem not because platform-gating is inherently safe, but because ci.yml never runs a job on those OSes; if it ever did, ULTRAVNC would break the same way.

Considered dropping the flag for a port_open() probe instead. Rejected it: that would make a genuinely broken Tier 2 setup.sh degrade to 0 registered tests instead of 4 loud failures naming `tests/servers/qemu-kvm/setup.sh`, which is backwards from this repo's fail-loudly stance (server-compatibility-plan.md Phase 1).

So some explicit signal seems structurally necessary, not a workaround -- but the *shape* of what I shipped was off: qemu-specific naming instead of matching the existing VNCDOTOOL_OS_SERVER_* family, and a comment block that read like a PR retrospective. Fixed both in daaaf42 (VNCDOTOOL_OS_SERVER_LINUX, two-line comment). If there's a design you have in mind that avoids the signal entirely, I'm listening -- but I couldn't find one that doesn't either weaken the fail-loudly guarantee or require restructuring ci.yml's discover into multiple scoped invocations, which felt like the more disruptive change for a stable job.

sibson: Good push -- you're right, the platform check was redundant. Dropped it in be428b4: os_servers() now only checks VNCDOTOOL_OS_SERVER_LINUX, no platform== condition. Verified darwin+var-set now also returns qemu-kvm alongside screen-sharing, which can't happen in real CI (the var is only ever set by setup.sh, which is Linux/apt-only, and jobs don't share environment across OS boundaries) -- treating that as an accepted corner rather than something to guard against.

.github/actions/os-server/macos.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 # Turn this macOS machine into a live Screen Sharing server for
-# tests/functional/test_os_servers.py. See
+# tests/functional/test_os_server_compat.py. See
```
sibson: test_server_compat_native and test_server_compat_docker should the file names

PR #413, opened 2026-08-22
CI: comment on a PR when the decode call counts move

sibson:
A spike: CI replays the golden once on the branch and compares the vncdotool call
counts against the ones already committed in `bench.jsonl`, commenting on the pull
request only when they moved. No second bench run, and no timings -- a shared
runner cannot measure those well enough to post.

The baseline comes from the base branch, not the workspace, so a row a pull request
appends cannot stand in as its own baseline. `bench.jsonl` stays hand-recorded.

Tested by running the workflow's steps by hand: silent on an unmodified tree, and
with a `self.bypp` added to `updateRectangle` it reported `rfb.py:bypp 175 -> 349`
and nothing else. The job has never run on a runner.


tests/goldens/report.py
```diff
@@ -93,20 +93,66 @@ def _diff(rows: List[Row], refs: Sequence[str]) -> None:
           f"  {sum(counts_after.values()) - sum(counts_before.values()):+}")
 
 
+def _baseline(rows: List[Row], fixture: str) -> Row:
+    """Dirty rows are skipped: their counts belong to a working tree
```
sibson: rather than skip dirty we should just never record them

sibson: 🤖 Claude Code: Agreed, and done in c5d0ff5. `--record` now refuses when the tree is dirty (timings still print, only the row is withheld), so the `dirty` field is gone from the schema and from the two committed rows, and the baseline no longer filters. Verified both ways: dirty tree prints the refusal and appends nothing, clean tree records as before.

sibson:
🤖 Claude Code: Superseded by #416, which takes a different approach to the same goal. This PR compared against the counts recorded in `bench.jsonl`, which goes stale the moment a change lands — every later PR would report that change until someone recorded a new row. #416 counts the base branch during the run instead, so a landed change stops being reported once it is the base.

It also carried a real defect worth recording: `--check` counted with a cold `functools.lru_cache` on `pixelformat.raw_mode` while recorded rows are counted warm, so it would have posted a phantom +8 on every branch. #416 fixes that inside `call_counts` so no caller can get it wrong.

The dirty-tree refusal from this branch is carried over into #416.

PR #415, opened 2026-08-22
Move PixelFormat below rfb.py, drop the Rect alias

sibson:
`rfb.py` imports the decoders package, so it sits at the top of the graph — but it also defined `PixelFormat` and `Rect`, so everything below had to import back up. `pixelformat.py` imported `rfb` purely to name `rfb.PixelFormat`, and each decoder carried a `TYPE_CHECKING` block with string-literal annotations to dodge the cycle.

`PixelFormat` moves to `pixelformat.py`, which now imports nothing else from the package. `rfb.py` re-exports it, so `rfb.PixelFormat` is the same object as before; fields, `STRUCT` and behaviour are untouched, and nothing pickles or isinstance-checks it.

`Rect` is deleted rather than relocated: a bare four-tuple alias with seven annotation sites that every use site already destructures. **`rfb.Rect` is gone** — that is the one break, and it has a CHANGELOG line.

Every conditional import is gone too. `DecodeError` moves to a new `decoders/errors.py`, which is what let `base.py` import `RectBuffer` plainly: `buffer.py` raises `DecodeError`, so while that exception lived in `base.py` the two modules could not both import each other.

One thing a reviewer would not guess from the diff: `docs/modules.rst` gains a `pixelformat` section, because `automodule` documents only members a module defines and `PixelFormat` would otherwise have dropped out of the published docs.

Decoder phases 3 (#405) and 4 (#406) touch these same imports and will need a trivial rebase; neither branch was touched here.

Tested: `make test` 270 tests, OK (expected failures=2). `uv run flake8 --count --statistics vncdotool tests` clean.

### DIFF
--- a/specs/decoder-architecture.md
+++ b/specs/decoder-architecture.md
@@ -165,12 +165,6 @@ Decoders depend on nothing from `rfb.py`: the pump keeps `_rectBuffer`,
 `_pumpPixels` and `_pumpForClient` — live with the pump rather than on the
 decoder.
 
-That holds at import time too. The two types the pump and a decoder both name
-sit below both: `PixelFormat` in `pixelformat.py`, and `Rect` — the rectangle
-the pump hands to `decodeForClient` — beside the interface that receives it.
-`rfb.py` imports them and re-exports both, since `rfb.PixelFormat` and
-`rfb.Rect` are public API.
-
 There is no separate sink object. The pump calls the existing client callbacks —
 `updateRectangle`, `copyRectangle`, `updateCursor` — which is the vocabulary the
 codebase already uses.
@@ -281,8 +275,9 @@ diagnosed disconnect. This is a larger user-facing win than the split itself.
 ```
 vncdotool/pixelformat.py         PixelFormat, its Pillow raw mode, CPIXEL/TPIXEL widths
 vncdotool/decoders/__init__.py   registry, built from the decoder classes
-vncdotool/decoders/base.py       Decoder, PixelDecoder, ClientDecoder, Rect
+vncdotool/decoders/base.py       Decoder, PixelDecoder, ClientDecoder
 vncdotool/decoders/buffer.py     RectBuffer
+vncdotool/decoders/errors.py     DecodeError
 vncdotool/decoders/{raw,rre,corre,hextile,zrle,cursor}.py
 vncdotool/decoders/control.py    DesktopSize, LastRect, QEMU extended key
 vncdotool/rfb.py                 negotiation, auth, message framing, the pump
--- a/specs/decoder-architecture.md
+++ b/specs/decoder-architecture.md
@@ -165,6 +165,12 @@ Decoders depend on nothing from `rfb.py`: the pump keeps `_rectBuffer`,
 `_pumpPixels` and `_pumpForClient` — live with the pump rather than on the
 decoder.
 
+That holds at import time too. The two types the pump and a decoder both name
+sit below both: `PixelFormat` in `pixelformat.py`, and `Rect` — the rectangle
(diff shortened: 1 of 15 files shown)


specs/decoder-architecture.md
```diff
@@ -165,6 +165,12 @@ Decoders depend on nothing from `rfb.py`: the pump keeps `_rectBuffer`,
 `_pumpPixels` and `_pumpForClient` — live with the pump rather than on the
 decoder.
 
+That holds at import time too. The two types the pump and a decoder both name
```
sibson: narrates code

specs/pixel-format.md
```diff
@@ -99,11 +99,17 @@ by the goldens' cross-format comparison.
...
-def cpixel_bytes(pixel_format: rfb.PixelFormat) -> int: ...  # 3 or bypp
-def cpixel_offset(pixel_format: rfb.PixelFormat) -> int: ... # 0 or bypp - 3
+class PixelFormat: ...                                   # RFC 6143 §7.4
+
+def raw_mode(pixel_format: PixelFormat) -> str: ...      # raises UnsupportedPixelFormat
+def cpixel_bytes(pixel_format: PixelFormat) -> int: ...  # 3 or bypp
+def cpixel_offset(pixel_format: PixelFormat) -> int: ... # 0 or bypp - 3
 ```
 
+`pixelformat.py` defines `PixelFormat` and imports nothing else from the
```
sibson: not interesting

vncdotool/decoders/base.py
```diff
@@ -1,15 +1,18 @@
...
-from typing import TYPE_CHECKING, ClassVar, Iterator
+from typing import TYPE_CHECKING, ClassVar, Iterator, Tuple
 
 from ..const import Encoding
+from ..pixelformat import PixelFormat
 
-# rfb.py imports this package, so importing from it at runtime is a cycle.
+# buffer.py raises DecodeError, so it imports this module and cannot be
+# imported back from it at runtime.
 if TYPE_CHECKING:  # pragma: no cover
```
sibson: the goal was to remove this conditional

vncdotool/rfb.py
```diff
@@ -44,60 +42,15 @@
 
 from . import decoders
 from .const import Encoding, HextileEncoding, AuthTypes, MsgC2S, MsgS2C
+from .decoders import Rect
```
sibson: Do we need Rect at all?

specs/decoder-architecture.md
sibson: 🤖 Claude Code: Agreed, deleted. The module-layout table above already said where each type lives; this paragraph re-narrated the diff. Table updated for `errors.py` instead.

specs/pixel-format.md
sibson: 🤖 Claude Code: Deleted.

vncdotool/decoders/base.py
sibson: 🤖 Claude Code: Gone. The blocker was `DecodeError` living in `base.py` while `buffer.py` raises it, so the two could not both import each other. `DecodeError` moves to a new `decoders/errors.py`, and `base.py` imports `RectBuffer` plainly. No decoder module has a conditional import left.

vncdotool/rfb.py
sibson: 🤖 Claude Code: No. Deleted it — bare four-tuple alias, seven annotation sites, no behaviour, and every use site already destructures the tuple. Removing it also removes the question of which module owns it and the re-export from `rfb.py`. `rfb.Rect` is gone rather than relocated, so there is a `[BREAKING]` CHANGELOG line; say the word if you would rather keep the name.

PR #419, opened 2026-08-22
Add raw QEMU/KVM as a third Tier 2 server

sibson:
## Summary

Stacked on #418 (the compat-suite rename) -- this PR is qemu-kvm-server minus that rename, split per review. Adds a third Tier 2 OS-hosted server alongside UltraVNC (Windows) and Screen Sharing (macOS): a raw `qemu-system-x86_64` VNC server started directly on a Linux runner, no container, no guest OS. `specs/server-compatibility-plan.md` records why it's raw rather than Tier 1's Docker-based `qemu` service.

- `tests/servers/qemu-kvm/{setup.sh,collect-diagnostics.sh,README.md}` -- `setup.sh` opens up `/dev/kvm` permissions on hosted runners and fails loudly rather than silently falling back to TCG.
- `tests/functional/utils.py` -- registers `QEMU_KVM` under `OS_SERVERS_BY_PLATFORM["linux"]`, gated on `VNCDOTOOL_OS_SERVER_LINUX` (set only by `setup.sh`) since `ubuntu-latest` also runs `ci.yml`'s unrelated Docker-fleet job. Introduces an `os_server()` factory so the three native servers state only what's their own.

## Test plan

- [x] flake8 clean, unit suite unchanged (270 tests)
- [x] Local proof under TCG (this dev machine has no `/dev/kvm`): `setup.sh`, then the shared connect/keypress/mousemove/capture suite run directly against `QEMU_KVM` -- all clean.
- [ ] Not yet proven: the real KVM-accelerated path, and CI wiring into `os-servers.yml`. Deliberately left out pending a manual `workflow_dispatch` proof run on `ubuntu-latest` -- follow-up.

Diffs against `main`: this branch = #418 + the above. Once #418 merges, this PR's base moves to `main` automatically and the diff shrinks to just the qemu-kvm addition.

### DIFF
--- a/specs/server-compatibility-plan.md
+++ b/specs/server-compatibility-plan.md
@@ -219,7 +219,10 @@ Concretely:
 
 TigerVNC, TightVNC, x11vnc, QEMU, LibVNCServer examples, and later
 websockify/noVNC are all runnable on Linux, so we own them outright as a
-**Docker Compose fleet**:
+**Docker Compose fleet**. (QEMU here is a containerized guest for local-dev
+convenience — Docker Desktop has no `/dev/kvm`, so it runs TCG-only and
+boots slower; the CI-real, KVM-accelerated QEMU coverage lives in Tier 2
+below, run raw on the runner rather than in a container.)
 
 - One `tests/servers/docker-compose.yml` defines a service per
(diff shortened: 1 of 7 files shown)


sibson:
Superseded by #408 -- both ended up with identical content after rebasing onto merged main (#418); keeping #408 since Autofix and the review thread are already there.

PR #426, opened 2026-08-23
Split test_decoders.py into one file per decoder

sibson:
`tests/unit/test_decoders.py` held 25 tests in one `TestEncodings` class, and ZRLE, Tight and Cursor each add a comparable block. Split into `test_decoder_hextile.py` (12), `test_decoder_rre.py` (6 RRE + 6 CoRRE), `test_decoder_copyrect.py` (1), over shared builders in `tests/unit/decoder_fixtures.py` — not named `test*.py`, so discovery skips it.

Two placement calls a reviewer might have made differently: the `hextile_*` builders stayed in the hextile file rather than the shared module, since they have one caller; CoRRE's tests share `test_decoder_rre.py` because `CoRREDecoder` subclasses `RREDecoder` and the two sets are line-for-line parallel.

Test bodies are byte-identical moves. The enclosing class names differ, since `TestEncodings` is the wrong name in a per-decoder module. Imports are absolute so each module resolves both under `make test` and standalone.

Tested: `make test` 302 before, 302 after; flake8 clean; each new module runs on its own (12/12/1).


tests/unit/utils.py
```diff
@@ -0,0 +1,82 @@
+"""Fixture builders shared by the per-encoding decode tests.
```
sibson, line 1: utils? or decoder_test_utils?

CLAUDE.md
```diff
@@ -70,6 +70,13 @@ fix the underlying bug, move the test into the topical file, drop its
 `@unittest.expectedFailure` marker, and rename it for the behaviour it checks
 rather than the issue number.
 
+Per-encoding decode tests get one file per decoder, `test_decoder_<name>.py`,
```
sibson: I don't think we need this, it's should be obvious from existing code once this PR lands

tests/unit/utils.py
sibson: 🤖 Claude Code: Went with `utils.py`, matching `tests/functional/utils.py` — 9edb914. It also drops out of discovery on its own, so the header no longer needs the line explaining why it is not `test*.py`.

CLAUDE.md
sibson: 🤖 Claude Code: Removed — 110c6f1.

PR #427, opened 2026-08-23
Drop the cross-format golden oracle, and stop inferring a capture's format

sibson:
Removes the cross-format golden check, which cannot fail on its own: every fixture is already compared to its scene PNG at tolerance 0, so one member of a pair equals the PNG exactly and comparing the other against it is that same comparison at a looser bound. Reduced depth, where it was meant to start earning its place, does not change that.

So R3 is checked per format by each fixture against its PNG. Three specs said otherwise, and `pixel-format.md` additionally claimed this check compensates for a wide reduced-depth tolerance, which it does not — that risk is now recorded as unmitigated.

`--pixel-format` becomes required at capture, so a fixture can no longer record `null` and be replayed at whatever ServerInit announced. All six committed fixtures already name a format: they replay untouched and conditions.json is unchanged.

### DIFF
--- a/specs/decoder-goldens.md
+++ b/specs/decoder-goldens.md
@@ -167,6 +167,11 @@ opening anything.
 
 ## conditions.json
 
+`--pixel-format` is required, so `pixel_format` always names a format rather
+than recording "the server's own". Nothing on the wire would correct it — the
+server never acknowledges `SetPixelFormat` — so a fixture that does not say
+what it asked for cannot be replayed at the format its pixels are in.
+
 Written by the harness from what happened, never from what was intended: the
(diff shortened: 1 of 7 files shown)


specs/decoder-goldens.md
```diff
@@ -203,6 +203,13 @@ Cross-format self-consistency runs today. The tolerance oracle waits on a
 reduced-depth format: every format captured so far is 32bpp truecolour, which
 is why the tolerance is zero.
 
+A pair is compared at the format it negotiated, not the one its name and
```
sibson: "it requested"

PR #428, opened 2026-08-23
fix(api): disconnect() after a failed command, not just success

### LINKED ISSUE #146: Screen capture to non existing directory hangs forever
When using the CLI tool to capture a screenshot and the directory is not exists, the command hang forever instead of returning an error.

`vncdo capture non-exists/test.png`

When using the API it is throwing an exception:
```
client = api.connect('127.0.0.1', password=None)
client.captureScreen('non-exists/screenshot.png')
```
The error is:
```
  File "/Users/arossert/tests/vnc_test/vnc_test.py", line 5, in <module>
    client.captureScreen('1/screenshot.png')
  File "/Users/arossert/tests/venv2/lib/python2.7/site-packages/vncdotool/api.py", line 111, in proxy_call
    raise VNCDoException(result)
vncdotool.api.VNCDoException: [Failure instance: Traceback: <type 'exceptions.IOError'>: [Errno 2] No such file or directory: '1/screenshot.png'
/Users/arossert/tests/venv2/lib/python2.7/site-packages/vncdotool/rfb.py:245:_doConnection
/Users/arossert/tests/venv2/lib/python2.7/site-packages/vncdotool/client.py:393:commitUpdate
/Users/arossert/tests/venv2/lib/python2.7/site-packages/twisted/internet/defer.py:460:callback
/Users/arossert/tests/venv2/lib/python2.7/site-packages/twisted/internet/defer.py:568:_startRunCallbacks
--- <exception caught here> ---
/Users/arossert/tests/venv2/lib/python2.7/site-packages/twisted/internet/defer.py:654:_runCallbacks
/Users/arossert/tests/venv2/lib/python2.7/site-packages/vncdotool/client.py:249:_captureSave
/Users/arossert/tests/venv2/lib/python2.7/site-packages/PIL/Image.py:1991:save
```

The CLI is probably catching the exception and ignoring it somewhere.

sibson:
## Summary
- `ThreadedVNCClientProxy.disconnect()` chained a success-only `addCallback` onto `factory.deferred`; a prior command failure (e.g. `captureScreen` to a missing directory) left that deferred errored, so `disconnect()` blocked for its full timeout — forever with the default `timeout=None`.
- Switch to `addBoth`, disconnect using `self.protocol` rather than the chained result (which would be a `Failure`).
- The CLI-side hang in #146 was already fixed by #345; this closes the remaining `api.py` path, confirmed live against the docker fleet.

Closes #146

## Test plan
- [x] `tests/functional/test_api_lifecycle.py::test_disconnect_after_command_error_returns_promptly` added, exercises the failed-capture-then-disconnect path

### DIFF
--- a/vncdotool/api.py
+++ b/vncdotool/api.py
@@ -77,8 +77,10 @@ class ThreadedVNCClientProxy:
             event.set()
 
         def disconnector(_: Any) -> None:
-            # self.protocol, not the chained result: on a failed command
-            # that result is a Failure, not the protocol.
+            # The argument is unused on purpose: addBoth passes whatever
+            # factory.deferred resolved to, which after a failed command is
+            # a Failure, not the protocol. self.protocol was captured
+            # separately in connect() and is still good to disconnect.
             protocol = self.protocol
             if protocol is None:
                 event.set()
--- a/vncdotool/api.py
+++ b/vncdotool/api.py
@@ -77,10 +77,8 @@ class ThreadedVNCClientProxy:
             event.set()
 
         def disconnector(_: Any) -> None:
-            # addBoth, not addCallback: a prior command's failure leaves
-            # factory.deferred errored, and self.protocol -- not the
-            # chained result, which would be a Failure -- is what's still
-            # good to disconnect.
+            # self.protocol, not the chained result: on a failed command
+            # that result is a Failure, not the protocol.
             protocol = self.protocol
             if protocol is None:
                 event.set()
--- a/vncdotool/api.py
+++ b/vncdotool/api.py
@@ -76,11 +76,19 @@ class ThreadedVNCClientProxy:
         def on_disconnected(reason: Any) -> None:
             event.set()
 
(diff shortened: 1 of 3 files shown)


vncdotool/api.py
```diff
@@ -76,11 +76,19 @@ def disconnect(self) -> None:
         def on_disconnected(reason: Any) -> None:
             event.set()
 
-        def disconnector(protocol: VNCDoToolClient) -> None:
+        def disconnector(_: Any) -> None:
+            # The argument is unused on purpose: addBoth passes whatever
```
sibson: Comment unnesseaary.  using self.protocol is a reasonable thing to do.   not sure we even need the protocol = self.protocol line.

sibson: 🤖 Claude Code: Dropped both — comment and the local var. Checked: connect()'s capture_protocol and disconnect()'s disconnector both run as callbacks on the same factory.deferred, so all self.protocol access is on the reactor thread. No race for the local var to guard against. Fixed in 7f2e940.

PR #429, opened 2026-08-23
Decode Raw and CopyRect through a generator pump, with a whole-rectangle fast path

sibson:
## Summary

Phase 2 of the decoder architecture rework: Raw and CopyRect move off the old continuation-passing decode path onto a generator-based pump with a decoder registry. A `PixelDecoder.wholeRectangle()` fast path lets Raw skip the buffer/blit machinery entirely, since it already has all its bytes in one read.

## Perf

Measured on `tigervnc-raw-bgrx8888` (8 updates x 87 rects, 300 replays), best/median in microseconds:

| | best | median |
|---|---|---|
| pre-Phase-2 baseline | 808 | 863 |
| pump, first written | 910 | 975 |
| pump + whole-rectangle path | 577 | 639 |

Net ~25-30% faster than the pre-Phase-2 baseline, not just a regression fix. Full writeup and rejected alternatives in `specs/decoder-architecture.md`.

## Test plan

- [x] `make test` (unit suite)
- [x] Benchmark comparison recorded in the design doc

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,7 @@
 2.0.0.dev0 (UNRELEASED)
 ----------------------
+  - Malformed or oversized pixel data now ends the session with a reported error instead of waiting for bytes that never arrive (@sibson)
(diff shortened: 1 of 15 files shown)


sibson:
🤖 Claude Code: Superseded by #430. This branch forked before Phase 2 (#402) landed on main and independently reimplemented the same generator-pump work, so merging it as-is would have reverted Phase 3/4 (RRE/CoRRE, Hextile) and a lot of unrelated infra. #430 rebuilds just the whole-rectangle fast path on top of current main.

PR #430, opened 2026-08-23
Raw hands its bytes to the client without a buffer to blit through

sibson:
## Summary

`PixelDecoder` may declare a `wholeRectangle()` byte count when its whole output is one contiguous run of pixels — Raw's is, on every rectangle. The pump then reads them and calls `updateRectangle` directly, skipping the buffer/generator/blit machinery entirely. Defaults to `None`, so every other decoder keeps the generator path unchanged.

Supersedes #429, which forked before Phase 2 (#402) landed on main and duplicated it independently. This reimplements the same idea directly on top of current main instead of rebasing that history.

## Perf

Measured on `tigervnc-raw-bgrx8888` (8 updates x 87 rects, 300 replays):

| | best | median |
|---|---|---|
| generator pump, no fast path | 634us | 680us |
| + wholeRectangle fast path | 561us | 606us |

~11% faster. Writeup in `specs/decoder-architecture.md`.

## Test plan

- [x] `make test` (302 tests, unit suite)
- [x] `flake8`
- [x] Benchmark comparison recorded in the design doc


vncdotool/rfb.py
```diff
@@ -429,15 +429,40 @@ def _pumpFor(self, decoder: decoders.Decoder) -> Callable[..., None]:
     def _pumpPixels(
         self, decoder: decoders.PixelDecoder, x: int, y: int, width: int, height: int
     ) -> None:
+        pixel_format = self.pixel_format
+        size = decoder.wholeRectangle(width, height, pixel_format)
```
sibson: this appears to be the only call site.  Why use a function rather than inline as this is perf code?

vncdotool/rfb.py
```diff
@@ -429,15 +429,40 @@ def _pumpFor(self, decoder: decoders.Decoder) -> Callable[..., None]:
...
         if target is None:
             return
         self._pumpBlock(
             None,
-            decoder.decodePixels(target, self.pixel_format),
+            decoder.decodePixels(target, pixel_format),
             (decoder, target, (x, y, width, height)),
         )
 
+    def _finishWholeRectangle(
```
sibson: Should the non-fast path also call this?

vncdotool/rfb.py
sibson: 🤖 Claude Code: Fair point. Moved it: `_pumpFor` now decides once per connection whether a decoder ever answers `wholeRectangle` (`type(decoder).wholeRectangle is not PixelDecoder.wholeRectangle`) and routes it to a new `_pumpWholeRectangle`, same as the existing PixelDecoder/ClientDecoder split. Every non-Raw decoder now goes through `_pumpPixels` unchanged and never calls `wholeRectangle` at all — the per-rectangle call is gone, not just cheap. b1f57e7.

vncdotool/rfb.py
sibson: 🤖 Claude Code: Yes, already does — `_rectBuffer` calls `_rectFits` internally (further down this same diff), so the buffered path gets the same R6 dimension check. And `_doConnection` is already called on the buffered path too, from `_pumpBlock` after `_finishRectangle`.

sibson: 🤖 Claude Code: Misread this the first time — my prior reply was about _rectFits, not what you were pointing at. You're right: _finishWholeRectangle and _finishRectangle each called updateRectangle with the same four args and the same decoder.output_format(self.pixel_format) tag, differing only in how the bytes were obtained (block already-bytes vs target.tobytes()). Factored that into a shared _paintRectangle, called by both. 17e5a28.

sibson: 🤖 Claude Code: Went further — collapsed to one _finishRectangle(block, decoder, rect, advance) taking the bytes directly rather than their source, so _paintRectangle isn't needed either. b7b8205.

specs/decoder-architecture.md
```diff
@@ -163,8 +163,8 @@ no tag to keep in step with the class hierarchy.
 
 Decoders depend on nothing from `rfb.py`: the pump keeps `_rectBuffer`,
 `_pumpBlock` and `_doConnection` to itself, and the entry points that use them —
-`_pumpPixels` and `_pumpForClient` — live with the pump rather than on the
-decoder.
+`_pumpPixels`, `_pumpWholeRectangle` and `_pumpForClient` — live with the pump
```
sibson: narriating a change?

sibson:
🤖 Claude Code: Recorded — `make bench-record` at e78038f: best 570us, median 617us, 1264 total calls (matches your -870 delta exactly). Numbers also written up in `specs/decoder-architecture.md` (N1 section).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

specs/decoder-architecture.md
```diff
@@ -185,6 +185,37 @@ The genuine special cases are the pseudo-encodings phase 5 migrates: `LastRect`
 mutates the rectangle loop counter, and the QEMU extended key encoding mutates
 `negotiated_encodings` and removes the entry it just appended to `rectanglePos`.
 
+### A decoder that is only pixels does not need the machinery
```
sibson: no diea what that even means bad titile

specs/decoder-architecture.md
```diff
@@ -185,6 +185,37 @@ The genuine special cases are the pseudo-encodings phase 5 migrates: `LastRect`
...
+
+The apparatus above — a buffer, a generator, two `send` calls, a blit, a
+`tobytes` — exists to let a decoder write a rectangle in pieces, out of order,
+over several reads. Raw writes it in one piece, in order, in one read, and pays
+for all of it anyway. Since Raw is the only encoding in live use, that is the
+common case paying for the general one.
+
+So a `PixelDecoder` may answer `wholeRectangle(width, height, pixel_format)`
```
sibson: isFullScreen() might be a better name?

specs/decoder-architecture.md
```diff
@@ -185,6 +185,37 @@ The genuine special cases are the pseudo-encodings phase 5 migrates: `LastRect`
...
+with a byte count, meaning: my entire output is that many bytes of pixels, laid
+out left-to-right and top-to-bottom in `output_format`. The pump then reads
```
sibson: and shortcuts the intermeediate buffer

specs/decoder-architecture.md
```diff
@@ -185,6 +185,37 @@ The genuine special cases are the pseudo-encodings phase 5 migrates: `LastRect`
...
+them and calls `updateRectangle` with them directly, and no buffer is built.
+
+It is one optional method with a `None` default, not a second architecture. A
```
sibson: unneeded

specs/decoder-architecture.md
```diff
@@ -185,6 +185,37 @@ The genuine special cases are the pseudo-encodings phase 5 migrates: `LastRect`
...
+decoder that says nothing gets the generator path unchanged, which is what
+every encoding after Raw is expected to want — none of RRE, Hextile, ZRLE or
+Tight can answer it, because none of them knows its byte count before
+decoding.
+
+Whether a decoder ever answers `wholeRectangle` is fixed for the class, not
+per rectangle, so it is resolved the same place and the same way the
```
sibson: describing code, we can read code to undrstand this

vncdotool/decoders/buffer.py
```diff
@@ -34,15 +34,20 @@ def _check_rect(self, x: int, y: int, w: int, h: int) -> None:
             )
 
     def blit(self, x: int, y: int, w: int, h: int, pixels: bytes) -> None:
-        self._check_rect(x, y, w, h)
-        expected = w * h * self.bypp
-        if len(pixels) != expected:
-            raise DecodeError(f"blit expected {expected} bytes, got {len(pixels)}")
-
         if x == 0 and y == 0 and w == self.width and h == self.height:
+            # Covering the buffer exactly already proves the write is in
```
sibson: # full screen update, just do it

vncdotool/decoders/buffer.py
```diff
@@ -34,15 +34,20 @@ def _check_rect(self, x: int, y: int, w: int, h: int) -> None:
...
+            # bounds, so _check_rect is skipped rather than called.
+            expected = self._nbytes
+            if len(pixels) != expected:
+                raise DecodeError(f"blit expected {expected} bytes, got {len(pixels)}")
             self._whole = bytes(pixels)
             self._covered = True
             return
+
+        self._check_rect(x, y, w, h)
```
sibson: check_rect() needs a bettern name,  rect_contained_by_screen?

vncdotool/rfb.py
```diff
@@ -423,9 +423,24 @@ def _handleRectangle(self, block: bytes) -> None:
 
     def _pumpFor(self, decoder: decoders.Decoder) -> Callable[..., None]:
         if isinstance(decoder, decoders.PixelDecoder):
+            # A decoder's wholeRectangle support is fixed for its class, not
```
sibson: yes, that is what this code does, why narrate it?

vncdotool/rfb.py
```diff
@@ -423,9 +423,24 @@ def _handleRectangle(self, block: bytes) -> None:
...
+            if type(decoder).wholeRectangle is not decoders.PixelDecoder.wholeRectangle:
+                return self._pumpWholeRectangle
             return self._pumpPixels
         return self._pumpForClient
 
+    def _pumpWholeRectangle(
+        self, decoder: decoders.PixelDecoder, x: int, y: int, width: int, height: int
+    ) -> None:
+        size = decoder.wholeRectangle(width, height, self.pixel_format)
+        # _pumpFor never routes here a decoder whose wholeRectangle can
```
sibson: unnneded, discoverable by reasoning

vncdotool/rfb.py
```diff
@@ -448,24 +463,37 @@ def _pumpForClient(
 
     def _finishRectangle(
         self,
+        block: bytes,
         decoder: decoders.PixelDecoder,
-        target: decoders.RectBuffer,
         rect: tuple[int, int, int, int],
+        advance: bool,
     ) -> None:
+        """`advance` is False from the buffered path, which reaches this
```
sibson: bending my mind to understand

vncdotool/rfb.py
```diff
@@ -448,24 +463,37 @@ def _pumpForClient(
...
-            x, y, width, height, target.tobytes(),
-            decoder.output_format(self.pixel_format),
+            x, y, width, height, block, decoder.output_format(self.pixel_format)
         )
+        if advance:
+            self._doConnection()
 
-    def _rectBuffer(self, width: int, height: int) -> decoders.RectBuffer | None:
-        """A buffer for one rectangle, or None having failed the connection."""
+    def _rectFits(self, width: int, height: int) -> bool:
```
sibson, line 470: is this the same as our other check_rect()?

vncdotool/rfb.py
```diff
@@ -448,24 +463,37 @@ def _pumpForClient(
...
         limit_w = self.width or self.MAX_DESKTOP_SIZE
         limit_h = self.height or self.MAX_DESKTOP_SIZE
         if not (0 <= width <= limit_w and 0 <= height <= limit_h):
             self.abortConnection(
                 f"rectangle {width}x{height} does not fit a {limit_w}x{limit_h} framebuffer"
             )
+            return False
+        return True
+
+    def _rectBuffer(self, width: int, height: int) -> decoders.RectBuffer | None:
```
sibson: name maybe could be better

specs/decoder-architecture.md
sibson: 🤖 Claude Code: Fair — stopped enumerating pump method names in that sentence; it now says what the pump keeps to itself without a list that needs editing every time one is added. 888e997.

specs/decoder-architecture.md
sibson: 🤖 Claude Code: Retitled: ### wholeRectangleSize skips the buffer for a decoder that doesn't need it. 888e997.

specs/decoder-architecture.md
sibson: 🤖 Claude Code: Renamed to wholeRectangleSize instead — isFullScreen() reads as a bool but the method returns a byte count (or None), so a predicate-shaped name would misstate the contract. 888e997.

specs/decoder-architecture.md
sibson: 🤖 Claude Code: Used it — the section now says Raw shortcuts the intermediate buffer. 888e997.

specs/decoder-architecture.md
sibson: 🤖 Claude Code: Cut. 888e997.

specs/decoder-architecture.md
sibson: 🤖 Claude Code: Cut. 888e997.

vncdotool/decoders/buffer.py
sibson: 🤖 Claude Code: Deleted. 888e997.

vncdotool/decoders/buffer.py
sibson: 🤖 Claude Code: Renamed, but not to rect_contained_by_screen -- this one checks bounds against the RectBuffer's own width/height, not the screen (that's rfb.py's separate _rectFits, see below). Went with _check_bounds. 888e997.

vncdotool/rfb.py
sibson: 🤖 Claude Code: Deleted. 888e997.

vncdotool/rfb.py
sibson: 🤖 Claude Code: Deleted. 888e997.

vncdotool/rfb.py
sibson: 🤖 Claude Code: Removed the advance parameter entirely rather than reword the docstring -- it always called _doConnection when painting, so it now just does that unconditionally. _pumpBlock's ClientDecoder branch (nothing painted) calls _doConnection itself instead of threading a bool through the painting path to suppress it there. 888e997.

vncdotool/rfb.py
sibson: 🤖 Claude Code: No -- different scope. This one (_rectFits) checks a wire rectangle against the framebuffer/screen dimensions, R6's malformed-input guard. buffer.py's was checking a sub-rectangle against the RectBuffer's own width/height (partial blits/fills within one rectangle, e.g. RRE subrects) -- renamed that one to _check_bounds so the two stop reading alike. 888e997.

vncdotool/rfb.py
sibson: 🤖 Claude Code: Renamed _rectBuffer to _allocateRectBuffer -- reads as the factory method it is. 888e997.

vncdotool/rfb.py
```diff
@@ -423,13 +423,24 @@ def _handleRectangle(self, block: bytes) -> None:
 
     def _pumpFor(self, decoder: decoders.Decoder) -> Callable[..., None]:
         if isinstance(decoder, decoders.PixelDecoder):
+            if type(decoder).wholeRectangleSize is not decoders.PixelDecoder.wholeRectangleSize:
```
sibson: couldnt this just be a property on the Decoder class that we check?

vncdotool/rfb.py
```diff
@@ -423,13 +423,24 @@ def _handleRectangle(self, block: bytes) -> None:
...
+        assert size is not None
+        if not self._rectFits(width, height):
+            return
+        self.expect(self._finishRectangle, size, decoder, (x, y, width, height))
+
     def _pumpPixels(
         self, decoder: decoders.PixelDecoder, x: int, y: int, width: int, height: int
     ) -> None:
-        target = self._rectBuffer(width, height)
+        target = self._allocateRectBuffer(width, height)
```
sibson: allocateBuffer()

vncdotool/rfb.py
```diff
@@ -423,13 +423,23 @@ def _handleRectangle(self, block: bytes) -> None:
 
     def _pumpFor(self, decoder: decoders.Decoder) -> Callable[..., None]:
         if isinstance(decoder, decoders.PixelDecoder):
+            if not decoder.buffered:
+                return self._pumpDirect
```
sibson: self._pumpRectangle

vncdotool/rfb.py
```diff
@@ -423,13 +423,23 @@ def _handleRectangle(self, block: bytes) -> None:
...
             return self._pumpPixels
```
sibson: self._pumpBufferedRectangle

vncdotool/rfb.py
sibson: 🤖 Claude Code: Already done in f97b9df, just before this landed -- PixelDecoder.buffered: ClassVar[bool] replaces the wholeRectangleSize method entirely. RawDecoder sets buffered = False; _pumpFor checks the attribute directly.

vncdotool/rfb.py
sibson: 🤖 Claude Code: Renamed to _allocateBuffer. 887b178.

vncdotool/rfb.py
sibson: 🤖 Claude Code: Renamed to _pumpRectangle. 887b178.

vncdotool/rfb.py
sibson: 🤖 Claude Code: Renamed to _pumpBufferedRectangle. 887b178.

PR #431, opened 2026-08-23
build: stop tracking uv.lock, which .gitignore already claims to ignore

sibson:
#382 committed `uv.lock`; #383 added it to `.gitignore` but never removed it from the index. A `.gitignore` entry does nothing for an already-tracked file, so the line has been inert since — the lock is tracked *and* ignored at the same time.

It only bites when a branch changes dependencies: uv re-locks on the next `uv run`, the tracked file goes dirty, and the diff either rides along in an unrelated PR or leaves the tree permanently modified. #401 is the example: adding two dev dependencies landed a 323-line lock diff in what was otherwise a three-file change.

`git rm --cached` completes what #383 intended. The file stays on disk and uv keeps regenerating it; git stops watching. `uv sync` now resolves current releases instead of replaying a pinned set — that is the tradeoff #383 already chose, just now actually in force. Worth knowing: CI loses the ability to reproduce an old run exactly, and `astral-sh/setup-uv`'s cache no longer keys on a committed lock.

Tested: `uv sync` from scratch, `make test` 300 pass, `flake8` clean, and `git status` stays clean after a sync that rewrites the lock.


sibson:
🤖 Claude Code: Closing — this had the tradeoff backwards.

uv's docs say the lockfile "should be checked into version control", and both maintainers on astral-sh/uv#10730 recommend that for libraries too. The two objections I was implicitly relying on don't hold: `uv.lock` has no effect on what pip resolves for someone installing from PyPI (`[project.dependencies]` ranges are what ship), and uv's lockfile is universal rather than platform-specific, so one file covers the whole 3.10-3.13 matrix.

The churn that motivated #383's gitignore is real, but untracking pays for it with reproducibility. `uv sync --locked` in CI addresses it directly instead: a stale lockfile becomes a loud failure rather than a silent re-resolve. Replacement PR follows.

PR #438, opened 2026-08-23
Clear mypy's remaining errors in loggingproxy.py

sibson:
`vncdotool` alone: 89 → 79 errors. Six distinct causes in `loggingproxy.py`, each fixed independently:

- `RFBServer.transport` — same MRO conflict #407 already found for this attribute; `cast()` at each call site instead of `isinstance`. `client.py:86` uses `isinstance` because that class genuinely serves both TCP and UNIX sockets; `RFBServer` here is only ever reached via `listenTCP`, so there's no real ambiguity to check — and `isinstance` against a zope Interface silently returns `False` for a plain `Mock`, which was quietly making `setTcpNoDelay` a no-op under every existing test. Added `TestRFBServer` to catch that.
- `VNCLoggingClient.recorder` — set externally right after construction, never declared (same shape as the `RFBFactory.username` gap); declared it with the `assert` guard `VNCLoggingServerProxy.recorder` already uses.
- `NullTransport` — wasn't actually a valid `ITransport` (missing `getPeer`/`getHost`, wrong `writeSequence` signature); now a real `@implementer(ITransport)`.
- `VNCLoggingServerProxy.factory` — one `# type: ignore[assignment]`; Twisted's `Factory[T]` is invariant, so narrowing it properly would mean casting at all 16 attribute-access call sites instead.
- `getPeer().host` — `IAddress` doesn't guarantee `.host`; added `_peerHost()` (a `getattr` with a `None` fallback) rather than importing/`isinstance`-checking two address classes for a log line.
- `connectionLost` signature — restored the default `reason` argument (`= connectionDone`) that `Protocol.connectionLost` promises callers. This one's a genuine, if never-yet-triggered, Liskov violation, not just a typing nit — `client.py:89` has the identical bug, out of scope here.

Tested: `make test` 299 pass (adds a `connectionLost()`-with-no-argument regression test), `flake8` clean, `mypy vncdotool` 79 (`loggingproxy.py`: 0).

### DIFF
--- a/vncdotool/loggingproxy.py
+++ b/vncdotool/loggingproxy.py
@@ -8,12 +8,14 @@ import socket
 import sys
 import time
 from struct import unpack, unpack_from
-from typing import IO, Callable, Sequence
+from typing import IO, Callable, Iterable, Sequence, cast
 
 from twisted.internet.error import ReactorNotRunning
-from twisted.internet.protocol import Protocol
+from twisted.internet.interfaces import IAddress, ITCPTransport, ITransport
+from twisted.internet.protocol import Protocol, connectionDone
 from twisted.protocols import portforward
 from twisted.python.failure import Failure
+from zope.interface import implementer
 
 from . import __version__ as VNCDOTOOL_VERSION
 from .capture import CaptureWriter, HandshakeScrubber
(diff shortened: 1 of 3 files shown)


vncdotool/loggingproxy.py
```diff
@@ -46,7 +48,8 @@ class RFBServer(Protocol):
 
     def connectionMade(self) -> None:
         super().connectionMade()
-        self.transport.setTcpNoDelay(True)
+        if isinstance(self.transport, ITCPTransport):
```
sibson: this is a little yuck, typing forcing an instance check

sibson: 🤖 Claude Code: Agreed, and it was worse than yuck — checked, and `isinstance` against a zope Interface silently returns False for a plain Mock, so this was quietly turning `setTcpNoDelay` into a no-op under every existing test. Root cause: unlike client.py's transport (genuinely TCP-or-UNIX, where isinstance is real), this RFBServer is only ever reached via listenTCP — no real ambiguity to check. Switched to cast() and added a regression test (TestRFBServer) that fails under the old isinstance code and passes now. Pushed.

PR #439, opened 2026-08-23
Migrate ZRLE to the pluggable decoder registry

sibson:
Phase 5 (ZRLE only) of the decoder migration in specs/decoder-architecture.md: moves ZRLE off rfb.py's callback chain onto the generator-based decoder registry, alongside Raw/CopyRect/RRE/CoRRE/Hextile. `--encodings zrle` now works.

Fixes two defects the old code carried while ZRLE was unreachable: it hardcoded the low-byte CPIXEL placement instead of reading it from the negotiated format, and it unpacked packed-palette tiles as one continuous bitstream instead of byte-padding each tile row independently (RFC 6143 §7.7.6). The row-padding bug was found by replaying a live tigervnc capture and hand-tracing it against the RFC — see the new regression test that uses a non-byte-aligned tile width to catch it.

Validated against every server the local fleet has: full docker-compat round trip (tigervnc, tigervnc-auth, x11vnc, libvncserver-example) and the encodings/pixel-format functional suites, all green; goldens captured fresh from tigervnc at both pixel formats. Cursor and the control pseudo-encodings are left for a follow-up PR.

Tests: `make test` (314 unit), flake8 clean, functional suite against the docker fleet (29 tests) all pass. mypy shows only the same pre-existing baseline errors every other decoder module has.


vncdotool/decoders/zrle.py
```diff
@@ -0,0 +1,135 @@
...
+            shift = 8 - bits - slot * bits
+            indices.append((b >> shift) & mask)
+    return indices
+
+
+class ZRLEDecoder(PixelDecoder):
+    ENCODING: ClassVar[Encoding] = Encoding.ZRLE
+
+    def __init__(self) -> None:
+        # RFC 6143 7.7.6: one zlib stream per connection, rectangles decoded
```
sibson, line 43: over descriptive

PR #440, opened 2026-08-23
Phase 5: migrate Cursor, DesktopSize, QEMU key to the decoder registry

sibson:
Adds a third decoder base class, `ControlDecoder`, for pseudo-encodings that call a client method as a side effect but carry no framebuffer content (DesktopSize, QEMU extended key). Cursor stays a `ClientDecoder` since it does produce content, same as CopyRect.

Also moves `rectanglePos` bookkeeping from append-before-dispatch (with DesktopSize/QEMU undoing it after) to append-on-success, in each pump path. Fixes a latent bug where a rect that failed validation or errored mid-decode still left a phantom `rectanglePos` entry.

LastRect stays inline in `rfb.py` -- it mutates the rectangle loop counter itself rather than calling a client method, which doesn't fit either base class. Deferred to a follow-up PR.

No user-visible behavior change (R5 in `specs/decoder-architecture.md`): these pseudo-encodings already worked via the old inline handling.

**Tested:** `make test` (318 unit tests) and `flake8` clean.

### DIFF
--- a/vncdotool/decoders/base.py
+++ b/vncdotool/decoders/base.py
@@ -42,3 +42,9 @@ class PixelDecoder(Decoder):
 
 class ClientDecoder(Decoder):
     """Consumes bytes, calls a client method."""
+
+
+class ControlDecoder(Decoder):
+    """Calls a client method as a side effect; carries no framebuffer
+    content, so the pump never records its rectangle as a screen change.
+    """
--- a/vncdotool/decoders/base.py
+++ b/vncdotool/decoders/base.py
@@ -27,7 +27,11 @@ class Decoder:
         raise NotImplementedError
 
 
-class PixelDecoder(Decoder):
+class RectDecoder(Decoder):
+    """Its rectangle is a real screen change, recorded as one."""
+
+
+class PixelDecoder(RectDecoder):
     """Consumes bytes, fills a rect buffer."""
 
     # False when the decoder's wire bytes are already its output bytes, in order.
@@ -40,7 +44,7 @@ class PixelDecoder(Decoder):
         return pixel_format
 
 
-class ClientDecoder(Decoder):
+class ClientDecoder(RectDecoder):
     """Consumes bytes, calls a client method."""
 
 
--- a/vncdotool/decoders/base.py
+++ b/vncdotool/decoders/base.py
@@ -45,6 +45,5 @@ class ClientDecoder(Decoder):
 
 
 class ControlDecoder(Decoder):
-    """Calls a client method as a side effect; carries no framebuffer
-    content, so the pump never records its rectangle as a screen change.
-    """
+    """Calls a client method as a side effect; its rectangle is never
+    recorded as a screen change."""
--- a/vncdotool/decoders/control.py
+++ b/vncdotool/decoders/control.py
@@ -1,7 +1,4 @@
-"""DesktopSize and QEMU extended key pseudo-encodings: no payload, a client
-side effect only. LastRect stays on rfb.py's own path -- it mutates the
-rectangle loop counter itself rather than calling a client method.
-"""
+"""DesktopSize and QEMU extended key pseudo-encodings. rfbproto."""
 from __future__ import annotations
 
 from typing import ClassVar, Iterator
--- a/vncdotool/decoders/control.py
+++ b/vncdotool/decoders/control.py
@@ -1,0 +1,34 @@
+"""DesktopSize and QEMU extended key pseudo-encodings: no payload, a client
+side effect only. LastRect stays on rfb.py's own path -- it mutates the
+rectangle loop counter itself rather than calling a client method.
+"""
+from __future__ import annotations
+
+from typing import ClassVar, Iterator
+
+from ..const import Encoding
+from ..pixelformat import PixelFormat
+from .base import ControlDecoder
+
+
+class DesktopSizeDecoder(ControlDecoder):
+    ENCODING: ClassVar[Encoding] = Encoding.PSEUDO_DESKTOP_SIZE
+
+    def decodeForClient(
+        self, client: object, rect: tuple[int, int, int, int], pixel_format: PixelFormat
+    ) -> Iterator[int]:
+        _, _, width, height = rect
+        client.updateDesktopSize(width, height)
+        return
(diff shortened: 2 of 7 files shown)


vncdotool/decoders/control.py
```diff
@@ -0,0 +1,31 @@
...
+
+class DesktopSizeDecoder(ControlDecoder):
+    ENCODING: ClassVar[Encoding] = Encoding.PSEUDO_DESKTOP_SIZE
+
+    def decodeForClient(
+        self, client: object, rect: tuple[int, int, int, int], pixel_format: PixelFormat
+    ) -> Iterator[int]:
+        _, _, width, height = rect
+        client.updateDesktopSize(width, height)
+        return
```
sibson, line 19: should we raise StopIteration?  both yeild and return is odd

vncdotool/decoders/base.py
```diff
@@ -40,5 +44,10 @@ def output_format(self, pixel_format: PixelFormat) -> PixelFormat:
         return pixel_format
 
 
-class ClientDecoder(Decoder):
+class ClientDecoder(RectDecoder):
```
sibson, line 47: ParameterizedControlDecoder?

vncdotool/decoders/control.py
```diff
@@ -0,0 +1,31 @@
...
+
+from ..const import Encoding
+from ..pixelformat import PixelFormat
+from .base import ControlDecoder
+
+
+class DesktopSizeDecoder(ControlDecoder):
+    ENCODING: ClassVar[Encoding] = Encoding.PSEUDO_DESKTOP_SIZE
+
+    def decodeForClient(
```
sibson, line 14: this passes a rect in, how is it not a ClientDecoder?

vncdotool/decoders/control.py
```diff
@@ -0,0 +1,31 @@
...
+        self, client: object, rect: tuple[int, int, int, int], pixel_format: PixelFormat
```
sibson, line 15: discards rect

vncdotool/decoders/control.py
```diff
@@ -0,0 +1,31 @@
...
+        yield  # pragma: no cover -- never runs; keeps this a generator
+
+
+class QemuExtendedKeyDecoder(ControlDecoder):
+    ENCODING: ClassVar[Encoding] = Encoding.PSEUDO_QEMU_EXTENDED_KEY_EVENT
+
+    def decodeForClient(
+        self, client: object, rect: tuple[int, int, int, int], pixel_format: PixelFormat
+    ) -> Iterator[int]:
+        client.negotiated_encodings.add(Encoding.PSEUDO_QEMU_EXTENDED_KEY_EVENT)
```
sibson, line 29: discards rect, doe s it get one passed to it?

vncdotool/rfb.py
```diff
@@ -399,26 +386,38 @@ def _pumpBufferedRectangle(
...
-            decoder.decodePixels(target, self.pixel_format),
-            (decoder, target, (x, y, width, height)),
-        )
+        rect = (x, y, width, height)
+
+        def on_done() -> None:
+            output_format = decoder.output_format(self.pixel_format)
+            self._finishRectangle(target.tobytes(), output_format, rect)
+
+        self._pumpBlock(None, decoder.decodePixels(target, self.pixel_format), on_done)
```
sibson, line 395: can we come up with a better name than pumpBlock.  decodeBytes?

vncdotool/rfb.py
```diff
@@ -452,19 +451,15 @@ def _pumpBlock(
...
             size = generator.send(block)
         except StopIteration:
-            if finish is None:
+            if on_done is None:
                 self._doConnection()
             else:
-                decoder, target, rect = finish
-                output_format = decoder.output_format(self.pixel_format)
-                self._finishRectangle(target.tobytes(), output_format, rect)
+                on_done()
```
sibson, line 462: why use a closure here?

vncdotool/decoders/control.py
sibson: 🤖 Claude Code: Fair point. Fixed by not making it a generator at all — it never consumes bytes, so there's nothing to StopIteration on. New method is a plain `decodeForControl(client, width, height) -> None`. See #444.

vncdotool/decoders/control.py
sibson: 🤖 Claude Code: Right catch. It shouldn't have been — same signature as ClientDecoder made the split invisible at the method level. Now it's decodeForControl(client, width, height) -> None: no rect, no pixel_format, no generator. See #444.

vncdotool/decoders/control.py
sibson: 🤖 Claude Code: Yes, the full rect (currently unused). Now it only gets width/height, and doesn't get them either since QEMU key needs neither -- decodeForControl(client, width, height) -> None, ignored here. See #444.

vncdotool/decoders/base.py
sibson: 🤖 Claude Code: Kept the name -- the awkwardness was the identical decodeForClient signature, not the name itself. Now that ControlDecoder has its own decodeForControl(client, width, height) -> None, distinct from ClientDecoder's generator contract, 'ControlDecoder' reads correctly again without needing 'Parameterized'. See #444.

vncdotool/rfb.py
sibson: 🤖 Claude Code: Fair -- it doesn't decode anything, it just drives whichever generator it's handed until StopIteration. Renamed to _pumpGenerator (decodeBytes would have been misleading, since it's decoder-agnostic). See #444.

vncdotool/rfb.py
sibson: 🤖 Claude Code: Two different completions -- the buffered path finishes through _finishRectangle, the client path just appends+_doConnection -- and the shared driver stays decoder-agnostic rather than branching on type internally, which is what it was refactored away from. Kept the closures (this matches the codebase's own pattern for deferred completions, e.g. api.py's on_disconnected/disconnector/result_callback) but gave them distinct names (finish_buffered/finish_client) instead of both being on_done. See #444.

PR #441, opened 2026-08-24
Fix stale width/height after mid-session desktop resize

sibson:
Fixes `VNCDoToolClient.updateDesktopSize` never updating `self.width`/`self.height` after a mid-session `PSEUDO_DESKTOP_SIZE` resize, even though `self.screen` resized correctly. Also adds a live functional test against `libvncserver-example` -- the one fleet server that can trigger a real resize -- proving the client decodes it correctly; see the commits for why it's the only one that can.

Tests: `make test` (317 passing), `flake8` clean, and the new functional test run live against the fleet, including a check that the existing smoke suite's size assertion for the same server still passes afterward.

### DIFF
--- a/tests/functional/test_desktop_resize.py
+++ b/tests/functional/test_desktop_resize.py
@@ -3,12 +3,9 @@ scene-catalogue testing cannot reach, since the scene player has no way to
 make a server change geometry (see "Deferred" in decoder-goldens.md and
 "Testing" in decoder-architecture.md).
 
-libvncserver-example is the one fleet server that can: its ``dokey()``
-(examples/example.c, LibVNCServer 0.9.14, matching this repo's pinned
-``LIBVNCSERVER_VERSION``) handles ``XK_Up``/``XK_Down`` by calling
-``rfbNewFramebuffer()``, cycling 640x480 -> 800x600 -> 1024x768 and back.
-There is no signal, stdin, or xrandr path -- only those two keysyms, which
-``vncdotool.keys.KEYMAP`` maps "up"/"down" onto exactly.
+libvncserver-example resizes only in response to ``XK_Up``/``XK_Down``
+(examples/example.c, LibVNCServer 0.9.14); no signal, stdin, or xrandr path
+exists.
 
 The cycle position is a static variable in the server process, not
 per-connection state, so a container a previous run already resized reports
@@ -19,7 +16,7 @@ from unittest import TestCase
 
 from PIL import Image
 
-from .utils import HOST, LIBVNCSERVER_EXAMPLE, port_open, run_vncdo, screenshot_dir
+from .utils import HOST, LIBVNCSERVER_EXAMPLE, distinct_colours, has_expected_content, port_open, run_vncdo, screenshot_dir
 
 FLOOR_SIZE = (640, 480)
 RESIZED_SIZE = (800, 600)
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -1,10 +1,18 @@
 name: VNCDotool CI
```
sibson, line 1: probably sholdn't e here

tests/functional/test_desktop_resize.py
```diff
@@ -0,0 +1,70 @@
+"""Mid-session desktop resize: the case specs/decoder-architecture.md's
+scene-catalogue testing cannot reach, since the scene player has no way to
+make a server change geometry (see "Deferred" in decoder-goldens.md and
+"Testing" in decoder-architecture.md).
+
+libvncserver-example is the one fleet server that can: its ``dokey()``
```
sibson: narration

tests/functional/test_desktop_resize.py
```diff
@@ -0,0 +1,70 @@
...
+            self.assertEqual(
+                image.size, FLOOR_SIZE,
+                f"{LIBVNCSERVER_EXAMPLE.name}: two Downs did not floor the framebuffer at {FLOOR_SIZE}",
+            )
+
+        with Image.open(resized_png) as image:
+            self.assertEqual(
+                image.size, RESIZED_SIZE,
+                f"{LIBVNCSERVER_EXAMPLE.name}: one Up from the floor did not grow the framebuffer to {RESIZED_SIZE}",
+            )
```
sibson, line 59: what about actual image content

.github/workflows/ci.yml
sibson: 🤖 Claude Code: agreed, that was a leftover from the first rebase attempt dragging in unrelated commits from a diverged local main. Redid the rebase scoped to just this branch's 3 commits (see the force-push after) — ci.yml isn't part of this diff anymore.

tests/functional/test_desktop_resize.py
sibson: 🤖 Claude Code: fair, trimmed to one fact -- which keysyms trigger a resize and that there's no other path. Dropped the cycling-order/KEYMAP-lookup narrative (50052cf).

tests/functional/test_desktop_resize.py
sibson: 🤖 Claude Code: good catch -- size alone doesn't prove pixels were decoded, just that the buffer was sized right. Added the same distinct_colours/has_expected_content flat-colour check the shared smoke suite uses for test_capture, applied to both the floor and resized captures (50052cf).

tests/functional/test_desktop_resize.py
```diff
@@ -0,0 +1,79 @@
+"""Mid-session desktop resize: the case specs/decoder-architecture.md's
+scene-catalogue testing cannot reach, since the scene player has no way to
```
sibson: narration

tests/functional/test_desktop_resize.py
```diff
@@ -0,0 +1,79 @@
...
+make a server change geometry (see "Deferred" in decoder-goldens.md and
+"Testing" in decoder-architecture.md).
+
+libvncserver-example resizes only in response to ``XK_Up``/``XK_Down``
```
sibson: comment should go with code, not here but then becomes redundent

tests/functional/test_desktop_resize.py
```diff
@@ -0,0 +1,79 @@
...
+(examples/example.c, LibVNCServer 0.9.14); no signal, stdin, or xrandr path
+exists.
+
+The cycle position is a static variable in the server process, not
```
sibson: drop, no idea what this even is trying to say

tests/functional/test_desktop_resize.py
sibson: 🤖 Claude Code: dropped the whole module docstring (8e6ba8a) rather than keep patching it paragraph by paragraph -- the test's own docstring and the key/pause/capture calls already say what it does.

tests/functional/test_desktop_resize.py
sibson: 🤖 Claude Code: agreed, and once it's redundant either way, the fix is to drop it rather than relocate it -- done in 8e6ba8a.

tests/functional/test_desktop_resize.py
sibson: 🤖 Claude Code: dropped -- gone with the rest of the docstring in 8e6ba8a.

PR #442, opened 2026-08-24
Add functional coverage for --localcursor against x11vnc

sibson:
## Summary
- New `tests/functional/test_cursor.py` proves the recently migrated `CursorDecoder` actually decodes a live Cursor pseudo-encoding rectangle end to end, not just in unit tests with a mocked transport.
- x11vnc is the only fleet server that sends a non-degenerate rect (tigervnc answers 0x0).
- Two checks: `--nocursor` vs `--localcursor` isolates that a decode happened at all (x11vnc bakes its own cursor into raw pixels unless the encoding is negotiated, so a plain capture is not a useful contrast); a plain capture vs `--localcursor`, via `imagediff.assert_images_match`, checks the client-side composite actually lands on the same pixels as the server's own render.

## Test plan
- `flake8 --count --statistics vncdotool tests` — clean
- `make test` (unit) — 318 passed
- `make test-func` — new tests pass; pre-existing failures are the macOS Screen Sharing OS-hosted server, unreachable in this environment (CI-only), unrelated to this change

### DIFF
--- a/tests/functional/test_cursor.py
+++ b/tests/functional/test_cursor.py
@@ -1,0 +1,60 @@
+"""``--localcursor`` decodes a live Cursor pseudo-encoding rectangle.
+
+x11vnc is the only fleet server exercised here: it answers the Cursor
+pseudo-encoding with a real 18x18 rectangle. tigervnc answers with a
+degenerate 0x0 rectangle instead, which decodes to no pixels and would pass
+this test whether or not CursorDecoder (vncdotool/decoders/cursor.py) ever
+ran.
+"""
+
+from unittest import TestCase
+
+from PIL import Image, ImageChops
+
+from .utils import HOST, X11VNC, port_open, run_vncdo, screenshot_dir
+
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,10 +1,18 @@
 name: VNCDotool CI
 
+# paths-ignore skips the run only when every changed file matches, so a
```
sibson: does not belong here

sibson: 🤖 Claude Code: Confirmed — those weren't intended changes. Local `main` in this worktree had diverged from `origin/main` (same fixes under different hashes), and `git rebase origin/main` replayed them as 4 spurious commits (ci.yml, CLAUDE.md, test_rfb.py, loggingproxy.py, rfb.py) alongside the real change. Stripped them out with a corrective rebase — this PR is now just the 3 test_cursor.py commits, current tip `44cb4af`.

PR #443, opened 2026-08-24
docs: describe --localcursor/--nocursor behavior, not rationale

sibson:
## Summary
- `--help` for `--localcursor`/`--nocursor` stated a use case rather than behavior; now describes what each flag does.
- Added a paragraph to `docs/usage.rst` covering when they're useful, hedged since other use cases aren't known.

## Test plan
- `flake8 --count --statistics vncdotool tests`: 0
- `python -m unittest discover tests/unit`: 318 passed

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -37,11 +37,14 @@ make screen captures of the session::
 
     > vncdo capture screenshot.png
 
-Some servers don't draw the mouse pointer into the framebuffer themselves, so
-it's absent from captures by default. ``--localcursor`` asks the server for
-the cursor shape and draws it in; ``--nocursor`` is the opposite, forcing the
-pointer out of captures even if the server would have included it. This is
-mostly only useful when the server does not include the cursor itself::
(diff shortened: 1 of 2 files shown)


docs/usage.rst
```diff
@@ -37,6 +37,14 @@ make screen captures of the session::
 
     > vncdo capture screenshot.png
 
+Some servers don't draw the mouse pointer into the framebuffer themselves, so
```
sibson: is that a known fact or assumption?  The reason is to improve responsivness for interactive clinets.

sibson: 🤖 Claude Code: assumption, and wrong one — verified against RFC 6143. The spec's actual reason for the cursor pseudo-encoding is local rendering to cut lag for an interactively-driven session, not servers omitting the cursor from the framebuffer. Reworded in 5042d44 to cite that and to explain why it barely applies to vncdo's scripted (non-interactive) model.

PR #444, opened 2026-08-24
Give ControlDecoder its own non-generator method

sibson:
Follow-up to #440's review comments (crossed with an accidental merge).

**ControlDecoder signature:** `decodeForClient` had the identical signature to `ClientDecoder.decodeForClient` -- nothing distinguished them at the method level, only which base class you'd inherited from. It also needed a `return`/`yield` trick just to satisfy the generator contract despite never yielding. Gives it its own `decodeForControl(client, width, height) -> None`: no `rect`, no `pixel_format`, no generator. `_pumpForControl` drops the generator machinery entirely.

**Pump naming:** `_pumpBlock` renamed to `_pumpGenerator` -- it doesn't decode anything, it drives whichever generator it's handed until StopIteration. Its two completion closures were both named `on_done`; renamed to `finish_buffered`/`finish_client`. Kept as closures -- matches the codebase's existing pattern for deferred completions (api.py's `on_disconnected`/`disconnector`/`result_callback`).

Addresses all 6 unresolved threads on #440.

**Tested:** 318 unit tests, flake8 clean.

### DIFF
--- a/vncdotool/decoders/base.py
+++ b/vncdotool/decoders/base.py
@@ -50,9 +50,10 @@ class ClientDecoder(RectDecoder):
 
 class ControlDecoder(Decoder):
     """Calls a client method as a side effect; its rectangle is never
-    recorded as a screen change. Consumes no bytes, so unlike the other two
-    this is a plain call, not a generator.
-    """
+    recorded as a screen change."""
 
     def decodeForControl(self, client: object, width: int, height: int) -> None:
+        """Unlike decodePixels/decodeForClient, not a generator: this
+        consumes no bytes, so there is nothing to yield for.
+        """
         raise NotImplementedError
--- a/vncdotool/decoders/base.py
+++ b/vncdotool/decoders/base.py
@@ -50,4 +50,9 @@ class ClientDecoder(RectDecoder):
 
 class ControlDecoder(Decoder):
     """Calls a client method as a side effect; its rectangle is never
-    recorded as a screen change."""
+    recorded as a screen change. Consumes no bytes, so unlike the other two
+    this is a plain call, not a generator.
+    """
+
+    def decodeForControl(self, client: object, width: int, height: int) -> None:
+        raise NotImplementedError
--- a/vncdotool/rfb.py
+++ b/vncdotool/rfb.py
@@ -408,8 +408,8 @@ class RFBClient(Protocol):
     def _pumpForControl(
         self, decoder: decoders.ControlDecoder, x: int, y: int, width: int, height: int
     ) -> None:
-        rect = (x, y, width, height)
-        self._pumpBlock(None, decoder.decodeForClient(self, rect, self.pixel_format), None)
+        decoder.decodeForControl(self, width, height)
+        self._doConnection()
 
     def _finishRectangle(
         self,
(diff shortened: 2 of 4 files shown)


vncdotool/decoders/base.py
```diff
@@ -50,4 +50,9 @@ class ClientDecoder(RectDecoder):
 
 class ControlDecoder(Decoder):
     """Calls a client method as a side effect; its rectangle is never
-    recorded as a screen change."""
+    recorded as a screen change. Consumes no bytes, so unlike the other two
```
sibson: other two, decribes a function later

vncdotool/rfb.py
sibson, line 365: where does this get called from, can we avoid the isinstance calls here

vncdotool/decoders/base.py
sibson: 🤖 Claude Code: Right, that sentence was about decodeForControl, not the class. Moved it onto decodeForControl's own docstring and named PixelDecoder/decodeForClient explicitly instead of 'the other two'.

vncdotool/rfb.py
sibson: 🤖 Claude Code: Called once per decoder in __init__ (the _decoders dict comprehension), not per rectangle -- the per-rectangle path in _handleRectangle is just entry = self._decoders.get(encoding); pump(decoder, ...), a dict lookup and a bound-method call, no isinstance. The isinstance cost is deliberately paid here, once per connection, so that path stays branch-free; specs/decoder-architecture.md states this as the design ('Which pump path a decoder takes is resolved once per connection'). Left the mechanism as-is and added a comment saying so, since it's a real reason rather than something to refactor away.

vncdotool/rfb.py
```diff
@@ -363,6 +363,11 @@ def _handleRectangle(self, block: bytes) -> None:
             self._doConnection()
 
     def _pumpFor(self, decoder: decoders.Decoder) -> Callable[..., None]:
+        # Called once per decoder in __init__, not per rectangle: the
```
sibson: why was this added? its obnoxiously long

sibson: 🤖 Claude Code: Fair, cut it to one line -- the rest was already in my reply above and in the cited spec doc, no need for it twice.

vncdotool/rfb.py
```diff
@@ -363,6 +363,8 @@ def _handleRectangle(self, block: bytes) -> None:
             self._doConnection()
 
     def _pumpFor(self, decoder: decoders.Decoder) -> Callable[..., None]:
+        # Called once per decoder at connect time, not per rectangle -- see
```
sibson: uneeded, response to a pr comment in code

sibson: 🤖 Claude Code: Agreed, deleted. It was answering the review thread, not telling a cold reader anything -- that's what the reply above and specs/decoder-architecture.md are for.

vncdotool/decoders/base.py
```diff
@@ -51,3 +51,9 @@ class ClientDecoder(RectDecoder):
 class ControlDecoder(Decoder):
     """Calls a client method as a side effect; its rectangle is never
     recorded as a screen change."""
+
+    def decodeForControl(self, client: object, width: int, height: int) -> None:
```
sibson, line 55: width/height seems like this overfitted to resizedesktop.  Does this also mean it's consuming bytes, which Control decoders aren't suppose to do

PR #451, opened 2026-09-05
Add a whole-rectangle decoder base class and the pump path for it

sibson:
## Summary
Tight needs a decoder path that produces a whole rectangle in one call, carrying its own `PixelFormat` (TPIXEL can be 3 bytes/pixel where the negotiated format is 4) instead of writing into a `RectBuffer`. Adds `WholeRectDecoder` and `_pumpWholeRectangle`, and threads the decoder generator's `StopIteration.value` through `on_done` to get the (bytes, PixelFormat) pair out.

Raw is unaffected — it never enters `_pumpGenerator`. No CHANGELOG entry: nothing is registered/selectable yet, so no user-visible change until Tight itself lands (next in the stack).

## Test plan
- [x] `make test` (unit suite)
- Benchmarked on tigervnc-raw-bgrx8888 to confirm no regression on the untouched Raw path (see commit body for numbers)

### DIFF
--- a/tests/unit/test_pump.py
+++ b/tests/unit/test_pump.py
@@ -10,6 +10,7 @@ from unittest import TestCase, mock
 
 from vncdotool import client, decoders, rfb
 from vncdotool.const import Encoding
+from vncdotool.pixelformat import PixelFormat
 
 FIXTURE = (
     Path(__file__).resolve().parent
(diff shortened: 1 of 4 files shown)


tests/unit/test_pump.py
```diff
@@ -357,3 +385,168 @@ def test_a_rectangle_larger_than_the_framebuffer_is_refused(self) -> None:
         cli.vncProtocolError.assert_called_once()
         cli.transport.loseConnection.assert_called_once()
         cli.updateRectangle.assert_not_called()
+
+
+class TestWholeRectPump(TestCase):
+    """A `WholeRectDecoder` hands back the whole rectangle's bytes and the
```
sibson: repeats the purpose of the class

tests/unit/test_pump.py
```diff
@@ -51,6 +52,33 @@ def raw_update(x: int, y: int, width: int, height: int, pixels: bytes) -> bytes:
...
+    def decodeRect(self, width, height, pixel_format):
+        colour = yield 3
+        (invert,) = yield 1
+        if invert:
+            colour = bytes(byte ^ 0xFF for byte in colour)
+        return colour * (width * height), TPIXEL
+
+
+def whole_rect_update(*rects: tuple[int, int, int, int, bytes]) -> bytes:
+    """A FramebufferUpdate of `FakeWholeRect` rectangles. RFC 6143 7.6.1."""
```
sibson: restates the function name

tests/unit/test_pump.py
```diff
@@ -51,6 +52,33 @@ def raw_update(x: int, y: int, width: int, height: int, pixels: bytes) -> bytes:
     return header + rect_header + pixels
 
 
+# Tight's TPIXEL: three bytes, red-green-blue order (rfbproto, Tight Encoding).
+TPIXEL = PixelFormat(24, 24, False, True, 255, 255, 255, 0, 8, 16)
```
sibson: should be imported from elsewhere?

vncdotool/decoders/base.py
```diff
@@ -44,6 +44,20 @@ def output_format(self, pixel_format: PixelFormat) -> PixelFormat:
         return pixel_format
 
 
+class WholeRectDecoder(RectDecoder):
+    """Consumes bytes, produces the whole rectangle itself.
+
+    The format it returns need not be the negotiated one: Tight's TPIXEL is
```
sibson: nonsensical

tests/unit/test_pump.py
sibson: 🤖 Claude Code: Deleted `TestWholeRectPump`'s class docstring in 9478ce2 — its test method names already carry it. Merged.

tests/unit/test_pump.py
sibson: 🤖 Claude Code: Trimmed `whole_rect_update`'s docstring to just its RFC citation in 9478ce2, dropping the sentence that renamed the function. Merged.

tests/unit/test_pump.py
sibson: 🤖 Claude Code: It was — `pixelformat.TPIXEL_FORMAT`, added by #450. Replaced the hand-rolled duplicate with the import in 9478ce2. Merged.

vncdotool/decoders/base.py
sibson: 🤖 Claude Code: Fixed in 7504ff5 — the docstring described a generic base class by naming Tight's TPIXEL, a format that appears nowhere in that PR. Now states the return-format contract on its own terms, matching `PixelDecoder.output_format` above it. Merged.

PR #452, opened 2026-09-06
Decode Tight fill, copy and palette rectangles

sibson:
## Summary
Registers `TightDecoder` (`--encodings tight`) and decodes FillCompression and BasicCompression's copy/palette filters, over the four independent per-connection zlib streams and their reset bits. GradientFilter and JpegCompression raise `DecodeError` naming what arrived — no fleet server emits either, and JPEG is stage 5 of this stack.

Every case replays bytes captured from real TigerVNC 1.12.0 traffic and checks decoded pixels against the committed scene PNG, so the oracle isn't our own decoder. `tests/goldens/tight_rects.py` (the fixture generator) reads the wire brief independently rather than importing `decoders/tight.py`, so its ~120 framing assertions aren't the decoder graded against itself.

## Test plan
- [x] `make test` (unit suite, 426 tests)
- [x] `flake8 --count --statistics vncdotool tests`


tests/goldens/capture_tight_rects.py
```diff
@@ -0,0 +1,278 @@
+"""Capture the per-rectangle Tight fixture under ``tests/unit/fixtures/tight``.
+
+Run by hand, via `make goldens`; CI replays what this writes.
+
+Beside ``capture.py``, which writes whole-session golden fixtures, this one
+writes a rectangle index: ``rects.bin.gz`` is every Tight rectangle's wire
+bytes concatenated in the order the server sent them, and ``rects.json`` says
+where each begins, how long it is, and how its header framed it.
+"""
```
sibson: why does this need to be a seperate file rather than one capture.py that takes mulitple options?  Cant we reuse 80% of the code between rect types?

vncdotool/decoders/tight.py
```diff
@@ -0,0 +1,199 @@
+"""Tight, RFB encoding 7. ``specs/tight-wire.md`` is the wire brief; it
+pins rfbproto at 152107db63cd34b3536ad8ddf54a0cfc9017a9f9, which is the
```
sibson: commentary

vncdotool/decoders/tight.py
```diff
@@ -0,0 +1,199 @@
...
+EXPLICIT_FILTER = 0x04
+
+FILTER_COPY = 0
+FILTER_PALETTE = 1
+FILTER_GRADIENT = 2
+
+# "if the data size after applying the filter but before the compression is
+# less th[a]n 12, then the data is sent as is" (rfbproto, Tight Encoding).
+# The wire never says which happened; the decoder decides from its own
+# height * rowSize. See specs/tight-wire.md section 6.
```
sibson: 4 lines of comment for a variable assignemnt

vncdotool/decoders/tight.py
```diff
@@ -0,0 +1,199 @@
...
+MIN_TO_COMPRESS = 12
+
+# rfbproto caps a Tight rectangle at 2048 pixels wide. TigerVNC servers
+# before 1.16.0 sent wider Fill rectangles, so its own decoder exempts Fill
+# from the check; see specs/tight-wire.md section 7.
```
sibson: 3 lines for a 1 line assignment

vncdotool/decoders/tight.py
```diff
@@ -0,0 +1,177 @@
+"""Tight, RFB encoding 7. See specs/tight-wire.md for the wire format."""
```
sibson: this comment is pointless and the info is discoverable.

tests/unit/test_decoder_tight.py
```diff
@@ -0,0 +1,383 @@
...
+from typing import Optional, Sequence, Tuple
+
+from vncdotool.decoders import DecodeError
+from vncdotool.decoders.tight import TightDecoder
+from vncdotool.pixelformat import PIXEL_FORMATS, tpixel_bytes
+
+PIXEL_FORMAT = PIXEL_FORMATS["bgrx8888"]
+TBYTES = tpixel_bytes(PIXEL_FORMAT)
+
+FILL = 0x08
```
sibson: should these not already be defined elswhere?

tests/goldens/capture_tight_rects.py
sibson: 🤖 Claude Code: Dropped the file entirely in 572125e. Tight now captures through `capture.py` like every other encoding, and the per-rectangle index it used to write is gone along with its 148KB blob and 1094-line JSON. Regression coverage is the ordinary golden (pulled forward from tight/4a), which `test_goldens.py` picks up with no edit; the framing rules that motivated the separate parser moved into `test_decoder_tight.py` as cases that assemble their own bytes.

vncdotool/decoders/tight.py
sibson: 🤖 Claude Code: Module docstring deleted in 572125e. Both facts it carried live a few lines below it: the encoding number in `Encoding.TIGHT`, and the spec pointer in the inline `specs/tight-wire.md section N` citations.

vncdotool/decoders/tight.py
sibson: 🤖 Claude Code: Cut to one line in 572125e: `# Below this height*rowSize, data arrives raw (specs/tight-wire.md section 6).`

vncdotool/decoders/tight.py
sibson: 🤖 Claude Code: Cut to one line in 572125e: `# Exempts Fill: TigerVNC before 1.16.0 sent wider Fill rects (section 7).`

vncdotool/decoders/tight.py
sibson: 🤖 Claude Code: Deleted in 572125e — agreed, the file it pointed at is discoverable and the same spec sections are cited inline where they actually apply.

tests/unit/test_decoder_tight.py
sibson: 🤖 Claude Code: They were. Now imported from `vncdotool.decoders.tight` (FILL, JPEG, FILTER_COPY, FILTER_PALETTE, FILTER_GRADIENT, MIN_TO_COMPRESS) in 572125e. Checked that this does not blunt the threshold tests: setting MIN_TO_COMPRESS to 16 still fails two of them, because they assert the wire shape (payload present raw / absent when compressed) rather than the constant.

PR #454, opened 2026-09-06
Derive the golden comparison tolerance from the pixel format, and capture Tight at rgb565

sibson:
Every golden so far was 32bpp truecolour, so the per-channel tolerance `conditions.json` reserved had never been anything but zero. A reduced-depth capture needs it, and three comparisons needed it at once.

`pixelformat.channel_tolerance` is the one place it's computed: a channel of n bits reaches only every 2\*\*(8-n)th 8-bit value, so an 8-bit value sits at most `255 >> n` short of one the channel can reach. Per channel rather than one scalar, because rgb565's 6-bit green permits half the error its 5-bit red and blue do.

The three consumers all take that call. `client.expect` gains a second match condition — every pixel within the negotiated format's tolerance — because at rgb565 no screen ever equals an 8-bit target's histogram and `expect FILE 0` polled until it timed out. At 8 bits the new condition is exact equality, which the old one already implied, so no existing script changes behaviour.

Then the rgb565 Tight golden itself. Mutation-checked: swapping the decoder's red and blue at a non-TPIXEL format fails step 1 on pixel (0,0), decoded (41,40,180) against (180,40,40); forcing `tpixel_bytes` to 3 fails the same pixel with (0,0,0).


tests/goldens/scenes.py
```diff
@@ -137,12 +137,43 @@ def stamp_patch(image: Image.Image, key: str) -> None:
     ImageDraw.Draw(image).rectangle(box, fill=colour)
 
 
-def read_patch(image: Image.Image) -> Optional[str]:
+def patch_candidates(image: Image.Image, tolerance: Tuple[int, int, int] = (0, 0, 0)) -> List[str]:
+    """The scene keys this frame's patch could be carrying.
+
+    The key rides in the red channel one unit per key, so a format that
```
sibson: the key sounds important but no idea what it is.  This is awkward language

tests/goldens/scenes.py
```diff
@@ -137,12 +137,43 @@ def stamp_patch(image: Image.Image, key: str) -> None:
...
+    """
     red, green, blue = image.convert("RGB").getpixel(_centre(image))
-    if (green, blue) != (_PATCH_GREEN, _PATCH_BLUE):
+    red_bound, green_bound, blue_bound = tolerance
+    if abs(green - _PATCH_GREEN) > green_bound or abs(blue - _PATCH_BLUE) > blue_bound:
+        return []
+    return [key for key in SCENES if abs(red - ord(key)) <= red_bound]
+
+
+def _nearest_scene(image: Image.Image, candidates: List[str]) -> str:
```
sibson, line 157: this seems like we're guessing something when it should always be deterministic.  We are in control of the test.

tests/unit/fixtures/goldens/tigervnc-zrle-rgbx8888/conditions.json
```diff
@@ -42,5 +42,9 @@
   },
   "pixel_format": "rgbx8888",
   "server": "tigervnc",
-  "tolerance": 0
+  "tolerance": [
```
sibson: why make this multiline and hard to read?

tests/goldens/scenes.py
sibson: 🤖 Claude Code: Reworded in ae4ddf3. The docstring now says what the patch actually is: `stamp_patch` draws a square into the middle of each scene image coloured `(ord(key), _PATCH_GREEN, _PATCH_BLUE)`, so the scene identity reaches the client as ordinary pixels and gets quantized along with them. Also names all four colliding keys (c, d, f, g) rather than three.

tests/unit/fixtures/goldens/tigervnc-zrle-rgbx8888/conditions.json
sibson: 🤖 Claude Code: Fixed in ae4ddf3. `distill.write_fixture` now renders arrays of numbers on one line, so this reads `"tolerance": [0, 0, 0]`. Worth noting the multiline form was pre-existing rather than something this PR introduced — `geometry`, `negotiated_version` and `security_types` were already columns of digits on main, and they collapse now too. Every committed conditions.json was re-serialised through the same writer so the files and the writer agree.

tests/goldens/scenes.py
sibson: 🤖 Claude Code: Agreed, and the root cause is worse than the tiebreak: the patch encodes the scene key as a *colour* (`ord(key)` in the red channel), so it goes through the server's pixel-format quantization like any other pixel. At rgb565 red keeps 5 bits, so c(99), d(100), f(102) and g(103) all land on 96 — hence the guess.

Per discussion, the real fix is to stamp the character glyph rather than a colour derived from it, which is deterministic at any format and deletes `_nearest_scene`, the candidate set and the tolerance plumbing through `patch_candidates`. That is being handled as a separate change, because the patch is baked into the committed scene PNGs and into every golden capture, so it needs `make scenes` plus a re-capture of all fixtures against the fleet — every encoding, not just Tight.

Left the tiebreak in place here, with a note on `_nearest_scene` recording that it is only reachable because the patch carries a colour and that stamping the glyph retires it. Not resolving this thread, since the underlying issue is still open.

PR #455, opened 2026-09-06
Decode Tight JPEG rectangles, and add --jpeg-quality to ask for them

sibson:
Decoding is unconditional: nothing in the wire format stops a server sending `0x9_`, so a client that cannot read one fails in the field. What is conditional is whether we invite it. `--jpeg-quality LEVEL` offers one of the ten JPEG Quality Level pseudo-encodings, and rfbproto is explicit that with none offered JpegCompression is not used at all, so an ordinary capture stays lossless.

The level is the RFB quality level 0-9 the wire carries, not a percentage — the protocol has exactly ten values and TigerVNC's mapping onto libjpeg quality is neither linear nor shared with other servers, so a percentage would invent a mapping no two of them agree on.

Pillow decodes the JFIF stream inside the decoder, emitting 24 bpp RGB tagged `TPIXEL_FORMAT` whatever was negotiated. That is the one deliberate exception to the architecture's "a decoder unit test needs no rendering library"; `specs/tight-encoding.md` records it.

Coverage is a lossy golden captured with `--jpeg-quality 9` (replayed against its scene within the JPEG tolerance) plus nine assembled cases in `test_decoder_tight.py` for what a capture cannot reach: the framing (`0x9_`, compact length, JFIF, no filter byte, no zlib, and the MIN_TO_COMPRESS rule not applying), that output stays 24 bpp RGB at a 16 bpp negotiated format, a reset bit on a type carrying no zlib data, a one-component greyscale JPEG, a wrong-sized image, and a header declaring 65500x65500 to trip Pillow's bomb check.


tests/goldens/capture.py
```diff
@@ -63,32 +75,47 @@ def main() -> int:
         default="raw",
         help="encoding to offer the server, forcing it off Raw [raw]",
     )
+    parser.add_argument(
+        "--jpeg-quality",
+        type=int,
+        help="offer this JPEG Quality Level, 0 (low) to 9 (high); only then "
+        "does a conforming server send a Tight JPEG rectangle at all",
```
sibson: skip the second sentence at the end, it's overly verbose

tests/goldens/capture.py
```diff
@@ -63,32 +67,45 @@ def main() -> int:
         default="raw",
         help="encoding to offer the server, forcing it off Raw [raw]",
     )
+    parser.add_argument(
+        "--jpeg-quality",
+        type=int,
+        help="offer this JPEG Quality Level, 0 (low) to 9 (high); only then "
```
sibson: dupe of text we already removed elsewhere

tests/goldens/capture.py
```diff
@@ -63,32 +67,45 @@ def main() -> int:
...
+    )
     parser.add_argument(
         "--name",
         help="fixture directory name [tigervnc-ENCODING-PIXEL_FORMAT]",
     )
     args = parser.parse_args()
+    lossy = args.jpeg_quality is not None
     name = args.name or f"tigervnc-{args.encoding}-{args.pixel_format}"
+    tolerance = (
+        JPEG_TOLERANCE if lossy
```
sibson: use simple if rather than ternary for readability

tests/goldens/scene-lossy.vdo
```diff
@@ -0,0 +1,26 @@
+# scene.vdo for a lossy capture: `expect` never matches a JPEG frame, so this
```
sibson: didn't we add a tolerance flag to the expect call? So this should still work?

vncdotool/client.py
```diff
@@ -538,15 +547,18 @@ class VNCDoToolFactory(rfb.RFBFactory):
...
     encodings: list[rfb.Encoding] | None = None
+    jpeg_quality: int | None = None
 
     def __init__(self) -> None:
         self.deferred = Deferred()
         self._disconnect_callbacks: list[Callable[[Failure], None]] = []
 
     def buildProtocol(self, addr: object) -> VNCDoToolClient:
-        protocol = super().buildProtocol(addr)
+        # Factory.buildProtocol is typed as returning any IProtocol.
```
sibson: why keep this

vncdotool/command.py
```diff
@@ -608,6 +609,14 @@ def vncdo(argv: list[str] | None = None) -> None:
         help="ask the server for FORMAT (%s) instead of accepting the one it "
         "announces" % ", ".join(sorted(pixelformat.PIXEL_FORMATS)),
     )
+    op.add_option(
+        "--jpeg-quality",
+        type="int",
+        metavar="LEVEL",
+        help="offer the JPEG Quality Level pseudo-encoding for LEVEL, 0 (low) "
+        "to 9 (high). Tight only, and lossy: without it a capture stays "
```
sibson: tight only should be enforced by code with a obvious error message not called out in help.  cature stays exact none is meaningless non-gramatically

tests/goldens/capture.py
sibson: 🤖 Claude Code: Cut in 95f0085 — the help is now just "offer this JPEG Quality Level, 0 (low) to 9 (high)".

tests/goldens/capture.py
sibson: 🤖 Claude Code: Right, that fact was already in three places (command.py help, the client.py constant comment, specs/tight-wire.md section 8). Dropped here in 95f0085.

tests/goldens/capture.py
sibson: 🤖 Claude Code: Changed to a plain if/else in 95f0085.

vncdotool/client.py
sibson: 🤖 Claude Code: Deleted in 95f0085.

vncdotool/command.py
sibson: 🤖 Claude Code: Both done in 95f0085. `--jpeg-quality` without Tight is now a usage error — "--jpeg-quality only applies to Tight; add --encodings tight" — covered by a new test for both the no-encodings and wrong-encoding cases. With the code enforcing it, the help drops to "offer the JPEG Quality Level pseudo-encoding for LEVEL, 0 (low) to 9 (high). Lossy [none]", so the ungrammatical trailing clause is gone too.

tests/goldens/scene-lossy.vdo
sibson: 🤖 Claude Code: Measured it rather than guessing, and no — but not for the reason the comment gave, so I have rewritten it.

Two different tolerances are in play. The one added in #454 (`_quantizedMatch`) is derived from the *negotiated pixel format*: at bgrx8888 it is (0, 0, 0), and the code returns early on `not any(tolerance)`, so it never fires for this capture. JPEG error is lossy-encoding error, not format quantization.

The pre-existing `maxrms` argument on `expect` could in principle absorb it, so I measured the committed JPEG golden: histogram RMS of each decoded step against its own scene PNG, and against every other scene.

```
step   own scene   nearest wrong   wrong key
0           0.00         1378.52   d
s           0.00         1394.25   d
d         166.29          178.16   x
x         171.30          194.95   d
g         169.47          803.22   c
p         250.92          264.04   c
c         274.05          358.52   p
f           0.00         2924.44   g
```

Per step the correct scene is always nearest, but there is no single threshold: accepting `c` needs maxrms >= 274, and at 274 the `d` frame also matches `x` at 178. `expect` thresholds, it does not rank. The lossless capture is 0.00 on every step for comparison.

So `scene-lossy.vdo` stays, and its header now says the measured reason instead of "expect never matches a JPEG frame".

PR #456, opened 2026-09-06
Test Tight against a real TigerVNC server, and measure it against Raw

sibson:
The golden fixtures do come from a real TigerVNC server, but they are captured by hand and then replayed: every CI run since has decoded the same recorded bytes. Nothing has checked that a server still agrees to send Tight at all. This adds functional tests that talk to a live tigervnc on every run — that the server negotiates Tight rather than quietly falling back to Raw, that what it sends decodes to the right picture, and what it costs in bandwidth.

All three hold. Against Raw at 256x192: a flat screen costs 820 bytes instead of 196,899, and dense noise 69,174 instead of 196,803.

Two things worth your attention:

**The encoding-negotiation test was broken and could never have caught a fallback.** It searched `vncdo -v` output for the encoding's name, but the only line carrying that is the client's own "Offering …", printed before the server has answered. Every encoding passed, including CoRRE — which tigervnc answers with Raw. It now reads what the server actually sent, out of the vnclog capture archive. Confirmed by re-adding CoRRE: it fails, naming what really arrived.

**Tight's decode time is recorded, at 1817us over the eight scenes**, alongside the other encodings in `bench.jsonl`. Most of it is `_unpalette`'s per-pixel Python loop, which is worth optimising on its own; it is not a reason to hold this up.

Tested: unit suite and flake8 locally; the functional tests here need the Docker fleet, so CI's functional job is what exercises them.


tests/functional/test_bandwidth.py
```diff
@@ -7,78 +7,59 @@
...
-                proxy.wait(timeout=PROXY_STARTUP_DEADLINE)
-                self.fail(f"vnclog never listened on {PROXY_PORT}")
-
-            proxied = TIGERVNC._replace(port=PROXY_PORT)
-            # A server sends pixels only when asked (RFC 6143 7.5.3), and
-            # of vncdo's commands only capture and expect ask.
-            result = run_vncdo(
-                proxied, "--encodings", encodings, "capture", str(Path(tmp) / "screen.png")
+            # A server sends pixels only when asked (RFC 6143 7.5.3), and of
+            # vncdo's commands only capture and expect ask.
```
sibson, line 37: narrative

PR #458, opened 2026-09-06
Document Tight, and check the architecture kept rfb.py out of it

sibson:
Last of the Tight stack: user documentation, a consolidated CHANGELOG, and the check the whole architecture was built to pass.

`docs/usage.rst` gains an Encodings section — what `--encodings` does, the names it takes, and that Tight sends less than Raw. It also states plainly what is *not* implemented, since each ends a session rather than degrading quietly: the gradient filter, TightPNG, Tight Without Zlib, and the Tight security type that TightVNC servers want before they will authenticate you. `--jpeg-quality` is documented as lossy and off by default, so a capture stays exact unless asked otherwise.

The CHANGELOG entries the stages wrote independently are consolidated into one.

**The R1 check.** The architecture exists so a new encoding is new files plus one registry line, rather than an edit to `rfb.py`. Measured across the whole stack:

| file | insertions | deletions | which stage |
|---|---|---|---|
| `vncdotool/const.py` | 0 | 0 | — |
| `vncdotool/rfb.py` | 30 | 5 | stage 2 alone |
| `vncdotool/decoders/__init__.py` | 12 | 1 | stages 2 and 3 |

`const.py` is untouched and the `rfb.py` diff is entirely stage 2, the whole-rectangle pump — infrastructure, not Tight. Adding the encoding itself cost one line in `decoders/__init__.py`. Verified against merged history rather than asserted.

Docs only, no code.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -75,23 +75,20 @@ preference order, and the server sends whichever of them it has::
     > vncdo --encodings tight capture screen.png
 
 The names are ``raw``, ``copyrect``, ``rre``, ``corre``, ``hextile``,
-``zrle`` and ``tight``.  On ordinary screen content tight sends the least.
+``zrle`` and ``tight``.  Tight sends less than raw on ordinary screen
+content.  Three parts of tight are not implemented: the gradient filter,
+TightPNG, and Tight Encoding Without Zlib.  A rectangle using any of them
+ends the session with a message naming what arrived, exiting 20.  The tight
+*security type*, which TightVNC servers want before they will authenticate
+you, is not implemented either.
 
-Some tight rectangles can be JPEG, which is lossy.  A server only sends
-those to a client that asked for a JPEG quality level, and vncdo asks for
-none by default, so a capture is exact unless you request otherwise.
-``--jpeg-quality`` requests it, on the RFB scale of 0 (low) to 9 (high)
-rather than a percentage::
+Some tight rectangles can be JPEG, which is lossy.  vncdo decodes them
+whether or not it asked for them; a conforming server sends them only to a
+client that asked for a JPEG quality level, and vncdo asks for none by
+default, so a capture is exact unless you request otherwise.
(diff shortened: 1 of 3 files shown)


docs/usage.rst
```diff
@@ -66,6 +66,31 @@ x=100, y=200 and is 400 pixels wide by 250 high you could do::
...
+*security type*, which TightVNC servers want before they will authenticate
+you, is not implemented either.
+
+Some tight rectangles can be JPEG, which is lossy.  vncdo decodes them
+whether or not it asked for them; a conforming server sends them only to a
+client that asked for a JPEG quality level, and vncdo asks for none by
+default, so a capture is exact unless you request otherwise.
+``--jpeg-quality`` asks for one, on the RFB scale of 0 (low) to 9 (high)::
+
+    > vncdo --encodings tight --jpeg-quality 8 capture screen.png
```
sibson, line 91: should provide a sample of multiple encodings.  Also should we just change the default to something like tight,hextile,raw

sibson: 🤖 Claude Code: The example is fixed in #459 — this PR merged a moment before I got to it, so the change went to its own branch. It now leads with `--encodings tight,hextile,raw`, says the order is a preference the server may disregard rather than an instruction, and that a server with none of your encodings sends raw.

On changing the default: I think yes, but as its own change rather than folded into a docs PR, and there is one thing to settle first. A server may send a Tight JPEG rectangle even to a client that offered no quality level — rfbproto says a conforming one will not, and our decoder reads it either way, but a non-conforming server would make `vncdo capture` lossy for someone who never asked for that. Raw-by-default makes captures exact by construction today, and it is worth deciding deliberately whether we keep that guarantee (for example by refusing an unsolicited JPEG rectangle when no quality level was offered) before turning compression on for everyone.

Everything else points at doing it: every other viewer defaults to better than raw, the bandwidth win is the reason the encodings exist, and CI now exercises tight against tigervnc, x11vnc, UltraVNC, macOS Screen Sharing and QEMU/KVM. Happy to write it up as a proposal with the JPEG question settled — say the word.

PR #462, opened 2026-09-06
Offer more than Raw by default

sibson:
`vncdo` now offers `tight,zrle,hextile,raw` instead of Raw alone — about a quarter of the bytes against a server that speaks Tight, at more CPU to decode. `specs/default-encodings.md` carries the measurements behind the ordering.

Based on #464 and should not land before it: the list is lossless, but captures only stay exact because #464 refuses a JPEG rectangle nobody asked for.

Two things the diff does not say. `VNCDoToolClient.encoding` is deleted rather than repointed — it named a single default the client no longer has, and nothing referenced it. And `--jpeg-quality` would have errored asking for `--encodings tight` while Tight was already being offered, so its guard now reads the list actually offered.


vncdotool/decoders/__init__.py
```diff
@@ -57,6 +57,11 @@
     "tight": Encoding.TIGHT,
 }
 
+# What --encodings defaults to, in preference order.
```
sibson: sigh

CHANGELOG.rst
```diff
@@ -4,6 +4,7 @@
   - Add ``vncdo stable SECONDS [FUZZ]`` and ``rstable SECONDS X Y W H [FUZZ]``, waiting until the screen stops changing rather than until it matches a reference image (@sibson)
   - Fix ``rexpect`` polling until ``--timeout`` and ``rcapture`` writing black pixels when the region runs off the screen, and ``expect`` doing the same against a target image larger than the screen. All three now fail with a message naming the region and the screen size, exiting 30 (@sibson)
   - A Tight JPEG rectangle from a server that was never offered a JPEG quality level now ends the session with a named error instead of being decoded. JPEG is lossy, so decoding one nobody asked for made a capture inexact without saying so; pass ``--jpeg-quality`` to ask for them (@sibson)
+  - [BREAKING] ``vncdo`` offers ``tight,zrle,hextile,raw`` instead of Raw alone, sending roughly a quarter of the bytes against a server that speaks Tight, at more CPU to decode. Every encoding on the list is lossless, so captures stay exact; ``--encodings`` still replaces the list, and ``VNCDoToolClient.encoding`` is gone (@sibson)
```
sibson: why is this breaking? it's still lossless so should not have user visible impact

vncdotool/decoders/__init__.py
sibson: 🤖 Claude Code: Dropped the restating line in 08a8d99. `# What --encodings defaults to, in preference order.` said what `DEFAULT_ENCODING_NAMES` already says. Kept the spec pointer, which is the one thing not in the file.

CHANGELOG.rst
sibson: 🤖 Claude Code: Agreed, it is not breaking in the sense the other markers use. Changed to [WARNING] in 08a8d99. Captures are byte-identical and --encodings scripts are unaffected; what changes is bandwidth down and CPU up. The one real API break is VNCDoToolClient.encoding being deleted, which the same entry names.

vncdotool/decoders/__init__.py
sibson: 🤖 Claude Code: Correcting my last reply -- the spec pointer went too, in 57c98ee. The comment block is gone entirely; specs/default-encodings.md is findable from specs/ without a signpost in the source.

docs/usage.md
```diff
@@ -80,8 +80,7 @@ frame is further from its target than any bound can separate from a wrong
 screen. `--expect-blur RADIUS` sets that radius, or `0` turns it off:
 
 ```
-> vncdo --encodings tight --jpeg-quality 5 --expect-fuzz 64 \
-        expect somescreen.png
+> vncdo --jpeg-quality 5 --expect-fuzz 64 expect somescreen.png
```
sibson: I wonder if --expect-fuzz should be --match-fuzz or just --fuzz as it's also used for stable.

PR #463, opened 2026-09-06
Compare screens perceptually in expect, dropping histogram RMS

sibson:
`expect` decided a match by thresholding the root-mean-square difference of two
histograms, discarding where the pixels are. It could not see a change smaller
than the encoding's own noise, and no threshold separated a correct screen from
a wrong one under JPEG or rgb565. It now bounds the perceived colour difference
of every pixel, tuned by `--expect-fuzz` and `--expect-blur`.

Breaking at `2.0.0.dev0`: the number in `expect FILE N` is on a new scale, and
`maxrms` is gone with no alias.

Not obvious from the diff: `scene-lossy.vdo` is deleted because `expect` can now
sequence a lossy capture; the new level-5 fixture is the proof, since level 9 is
the one quality tigervnc encodes without subsampling. `--jpeg-quality` implies a
blur, or it would quietly stop `expect` matching.

Beyond CI: both lossy fixtures re-captured against the fleet, and the
channel-swap mutation re-run, failing at 72 against a bound of 6.

Design: `specs/expect-matching.md`.


vncdotool/client.py
```diff
@@ -231,74 +233,73 @@ def _captureSave(
...
-        return self._expectCompare(None, (x, y, x + w, y + h), maxrms)
+        return self._expectCompare(
+            None, (x, y, x + w, y + h), self._expectFuzz(fuzz), self._expectBlur(blur)
+        )
 
-    def _quantizedMatch(self, image: Image.Image) -> bool:
-        """Whether the screen is as near the target as the negotiated format
-        can get.
+    def _expectFuzz(self, fuzz: float | None) -> float:
+        """What the caller asked for, else the run's default, else what the
```
sibson: ok, what?

vncdotool/client.py
```diff
@@ -231,74 +233,73 @@ def _captureSave(
...
 
-        return self._expectCompare(None, (x, y, x + w, y + h), maxrms)
+        return self._expectCompare(
+            None, (x, y, x + w, y + h), self._expectFuzz(fuzz), self._expectBlur(blur)
+        )
 
-    def _quantizedMatch(self, image: Image.Image) -> bool:
-        """Whether the screen is as near the target as the negotiated format
-        can get.
+    def _expectFuzz(self, fuzz: float | None) -> float:
```
sibson: fuzz: float.  What is the scale? would it make sense to make this a %?  So it's interger numbers we convert to float?

vncdotool/client.py
```diff
@@ -231,74 +233,73 @@ def _captureSave(
...
+        :param blur: blur both screens by this radius before comparing, which
+            is what carries a match through a lossy encoding.
         """
         log.debug("expectScreen %s", filename)
-        return self._expectFramebuffer(filename, 0, 0, maxrms)
+        return self._expectFramebuffer(filename, 0, 0, fuzz, blur)
 
     def expectRegion(
-        self, filename: str, x: int, y: int, maxrms: float = 0
+        self, filename: str, x: int, y: int, fuzz: float | None = None, blur: int | None = None
```
sibson: If someone is passing maxrms today as non-zero how much more/less sensitive is fuzz going to be?

vncdotool/client.py
sibson: 🤖 Claude Code: agreed — that sentence just restated the four lines under it, without even covering the `UnsupportedPixelFormat -> 0` branch. Deleted in the comment sweep (ec2e263). What is left on `_expectFuzz` is the fact that is not in the code: a server sending 5-bit red cannot reproduce most 8-bit values, so an exact comparison never comes true however long it polls.

vncdotool/client.py
sibson: 🤖 Claude Code: the scale is 0-255, but it is not a channel value — it is the per-pixel perceived difference in YIQ, normalised against 35215, the largest that difference can be for any two 8-bit pixels. So 255 means "as different as two pixels get", not "255 units of red".

A percentage would say the same thing more honestly, and I think you are right that it is better. `expect foo.png 5` reading as "within 5%" needs no explanation of what 255 is, and pixelmatch and Playwright both use the same axis as a 0-1 threshold (Playwright defaults to 0.2). The costs: the format-derived default becomes 4% rather than 8 (rgb565 needs 3.02%), and the tight end gets coarse — the level-9 fixture measures 0.4%, so a whole-number percent cannot express it and the fixture would round to 1%.

On int vs float: right now they disagree, and that is a defect either way. `--expect-fuzz` is `type="int"` while the trailing script token is parsed with `float()`, so `expect foo.png 0.5` works and `--expect-fuzz 0.5` is a usage error. Whichever scale we land on, both should take the same kind of number.

My recommendation: percent, accepted as int or float, stored internally as the fraction. Say the word and I will convert it — it is a contained change (`imagematch`, the two option definitions, the docstrings, and the two fixture `conditions.json` bounds).

vncdotool/client.py
sibson: 🤖 Claude Code: there is no conversion, because the two measure different things — a script passing a nonzero `maxrms` has to pick a new number rather than translate the old one.

Measured on the golden base scene (256x192), same image pair through both:

| change | old maxrms | new fuzz |
|---|---|---|
| identical | 0.0 | 0.00 |
| every pixel +1 on every channel | 3978.5 | 0.97 |
| every pixel +4 on every channel | 3978.5 | 3.86 |
| 2x2 red patch | 0.4 | 125.98 |
| 8x8 red patch | 4.9 | 149.76 |
| 32x32 red patch | 83.5 | 163.40 |
| a different scene entirely | 1378.5 | 220.63 |

So it depends entirely on the failure mode. For a small localised change — a button enabling, a spinner stopping, which is what most `expect` calls wait on — fuzz is far more sensitive: a 2x2 patch moves it 126 while it moves maxrms 0.4. For a uniform shift across the whole screen the old metric was more sensitive, though not usefully: note that rows 2 and 3 have the *same* maxrms, because a histogram records that bins moved, not how far.

`expect FILE 0` is unaffected — exact is still exact.

For anything else, the practical advice is to start at `--expect-fuzz 16` and add `--expect-blur 2` if the connection is lossy. I have put the sensitivity numbers in `specs/expect-matching.md`; happy to add a migration note to `docs/usage.rst` too if you would rather it were in the user docs.

vncdotool/client.py
sibson: 🤖 Claude Code: agreed, and done in f669a53 — int, no percent.

`--expect-fuzz` and the trailing script token now take the same kind of number, which they did not before. The token is still *read* as a float so that a script written against the old metric gets told what is wrong with its number rather than having it silently read as the next command:

    > vncdo expect somescreen.png 0.5
    expect takes a whole-number fuzz from 0 (exact) to 255, not 0.5

The score itself stays a float internally: a screen differing in one blue bit scores 0.44, and rounding that onto the whole-number scale would make `expect FILE 0` accept a screen that is not the target.

vncdotool/client.py
sibson: 🤖 Claude Code: migration note added to `docs/usage.rst` in 3e24109 — that an old number cannot be converted and has to be picked again, that `expect FILE 0` still means exact, that 16 is a reasonable place to start, and the one figure showing the direction (a 2x2 patch scores 126 where the old measurement scored 0.4). The full sensitivity table stays in `specs/expect-matching.md` rather than the user docs.

docs/usage.rst
```diff
@@ -50,7 +50,37 @@ captures::
 
 With Pillow_ installed, you can wait for the screen to match a known image::
 
-    > vncdo expect somescreen.png 0
+    > vncdo expect somescreen.png
+
+``expect`` blocks until every pixel of the screen is within a bound of the
+target image, measured as a perceived colour difference: a whole number from 0
+(identical) to 255, where 255 is as far apart as two pixels get. It is not a
```
sibson: verbosity loses clariy, not sure what this is trying to say.  Be more direct rather than explain

docs/usage.rst
```diff
@@ -50,7 +50,37 @@ captures::
...
+Left out, the bound is whatever the negotiated pixel format cannot
+express: a server sending 5-bit red cannot reproduce most 8-bit values, so
+demanding an exact match of one would poll until it timed out. Write a number
+after the filename, or pass ``--expect-fuzz``, to allow more::
+
+    > vncdo expect somescreen.png 16
+    > vncdo --expect-fuzz 16 expect somescreen.png
+
+A lossy encoding needs one thing more, because JPEG moves pixels far enough
+that no bound alone separates a correct screen from a wrong one.
```
sibson: expect-blur should enable by default for jpeg

tests/goldens/capture.py
```diff
@@ -14,22 +15,35 @@
...
 PROXY_PORT = 5999
 PROXY_STARTUP_DEADLINE = 10.0
 CAPTURE_DEADLINE = 60.0
 # Budget for the whole scene script. A scene that never arrives fails here
 # naming the image it waited for, rather than recording the one before it.
 SCENE_DEADLINE = 30.0
 
-# Measured at quality level 9, not computed from the format; specs/decoder-goldens.md.
-JPEG_TOLERANCE = (8, 8, 8)
+# Measured, not derived from the format: the keysym patch is a flat 48x48
```
sibson: explanatory

tests/goldens/capture.py
```diff
@@ -53,6 +67,21 @@ def _start_vnclog(archive: Path) -> subprocess.Popen:
     return proxy
 
 
+def _measured_fuzz(steps: list[distill.Step]) -> int:
+    """How far this capture's own frames landed from their scenes.
+
+    Recording the bound the driver ran under instead would assert only that
```
sibson: huh?

docs/usage.rst
sibson: 🤖 Claude Code: done in f53eeae. `--jpeg-quality` now implies `--expect-blur 2`; `--expect-blur 0` turns it back off. You were right that it was a trap as it stood — asking for a quality level quietly stopped `expect` working.

`capture.py` stops passing the radius and takes the default, so the number a fixture is measured at and the number the driver ran under are one constant instead of two that can drift.

tests/goldens/capture.py
sibson: 🤖 Claude Code: cut to what was measured, in f53eeae:

    # Measured against tigervnc at quality level 0, the worst it offers: the flat
    # 48x48 patch reads back within this.

The neighbouring one went the same way — `# Wide enough for the worst quality level, not the one being captured.`

tests/goldens/capture.py
sibson: 🤖 Claude Code: fair — deleted in f53eeae. The first line restated the function name and the second was commit-body rationale (it is in 7299c6b). The function is four lines and reads on its own.

docs/usage.rst
sibson: 🤖 Claude Code: rewritten in f53eeae, roughly half the length:

    Every pixel has to be within a bound of the target, a whole number from 0
    (identical) to 255. Left out, the bound is whatever the pixel format cannot
    express, so a 5-bit-red server does not poll forever waiting to reproduce an
    8-bit value. To allow more, write it after the filename or pass
    ``--expect-fuzz``.

Dropped the two sentences that were explaining rather than telling: that 255 is the furthest two pixels get, and that the number is not a channel value.

PR #464, opened 2026-09-06
Refuse a Tight JPEG rectangle nobody asked for

sibson:
JPEG is the only lossy thing we decode, and we decoded one whenever it arrived. A server that sends JPEG without ever being offered a quality level would make `vncdo capture` write a lossy screenshot with nothing saying so. The decoder is now told what was offered and refuses what was not.

vnclog's observer runs no handshake and so never calls setEncodings; the proxy now forwards the client's SetEncodings to it, or it would refuse the JPEG rectangles it exists to log.

Step 1 of #462.

### DIFF
--- a/vncdotool/decoders/tight.py
+++ b/vncdotool/decoders/tight.py
@@ -23,6 +23,11 @@ FILTER_COPY = 0
 FILTER_PALETTE = 1
 FILTER_GRADIENT = 2
 
+# The ten JPEG Quality Level pseudo-encodings (specs/tight-wire.md section 8).
+JPEG_QUALITY_LEVELS = frozenset(
+    Encoding(Encoding.JPEG_32 + level) for level in range(10)
+)
+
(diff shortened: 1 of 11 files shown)


vncdotool/decoders/tight.py
```diff
@@ -23,6 +23,11 @@
 FILTER_PALETTE = 1
 FILTER_GRADIENT = 2
 
+# The ten JPEG Quality Level pseudo-encodings (specs/tight-wire.md section 8).
+JPEG_QUALITY_LEVELS = frozenset(
```
sibson: duplicated? constants?

sibson: 🤖 Claude Code: Yes, duplicated. `client.py` built the same ten from its own `Encoding.JPEG_32 + level` comprehension as a list, `tight.py` as a frozenset. Moved to `const.py` as an ordered tuple (a level indexes it) in 313ad96; `tight.py` takes its frozenset from that, `client.py` imports it.

PR #466, opened 2026-09-06
Read ZRLE tiles by index, not a byte at a time

sibson:
ZRLE walked the decompressed block through `iter(data)`, paying a Python-level `next()` per CPIXEL byte, palette index and run length. Reading by index and widening CPIXEL runs with one strided slice assignment per byte lane takes `tigervnc-zrle-bgrx8888` from 33,000us to 2,811us median, 12x.

What the diff does not say:

- The fixture has no RLE tiles — 70 solid, 39 packed-palette, 67 raw. Those branches are rewritten on reasoning, not measurement, and unit tests are their only cover.
- An out-of-range palette index raised `IndexError`, which `_pumpGenerator` does not catch; both forms now raise `DecodeError`. The packed form must tell a bad index from row padding bits.
- The last commit is separate work: `_machine()` digested the OS release, splitting one M4 Pro in two. Dropped, and the 20 rows backfilled into one group.

ZRLE goes from 22x Tight's decode cost to 1.9x, bearing on #462.


tests/goldens/benchmark.py
```diff
@@ -140,15 +140,20 @@ def _machine() -> Dict[str, object]:
     """
     cpu = _cpu_model()
     ci = os.environ.get("GITHUB_ACTIONS") == "true" or os.environ.get("CI") == "true"
+    # Only what identifies the hardware. The OS release is reported beside
```
sibson: narrative

vncdotool/decoders/zrle.py
```diff
@@ -53,81 +76,130 @@ def decodePixels(
...
+                                    f"ZRLE palette index {entry} is past the end of a "
+                                    f"{palette_size}-colour palette"
+                                )
+                            pixel = palette[entry]
+                            if not index & 0x80:
+                                pixel_data += pixel
+                                num_pixels += 1
+                                continue
+
+                        # A run length is a sum: every 255 says another byte
```
sibson: ??

tests/goldens/benchmark.py
sibson: 🤖 Claude Code: Agreed — it argued for the design instead of stating anything the code does not, and that argument is already in the commit body. Deleted in f50ff36. Kept one line at the assignment itself, since nothing stops a later tidy-up from folding `release` back into the dict literal and silently re-digesting it.

vncdotool/decoders/zrle.py
sibson: 🤖 Claude Code: The comment was covering for the variable name — `run_length` held the run *minus one*, so the `+ 1` at both use sites read as arbitrary. Renamed to `run_minus_one` with a single `run` computed once, in f50ff36. The comment now carries only the 255-continuation encoding and cites RFC 6143 7.7.6 so it can be checked.

PR #467, opened 2026-09-06
Cap PR descriptions in words, and enforce the mechanical comment rules

sibson:
A PR description shipped at 315 words while passing the ~15-line cap: the cap counted newlines, which four unwrapped paragraphs pass at any length, and the hook enforcing it never saw a heredoc `--body` at all.

So descriptions are capped at 150 words and must arrive as `--body-file`, which is measurable where a shell string is not. The comment policy gains a pointer to a linter covering the half that needs no judgement — a comment narrating the change, citing an issue, drawing a banner, holding code — which warns as the line is written and refuses the commit, and asks for an independent sweep past ten added comment lines.

The hooks live in `~/.claude/hooks/` and are not in this diff; this is the convention text describing them.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -34,6 +34,14 @@ constraint the language does not express, a bound another system imposes.
 Why the design is this rather than something else goes in the commit body.
 Sweep the diff before committing, not the branch before opening a PR.
 
+`~/.claude/hooks/comment-lint.sh` catches the part of that needing no
+judgement -- a comment narrating the change, citing an issue, drawing a
+banner, holding code -- warning as the line is written and refusing the
+commit. Past ten added comment lines it refuses until a subagent that did not
+write them has swept the diff, since re-reading your own comment supplies the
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -34,6 +34,21 @@ constraint the language does not express, a bound another system imposes.
 Why the design is this rather than something else goes in the commit body.
 Sweep the diff before committing, not the branch before opening a PR.
 
+`~/.claude/hooks/comment-lint.sh` catches the part of that needing no
```
sibson: too verbose, this cost tokens every call

sibson: 🤖 Claude Code: Cut 14 lines to 3 in 97c0b8c — the hook names its own reasons and escapes when it fires, so CLAUDE.md only needs to say it exists. Also dropped the `--body-file` justification from the PR-description paragraph for the same reason.

PR #468, opened 2026-09-06
Give the pump one entry point, whatever a decoder produces

sibson:
The pump picked between five methods with an `isinstance` chain over the decoder base classes, plus a `buffered` flag as a second axis inside one of them, so a decoder with new framing cost an `rfb.py` edit — the half of R1 the Tight stack left open. Each base class now implements one `decode(client, rect, pixel_format)` in terms of the method its subclasses already override, returning an `Outcome`.

Raw pays about 8% for it (590 → 638 us on `tigervnc-raw-bgrx8888`, measured alternating), because it now runs as a one-yield generator rather than parking on `expect` directly. Recovering that means giving the pump a second entry shape, which is the flag this removes; I took the 8%.

`_rectFits` and `_allocateBuffer` become public `requireFits` and `rectBuffer` raising `DecodeError`, since decoders call them now — which changes the abort for an oversized rectangle to start "cannot decode this rectangle: ".


vncdotool/decoders/base.py
```diff
@@ -1,41 +1,88 @@
...
+Rect = Tuple[int, int, int, int]
+
+
+class Outcome(NamedTuple):
+    """What one rectangle did, for the pump to act on: whether it counts as a
+    screen change, and any pixels the pump is left to paste.
+    """
+
+    changed: bool
+    # Both or neither: pixels the pump cannot size are a rectangle silently
```
sibson: this should be enforced in code not commented.  As such this comment describes code

vncdotool/decoders/base.py
```diff
@@ -1,41 +1,88 @@
...
+    # lost, so there are no defaults to write one field without the other.
+    pixels: Optional[bytes]
+    pixel_format: Optional[PixelFormat]
+
+
+# Nothing reached the framebuffer, and the rectangle is not a screen change.
+NOTHING = Outcome(False, None, None)
+
+# The decoder changed the framebuffer itself; the pump has nothing to paste.
+CHANGED = Outcome(True, None, None)
```
sibson: Is there a better name we can use here?

vncdotool/decoders/base.py
sibson: 🤖 Claude Code: Agreed — the comment was describing the code, and the code was not enforcing it. `Paste` now holds the pixels and their format together, and `Outcome` holds one `Paste` or none, so "pixels with no format" is unwriteable rather than commented against. The pump reads one field, and the runtime check and its test are gone. 242737d

vncdotool/decoders/base.py
sibson: 🤖 Claude Code: `painted()` is gone rather than renamed. The thing it built is now a type — `Paste(pixels, pixel_format)` — and decoders write `Outcome(True, Paste(...))`, so there is no second word for the same idea. "Paste" is the vocabulary already used for this in the pump and in specs/decoder-architecture.md ("One paste per rectangle"). 242737d

PR #469, opened 2026-09-06
Test Hextile's subencoding byte as an int, not an IntFlag

sibson:
The largest line in Hextile's profile was `enum.IntFlag.__and__`, at 38% of the run once cumulative cost is counted — ahead of anything in this package. Each tile built a `HextileEncoding` and tested it seven times, and every `&` constructed a new flag object to be truth-tested. Binding the five bits to locals and testing the raw byte takes `tigervnc-hextile-bgrx8888` from 7,128us to about 4,800us median, ~1.45x.

The construction was validating nothing: `IntFlag`'s default boundary is `KEEP`, so `HextileEncoding(0xFF)` kept the undefined bits rather than rejecting them. No behaviour change.

`buffer.fill` now leads at 26%. Guarding its two no-op helper calls measured 3.8% and is not here, for a duplicated guard at four call sites. What is left is composing a tile locally and blitting once, which `specs/decoder-architecture.md` argues against — a decision, not a perf commit.


vncdotool/decoders/hextile.py
```diff
@@ -19,6 +19,13 @@ def decodePixels(
         self, target: RectBuffer, pixel_format: PixelFormat
     ) -> Iterator[int]:
         bypp = target.bypp
+        # Plain ints: IntFlag's & builds a new flag object per test, and
```
sibson: explainy

vncdotool/decoders/hextile.py
```diff
@@ -19,6 +19,13 @@ def decodePixels(
...
+        # these are tested seven times per 16x16 tile.
+        raw_bit = HextileEncoding.RAW.value
+        background_bit = HextileEncoding.BACKGROUND_SPECIFIED.value
+        foreground_bit = HextileEncoding.FOREGROUND_SPECIFIED.value
+        subrects_bit = HextileEncoding.ANY_SUBRECTS.value
+        coloured_bit = HextileEncoding.SUBRECTS_COLORED.value
         # Both carry over from the previous tile, but not across a raw tile,
```
sibson: both what?

vncdotool/decoders/hextile.py
sibson: 🤖 Claude Code: Cut to one line in f15c474 — `# IntFlag's & allocates a new flag object per test.` The rest was arguing for the change, which the commit body already does; what remains is the stdlib behaviour that would otherwise make someone put the enum back.

vncdotool/decoders/hextile.py
sibson: 🤖 Claude Code: Background and foreground — it was a forward reference to the assignment on the next line, and hoisting the bit constants above it pushed that line further away, so it now reads as trailing off my block. Named both variables in f15c474 so it stands alone.

vncdotool/decoders/hextile.py
sibson: 🤖 Claude Code: Gone in 5c3ed02. You are right that trimming it did not change what it was — it was still arguing for the change at any length, and that belongs in the commit body.

vncdotool/decoders/hextile.py
sibson: 🤖 Claude Code: Removed in 5c3ed02 as well.

PR #470, opened 2026-09-06
Skip the OS-native server tests off CI, fail them on it

sibson:
The OS-native server tests (Screen Sharing, UltraVNC, QEMU) are set up by `os-servers.yml` and nothing else, so they failed on every local `make test-func`. They now skip off CI and fail on it, keyed off `CI`/`GITHUB_ACTIONS` rather than a flag a developer has to remember. The exception is scoped to those servers alone — `skip_when_down` is set only by the `os_server()` factory — so a stopped Docker fleet still fails everywhere.

Not obvious from the diff: a runner setting neither variable never reaches the decision, because the workflow's `wait_for_servers.py` step already exits non-zero when the server isn't serving.

Beyond CI: `make test-func` with the fleet up gives `Ran 72 ... OK (skipped=4)`, and the native module under `CI=true` gives 4 failures.

### DIFF
--- a/tests/functional/utils.py
+++ b/tests/functional/utils.py
@@ -25,7 +25,7 @@ import tempfile
 import time
 import zipfile
 from pathlib import Path
-from typing import Any, Dict, List, NamedTuple, Optional, Tuple
+from typing import Any, Dict, List, Mapping, NamedTuple, Optional, Tuple
 from unittest import TestCase
 
 from PIL import Image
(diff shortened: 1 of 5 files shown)


tests/functional/utils.py
```diff
@@ -174,6 +177,19 @@ def select_servers(group: str) -> List[VNCServer]:
     return groups[group]
 
 
+CI_ENV_VARS = ("CI", "GITHUB_ACTIONS")
+NOT_CI_VALUES = ("", "0", "false", "no")
```
sibson, line 181: can't we just bool() cast rather than have constants?

tests/unit/test_native_server_gating.py
```diff
@@ -0,0 +1,124 @@
+"""tests.functional.utils is imported for the skip/fail rule alone: it needs
```
sibson, line 1: explainy

Issue #66, opened 2016-04-14
use ClientFence to syncronize state with Server


sibson:
Perhaps a new fence/block/sync command would help solve some of the issues around screen captures and mouse clicks.

https://github.com/sibson/vncdotool/blob/master/docs/rfbproto.rst#clientfence

sibson:
🤖 Claude Code:

Half of the groundwork for this lands in #471: the client answers a `ServerFence` instead of ending the session on an unknown message, and `clientFence()` exists to send one.

Deliberately left short of this issue. `PSEUDO_FENCE` is **not** offered — `VNCDoToolFactory.fence` defaults to False — because nothing yet initiates a fence or waits on one, and offering an encoding the client cannot act on is a claim it should not make. A server therefore never sends us a fence unless it does so unprompted.

What is left here:

- a `--fence` flag, and a `fence` command for the CLI, giving the encoding a caller;
- flipping `fence = True` once that exists;
- blocking on the response, which is the actual value for this issue: a fence with `BlockBefore` after a click, or `SyncNext` before a capture, so a screenshot is known to reflect the events sent before it.

Worth knowing from the measurement in #471: TigerVNC fences a client per framebuffer update only when ContinuousUpdates is on. Without it, a fence-offering session gets three fences at connect and nothing after — so the round trips only start costing anything once ContinuousUpdates lands alongside, which is the same Phase 3 item.

PR #471, opened 2026-09-06
Add Fence extension support

### LINKED ISSUE #322: vncdotool silently disconnects –  hangs on shared tigervnc

sibson:
Adds the Fence extension: the client offers `PSEUDO_FENCE` and answers a server's synchronisation request; `vnclog`'s proxy learns the message too. Supersedes #323, whose implementation this is based on and whose review comments it addresses.

Offering the encoding is what puts fences on the wire — TigerVNC leads every session with a `ServerFence`, solo or shared — so the advertisement and the handler have to land together. With only the first, `vncdo` exits 20 on `unknown message received <MsgS2C.SERVER_FENCE: 248>`.

This does not close #322. Probing TigerVNC 1.12.0 and 1.15.0 at the socket level, neither fences a client that did not offer the encoding, and one client's fencing never reaches another on a shared session, so that report did not reproduce. Asked there for a capture.


vncdotool/client.py
```diff
@@ -544,6 +546,7 @@ class VNCDoToolFactory(rfb.RFBFactory):
     pseudodesktop = True
     qemu_extended_key = True
     last_rect = True
+    fence = True
```
sibson: fence should default false?

vncdotool/rfb.py
```diff
@@ -445,6 +516,26 @@ def _handleServerCutTextValue(self, block: bytes) -> None:
         self.copy_text(block.decode("iso-8859-1"))
         self.expect(self._handleConnection, 1)
 
+    def _handleServerFence(self, block: bytes) -> None:
```
sibson, line 519: should this be build as a decoder rather than inline?

vncdotool/client.py
sibson: 🤖 Claude Code:

Measured it rather than guessing. Driving a session against the fleet's TigerVNC with and without the encoding offered:

```
no fence:   196707 bytes in,  0 fences in,  0 fences out
   fence:   196737 bytes in,  3 fences in (30 bytes),  3 fences out (30 bytes)
```

Three fences at session start, then nothing — 60 bytes once per connection, no per-update cost. TigerVNC only fences around each update when the client has enabled ContinuousUpdates, which we do not.

So the argument for `False` is not overhead, it is that nothing in vncdotool consumes a fence yet: we answer them and discard them. Offering the encoding today buys only the traffic above, while making every TigerVNC session exercise a new code path — if `_doServerFence` is wrong, it is wrong for every user rather than none. `VNCLoggingServerFactory` already sets `fence = False`, so `True` here is also the one place the two factories disagree.

The argument for `True` is that the other four pseudo-encoding flags all default on, and that `False` makes the end-to-end path untestable through the CLI: `vncdo` has no way to turn it on, so the two functional tests that drive the real client would have to go, leaving only unit tests and the socket-level ones. Adding a `--fence` flag to keep them is more surface for a feature with no consumer.

I lean `True` on that last point alone — the measured cost is 60 bytes and the alternative is a handler no real session ever runs. But it is a close call and the conservative default is defensible; happy to flip it and drop the two client-level tests if you would rather nothing changes for existing users until #66 gives fences a consumer.

vncdotool/rfb.py
sibson: 🤖 Claude Code:

It cannot be, as the framework stands. A decoder is a handler for a *rectangle encoding*: `_doConnection` reaches one only from inside the rectangle loop of a FramebufferUpdate, keyed on the rectangle's own encoding field (rfb.py:432), and `ControlDecoder.decodeForControl` takes `(client, width, height)` and consumes no bytes because its rectangle carries none.

Fence has no rectangle. `ServerFence` is a top-level server-to-client message with its own id, 248, and a variable-length payload; it arrives between updates, not inside one. `PSEUDO_FENCE` is only the token that tells the server the client can receive that message — unlike `PSEUDO_DESKTOP_SIZE` or `PSEUDO_QEMU_EXTENDED_KEY_EVENT`, it never appears as a rectangle, so there is nothing for a decoder to be dispatched on.

Hosting it would mean extending the framework to top-level messages, and fence would be its only member — `SERVER_CUT_TEXT` is the nearest existing case and stays inline too. The three methods here follow that one deliberately: `_handleServerCutText` / `_handleServerCutTextValue` is the same split, a fixed header that reads a length, then the variable part.

Worth revisiting if ContinuousUpdates and `EndOfContinuousUpdates` land for #66, since that would make three top-level messages in the same area rather than one.

vncdotool/client.py
sibson: 🤖 Claude Code: Flipped to `False` in e371192, and you are right that it is the honest default — the client answers fences, it does not support them.

Dropped the capture test that depended on the old default; `vncdo` no longer sees a fence at all, so it was proving nothing. The handler stays covered by the unit tests in `test_rfb.py`, and a new `test_fence_is_not_offered_by_default` drives the real factory, since `test_vncConnectionMade`'s factory is a Mock and reads True for every flag.

`VNCDoToolFactory.fence = True` still opts in. Recorded the rest — `--fence` flag, `fence` command, blocking on the response, then flipping the default — on #66.

vncdotool/rfb.py
sibson: 🤖 Claude Code: Queued as #474 — decompose the top-level message handlers out of `rfb.py` the way rectangle encodings went into `decoders/`.

Left inline here. `SERVER_FENCE` is the third instance of the same header-then-payload shape (`SET_COLOUR_MAP_ENTRIES` and `SERVER_CUT_TEXT` are the others), and the fourth and fifth arrive with ContinuousUpdates and Extended Clipboard; the issue notes the open questions, including whether `loggingproxy.py`'s `TYPE_LEN` table comes along, since it has the same shape in the other direction and already drifted once.

PR #473, opened 2026-09-06
Fail an off-screen expect or capture region instead of polling to the timeout

sibson:
`rexpect`/`rcapture` accepted an X,Y putting the region past the edge of the screen. `Image.crop` pads the outside with black rather than failing, so the padded crop is the right size and never matches: `rexpect` polled until `--timeout`, exit 40, with nothing naming the coordinates. A guard now fails with the region and the screen size, exit 30.

Two things the diff does not say. The check sits in `_expectCompare`, re-entered on every poll, so it also ends a wait that a `PSEUDO_DESKTOP_SIZE` resize shrinks the region off — a judgement call worth disagreeing with. And it reaches `expect`, not just `rexpect`: a target image larger than the screen was the same silent timeout, which is why `MSG_INIT` moves from 0x0 to 800x600.

`make test-func` passes against the Docker fleet; CI does not run it.


sibson:
🤖 Claude Code: the `usage.rst` paragraph this PR added is dropped in #476, which fills in `commands.rst` — the Command Reference was fifteen bare headings, which is why per-command detail like this kept landing in the narrative guide instead. The off-screen failure is now stated on each of the `rcapture`, `rexpect` and `expect` entries, where someone looking up an argument will find it; `usage.rst` keeps a one-line pointer to that page where the paragraph stood.

Nothing here is lost, and nothing in this PR needs changing — #476 resolves the merge that way.

PR #480, opened 2026-09-06
Sweep comments and docstrings across the repo

sibson:
Full-repo comment/docstring sweep against the `writing-code-comments` skill's Mode B: deletes comments restating code, naming callers/issues, or duplicating `specs/*.md` and `CLAUDE.md`; rewrites the rest to the fact they carry. Comments older than 6 months got more benefit of the doubt but weren't exempt.

Two stale/wrong comments caught along the way:
- `copyrect.py`'s bounds-check comment had the raise-vs-zero-fill behavior backwards.
- `decoders/__init__.py` misattributed why Cursor/DesktopSize/QemuExtendedKey are absent from `ENCODING_NAMES` (they're offered via `client.py`'s pseudo-* flags, not `--encodings`).

`docs/conf.py`'s ~60 lines of Sphinx-quickstart template prose are gone. Excluded `vncdotool/rfb.py` (vendored) and CI workflow files (audited, all comments sourced and current).

An independent subagent re-reviewed the staged diff before commit and caught one over-deletion, restored before landing.

617 unit tests pass; flake8 clean.


docs/conf.py
```diff
@@ -1,256 +1,44 @@
-# -*- coding: utf-8 -*-
```
sibson, line 1: this is a utf tag, not sure we should remove

Issue #474, opened 2026-09-06
Decompose top-level message handlers out of rfb.py, as rectangle encodings were


sibson:
Rectangle encodings moved out of `rfb.py` into `vncdotool/decoders/` (see `specs/decoder-architecture.md`): a class per encoding, registered in `DECODERS`, dispatched from the rectangle loop on the rectangle's encoding field. Top-level server-to-client messages never got the same treatment. They are still a chain of `elif`s in `_handleConnection`, each with its handler methods inline in `RFBClient`:

| Message | Handlers in `rfb.py` |
|---|---|
| `FRAMEBUFFER_UPDATE` | `_handleFramebufferUpdate`, `_handleRectangle`, `_handleDecodeRect`, … |
| `SET_COLOUR_MAP_ENTRIES` | `_handleServerColourMap`, `_handleServerColourMapValue` |
| `BELL` | inline |
| `SERVER_CUT_TEXT` | `_handleServerCutText`, `_handleServerCutTextValue` |
| `SERVER_FENCE` | `_handleServerFence`, `_handleServerFencePayload`, `_doServerFence` |

The shape repeats: a fixed header that reads a length, then a variable part, then back to `_handleConnection`. Three of the five are already that. Extending `RFBClient` for a new message means editing the `elif` chain and adding methods to a class that is already the largest in the project.

Adding ContinuousUpdates for #66 adds `EndOfContinuousUpdates`, and Extended Clipboard adds another. That is when the pattern is worth naming.

Worth deciding as part of this:

- whether the existing decoder base classes extend to top-level messages or whether this is a separate registry — a decoder is handed `(client, width, height)` or a pixel buffer, neither of which a fence has;
- what a handler that consumes a variable-length payload looks like, given `expect()` is how the current ones stage a second read;
- whether `loggingproxy.py`'s `TYPE_LEN` table and its own `elif` chain over client-to-server messages come along, since they have the same shape in the other direction and already drifted once (a `CLIENT_FENCE` missing from `TYPE_LEN` desyncs every message after it).

Not urgent. `SERVER_FENCE` landing inline in #471 is the third instance of the pattern, not a problem in itself; this is a note that the fourth and fifth are coming.

PR #485, opened 2026-09-07
Decompose non-rectangle message handlers out of rfb.py

sibson:
Phase 1 of #474: moves SetColourMapEntries, ServerCutText, Bell and ServerFence out of `rfb.py` into `vncdotool/messages/`, one handler class per message, mirroring the `DECODERS` registry from `specs/decoder-architecture.md`. Each handler is a generator, unit-testable with no reactor or `RFBClient`. `FRAMEBUFFER_UPDATE`/the rectangle loop are untouched (deferred, per the design doc's phasing).

Real bug fix included: `SetColourMapEntries`' colour count and `ServerCutText`'s length were server-declared and unbounded, so a malformed server could make the client buffer arbitrarily much data. `requirePayload()` now bounds both at 1 MiB and raises the existing `DecodeError`.

Tests: 629 unit tests pass, flake8 clean. New `tests/unit/test_messages.py` plus an integration test for the oversized-payload disconnect.

### DIFF
--- a/vncdotool/messages/fence.py
+++ b/vncdotool/messages/fence.py
@@ -11,12 +11,10 @@ class ServerFenceHandler(MessageHandler):
     MESSAGE = MsgS2C.SERVER_FENCE
 
     def handle(self, client: Any) -> Generator[int, bytes, None]:
-        # rfbproto ServerFence: 3 bytes padding, U32 flags, U8 payload-length.
         flags, length = unpack("!xxxIB", (yield 8))
         payload = (yield length) if length else b""
         if flags & FenceFlags.REQUEST:
-            # rfbproto ServerFence: masking to the flags handled here, rather
-            # than just clearing Request, is what lets the server tell which
-            # flags this client supports as new ones are defined.
+            # Masking, not clearing REQUEST alone, is what lets the server
+            # learn which flags this client understands.
             known = FenceFlags.BLOCK_BEFORE | FenceFlags.BLOCK_AFTER | FenceFlags.SYNC_NEXT
             client.clientFence(FenceFlags(flags) & known, payload)
--- a/vncdotool/messages/fence.py
+++ b/vncdotool/messages/fence.py
@@ -1,0 +1,22 @@
+from __future__ import annotations
+
+from struct import unpack
+from typing import Any, Generator
+
+from ..const import FenceFlags, MsgS2C
+from .base import MessageHandler
+
+
+class ServerFenceHandler(MessageHandler):
+    MESSAGE = MsgS2C.SERVER_FENCE
+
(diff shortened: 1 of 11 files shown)


vncdotool/messages/fence.py
```diff
@@ -0,0 +1,22 @@
...
+
+class ServerFenceHandler(MessageHandler):
+    MESSAGE = MsgS2C.SERVER_FENCE
+
+    def handle(self, client: Any) -> Generator[int, bytes, None]:
+        # rfbproto ServerFence: 3 bytes padding, U32 flags, U8 payload-length.
+        flags, length = unpack("!xxxIB", (yield 8))
+        payload = (yield length) if length else b""
+        if flags & FenceFlags.REQUEST:
+            # rfbproto ServerFence: masking to the flags handled here, rather
```
sibson: documenting the protocol docs?

vncdotool/rfb.py
```diff
@@ -78,6 +77,11 @@ class RFBClient(Protocol):
     # narrows it.
     MAX_DESKTOP_SIZE = 0x10000
 
+    # Bounds a server-declared message payload length (ServerCutText,
```
sibson: it's a max why do we decribe it as a max

vncdotool/rfb.py
```diff
@@ -312,17 +317,21 @@ def _handleConnection(self, block: bytes) -> None:
...
-        elif msgid == MsgS2C.BELL:
-            self.bell()
-            self.expect(self._handleConnection, 1)
-        elif msgid == MsgS2C.SERVER_CUT_TEXT:
-            self.expect(self._handleServerCutText, 7)
-        elif msgid == MsgS2C.SERVER_FENCE:
-            self.expect(self._handleServerFence, 8)
-        else:
+            return
+        handler_cls = messages.HANDLERS.get(msgid)
```
sibson: why get the class then get the handler later?

vncdotool/messages/fence.py
sibson: 🤖 Claude Code: Cut it — trimmed to a one-line rationale for the masking choice itself (fa75c99).

vncdotool/rfb.py
sibson: 🤖 Claude Code: Cut it — trimmed to the one non-obvious fact (policy not protocol, subclass-narrowable), matching MAX_DESKTOP_SIZE's comment style (fa75c99).

vncdotool/rfb.py
sibson: 🤖 Claude Code: Fixed — one lookup now. self._messages is keyed identically to HANDLERS (built from it in for_connection), so _handleConnection does self._messages.get(msgid) once and uses the same result for the None-check and the call (fa75c99).

PR #486, opened 2026-09-07
Fix vnclog hang, dropped QEMU events, and off-by-one in C2S parser

sibson:
Fixes three parsing bugs in vnclog's C2S message decoder (`loggingproxy.py`), independent of the larger redesign in #474.

- `CLIENT_CUT_TEXT` had no `TYPE_LEN` entry, so it consumed 0 bytes and spun `_handle_protocol` forever on the same buffer -- a clipboard paste through `vnclog` wedged the session.
- `QEMU_CLIENT_MESSAGE` was listed as 1 byte instead of 2, starving `unpack` and silently dropping every extended key event.
- The short-read path staged `nbytes + 1` instead of `nbytes`, stalling a complete message on one extra byte -- an unexplained off-by-one from 2011.

Added a structural test asserting every `MsgC2S` member is sized in `TYPE_LEN` or explicitly documented as unsupported.

Tested: `make test` (658 passed), `flake8` clean.

### DIFF
--- a/tests/unit/test_loggingproxy.py
+++ b/tests/unit/test_loggingproxy.py
@@ -8,12 +8,14 @@ from __future__ import annotations
 
 import dataclasses
 import logging
+import unittest
 from struct import pack
 from unittest import TestCase, mock
 
 from vncdotool import pixelformat
-from vncdotool.const import AuthTypes, Encoding, MsgC2S, MsgS2C
+from vncdotool.const import AuthTypes, Encoding, MsgC2S, MsgS2C, QemuClientMessage
 from vncdotool.loggingproxy import (
+    TYPE_LEN,
     VNCLoggingClientProxy,
     VNCLoggingServerFactory,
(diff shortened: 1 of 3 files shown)


tests/unit/test_loggingproxy.py
```diff
@@ -155,3 +157,120 @@ def test_an_unreadable_server_message_is_reported_not_raised(self) -> None:
...
+        message = pack("!BBxxI", MsgC2S.KEY_EVENT, 1, 65)  # 8 bytes total
+
+        self.server_proxy.dataReceived(message[:7])
+        self.server_proxy.handle_keyEvent.assert_not_called()
+        self.server_proxy.dataReceived(message[7:])
+
+        self.server_proxy.handle_keyEvent.assert_called_once_with(65, 1)
+
+
+# Messages the proxy deliberately does not decode: they fall to the `else:
```
sibson, line 217: veerbose

PR #487, opened 2026-09-07
Fix ZRLE decoding for misreported pixel depths

### LINKED ISSUE #483: ZRLE misdecodes against a server that reports depth=32 (libvncserver-example)
`cpixel_bytes()` (`vncdotool/pixelformat.py:177-186`) gates the ZRLE/Tight
3-byte CPIXEL narrowing strictly on the server's declared `depth <= 24`.
`libvncserver-example` declares `depth=32` in its `PixelFormat` -- already a
documented quirk (`specs/pixel-format.md:52-54`, quoting rfbproto: some
servers "will send a *depth* identical to *bits-per-pixel* for historical
reasons"). `raw_mode()` already ignores `depth` for exactly this reason
(`specs/pixel-format.md:118`), but `cpixel_bytes()` does not.

Because the server's declared depth is bogus, `cpixel_bytes()` computes
CPIXEL as 4 bytes, but the server's actual ZRLE encoder narrows to 3 bytes
regardless of what it declared -- so the client misparses the stream.

Reproduced against the live container (`docker compose -f
tests/servers/docker-compose.yml up -d libvncserver-example`, port 5935):

```
$ vncdo -s 127.0.0.1::5935 -v capture probe.png
...
INFO: Native PixelFormat(bpp=32, depth=32, ... redmax=255 greenmax=255 bluemax=255 redshift=0 greenshift=8 blueshift=16)
INFO: cannot decode this rectangle: ZRLE RLE tile at (0,0) decoded 4267 pixels, wanted 4096
```

Forcing a pixel format with `depth<=24` works:

```
$ vncdo -s 127.0.0.1::5935 --pixel-format rgbx8888 capture probe.png   # succeeds
$ vncdo -s 127.0.0.1::5935 --encodings raw capture probe.png          # succeeds (no ZRLE offered)
```

This never fired before because `vncdo` only ever offered Raw until #462
started offering ZRLE (and Tight) by default; #462 works around it by
dropping ZRLE from the default list, since Tight already dominates ZRLE on
every measured scene (`specs/default-encodings.md`) and this particular
server answers Tight with Raw anyway. `--encodings zrle` explicitly still
hits the bug against this server.

The real fix is in `cpixel_bytes()`/`cpixel_offset()`: determine 3-byte
CPIXEL eligibility from whether the color channels actually fit within
3 bytes at the negotiated shifts (as `raw_mode()` already does), rather than
trusting the declared `depth` field -- mirroring the historical-depth
leniency `specs/pixel-format.md` already documents for `raw_mode`. Needs a
protocol-doc check (RFC 6143 §7.7.5, rfbproto) before touching it, since
loosening the `depth <= 24` gate risks misreading a server where depth is
correctly reported.

kudala-bharani:
Fixes #483: ZRLE captures accept a depth-32 declaration when the actual color channels fit three bytes.

This follows LibVNCServer's encoder; genuine deep-color formats still need four bytes. The five regression cases fail on the original code.

The advisory type checker reports the same 74 errors before and after the fix. Live-server tests and prose lint were not run locally: Docker is stopped and Vale is unavailable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## 2.0.0.dev0 (UNRELEASED)
+- Fix ZRLE captures from servers that declare depth 32 for 24-bit color (@kudala-bharani, #483)
 - Fix `updateCursor` decoding a hide-pointer update (width or height 0) as an empty image instead of hiding the cursor, which raised inside Pillow or pasted a bogus zero-size image onto the screen (@sibson, #449)
 - Fix `RFBFactory` raising `AttributeError` on ARD authentication when used directly, instead of through `VNCDoToolFactory` (@sibson)
 - Add `vncdo stable SECONDS [FUZZ]` and `rstable SECONDS X Y W H [FUZZ]`, waiting until the screen stops changing rather than until it matches a reference image (@sibson)
--- a/specs/pixel-format.md
+++ b/specs/pixel-format.md
@@ -116,7 +116,6 @@ def channel_tolerance(pf: PixelFormat) -> tuple[int, int, int]:  # 8-bit error p
 ```
 
 `raw_mode` ignores `depth`: it does not affect where the channels sit.
-`cpixel_bytes` must obey it, since the encoder used the same declaration.
 
 `UnsupportedPixelFormat` rather than a guess, though under the policy below a
 caller only reaches it when a server ignores what was requested.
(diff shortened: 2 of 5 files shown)


kudala-bharani:
Resolved the changelog conflict against current `main`, retaining all upstream entries and the ZRLE fix. All 672 unit tests and flake8 pass; the five regression cases still fail when the fix is removed. The advisory type checker still reports the same 74 errors with and without the fix. Live-server tests and prose lint remain unrun locally because Docker is stopped and Vale is unavailable.

sibson:
🤖 Claude Code: Closing this in favor of #488, which Marc merged directly -- same root-cause fix (dropping the `depth <= 24` gate from `cpixel_bytes()`), plus a leftover-bytes guard in `ZRLEDecoder` and multi-server verification this PR didn't have. Recent refactoring (#474, #485/#486) also moved code this diff touches, so it's out of date against `main` regardless. Thanks for the fix, @kudala-bharani -- it correctly diagnosed the root cause and #488 built on that.

PR #488, opened 2026-09-07
Fix ZRLE CPIXEL narrowing to ignore a lying depth field (#483)

### LINKED ISSUE #483: ZRLE misdecodes against a server that reports depth=32 (libvncserver-example)

sibson:
## Summary

- Fixes #483: `cpixel_bytes()` gated ZRLE's 3-byte CPIXEL narrowing on `depth <= 24`, desyncing the decoder against servers (libvncserver-example) that report `depth=32` but narrow anyway. Depth is now ignored, matching `raw_mode()` — channel placement alone decides eligibility.
- Guards that assumption: ZRLE now ends the session with a named error if a rectangle's tiles leave bytes unconsumed, rather than misreading a genuinely 4-byte server as the next tile's garbage.
- Adds `tests/goldens/probe_pixel_format.py`, wired into CI as a non-blocking sweep, to build a record of real servers' depth behaviour.

## Test plan

- `make test`: 623 unit tests pass, `flake8` clean.
- New unit tests pin both fixes with synthetic bytes; every committed ZRLE golden still decodes clean under the leftover-bytes check.
- Manually probed every docker-fleet truecolor server at depth=32, including libvncserver-example's native ServerInit (previously the reported crash) — all decode clean. Notes in `specs/cpixel-depth-quirk.md`.


vncdotool/decoders/zrle.py
```diff
@@ -198,3 +198,13 @@ def short(tx: int, ty: int) -> DecodeError:
                             )
                         rows.append(b"".join(map(expanded.__getitem__, packed))[:stride])
                     target.blit(tx, ty, tw, th, b"".join(rows))
+
+        if pos != end:
+            # Every subencoding consumes exactly the bytes RFC 6143 7.7.6
```
sibson, line 203: verbose

PR #492, opened 2026-09-07
Collapse CHANGELOG entries for #483/#474 into net facts

sibson:
Collapses the 5 CHANGELOG bullets #483 and #474 landed into 3, per the writing-code-comments skill's Mode C cut test: the ZRLE fix and its leftover-bytes guard read as one fact, the `rfb.py`→`vncdotool/messages/` move is refactor mechanism rather than something a user would notice, and `vnclog`'s two C2S table bugs are one symptom (it eats certain client messages).


sibson:
🤖 Claude Code: Closing -- folding this into #490, which is becoming the single cleanup PR for the whole 2.0.0.dev0 section per Marc's request.

PR #497, opened 2026-09-07
Add vncdo resize W H, resizing the desktop through SetDesktopSize

### LINKED ISSUE #301: Feature request: support for SetDesktopSize
Newer vnc servers like TigerVNC support automatically resizing the vnc framebuffer when the client's window is resized. I think this is accomplished through SetDesktopSize:
https://vncdotool.readthedocs.io/en/0.8.0/rfbproto.html#setdesktopsize

Would it be possible for vncdotool to support resizing the desktop via this mechanism?

sibson:
Closes #301. Adds `vncdo resize W H`, which asks the server for a new desktop through `SetDesktopSize` and reports a refusal instead of exiting green.

What a reviewer would not guess from the diff: offering `ExtendedDesktopSize` changes what existing sessions receive, which deadlocked `capture` against libvncserver until `commitUpdate` learned to re-request incrementally — see the commit body for the loop rfbproto forbids.

Tested beyond CI: the wire shapes were read off TigerVNC 1.12 on a raw socket, not off this implementation, including the refusal path (reason 1, result 3). A new `tigervnc-resize` service on 5940 gives the functional test a server whose geometry nothing else measures.


tests/servers/docker-compose.yml
```diff
@@ -32,6 +32,19 @@ services:
     ports:
       - "127.0.0.1:5932:5900"
 
+  # A resize test changes the framebuffer geometry out from under itself, so
```
sibson, line 35: why not change it back when done rather than another server?

vncdotool/decoders/__init__.py
```diff
@@ -41,15 +46,16 @@
...
 }
 
-# The names --encodings accepts. Cursor, PointerPos, DesktopSize and
-# QemuExtendedKey are registered decoders too, but client.py offers them
-# from its own cursor / pseudodesktop / qemu_extended_key flags rather than
-# by name here.
+# The names --encodings accepts. Cursor, PointerPos, DesktopSize,
+# ExtendedDesktopSize and QemuExtendedKey are registered decoders too, but
+# client.py offers them from its own cursor / pseudodesktop /
+# qemu_extended_key flags rather than by name here.
```
sibson: restates code

vncdotool/decoders/control.py
```diff
@@ -1,10 +1,15 @@
-"""DesktopSize, PointerPos and QEMU extended key pseudo-encodings. rfbproto."""
+"""DesktopSize, ExtendedDesktopSize, PointerPos and QEMU extended key
```
sibson: let's not enumerate all the decoders, it's extra churn

tests/servers/docker-compose.yml
sibson: Considered it, but two things pushed toward a dedicated server instead:

- The test never lands back on 256x192 mid-run — it asserts grown (320x240) then shrunk (288x216) — so "reset" would mean one more `vncdo resize 256 192` call, over the same SetDesktopSize path the test exists to exercise. If that path has a bug, the cleanup is exactly as broken as the thing under test.
- `addCleanup`/`tearDown` only fire if the process unwinds normally. A CI timeout, a hung subprocess, or the reset call itself failing would leave the shared `tigervnc`/`tigervnc-auth` server at the wrong size for the rest of the run — and everything else against TigerVNC, including the decoder goldens, is a pixel-exact match cut to 256x192. That's a much wider blast radius than one failing test.

A server nothing else reads geometry from means a resize-test failure stays contained to itself. Happy to switch to reset-in-place if you'd rather keep the fleet smaller and accept that tradeoff — let me know.

---

vncdotool/client.py
```diff
@@ -551,6 +565,89 @@ def mouseDrag(self: TClient, x: int, y: int, step: int = 1) -> Iterator[Deferred
 
         returnValue(self)
 
+    def resize(self, width: int, height: int) -> Deferred:
```
sibson: requestDesktopResize()?

vncdotool/decoders/__init__.py
sibson: Fair, dropped the flag-name enumeration — just says these are offered elsewhere now.

---

vncdotool/decoders/control.py
sibson: Agreed, switched to naming the category instead of the members.

---

vncdotool/client.py
sibson: Done — renamed `resize` to `requestDesktopResize` and updated its call sites in `command.py` and the unit tests. The CLI verb (`vncdo resize W H`) is unchanged.

---

sibson:
`Windows - UltraVNC, TightVNC, TigerVNC` failed on the last two pushes: `AssertionError: False is not true : tightvnc: capture is a single flat colour, no screen content was decoded` ([run](https://github.com/sibson/vncdotool/actions/runs/35296312435/job/105449453549)).

Not this PR's: the assertion is in the shared `test_capture` smoke check (`tests/functional/utils.py:881`) against TightVNC on the Windows runner, and nothing in this diff touches TightVNC, its encodings, or that test — it passed on every earlier commit here. Reads as a screen-content timing flake on the Windows GUI server, not a regression from the resize/comment changes.

I don't have permission to re-run the job (`rerun-failed-jobs` 403s for this integration) — flagging rather than re-running blind. Will keep watching; a repeat failure with a different cause would change this.

---

vncdotool/client.py
sibson: Follow-up: `requestDesktopResize` didn't fit either — no other method uses a `request*` prefix. Renamed again to `resizeScreen`, matching the `<verb>Screen` pattern already used by `captureScreen`/`refreshScreen`/`renderScreen`/`expectScreen`/`stableScreen`. `self.screen` is what a desktop resize actually mutates (see `updateDesktopSize`), so it's also accurate, not just consistent.

---

tests/servers/docker-compose.yml
```diff
@@ -32,6 +32,19 @@ services:
...
+  # it needs a server no other test reads a size from.
+  tigervnc-resize:
```
sibson, line 37: Rather than start a new server, why can't we use the down down down to set to a known size before we start?

vncdotool/client.py
```diff
@@ -210,8 +214,16 @@ class VNCDoToolClient(rfb.RFBClient):
     cursor: Image.Image | None = None
     cmask: Image.Image | None = None
 
+    screens: tuple[rfb.Screen, ...] = ()
+    _resize: Deferred | None = None
```
sibson: this seems like a lot of members for the resize,  can we not attach them to an object?

vncdotool/client.py
```diff
@@ -231,6 +243,8 @@ def connectionMade(self) -> None:
 
     def connectionLost(self, reason: Failure = connectionDone) -> None:
         super().connectionLost(reason)
+        if self._resize is not None:
+            self._resizeFailed("connection lost before the server answered")
```
sibson, line 255: why do we need this as a special handling rather than just using existing deferred?

vncdotool/client.py
```diff
@@ -553,6 +567,89 @@ def mouseDrag(self: TClient, x: int, y: int, step: int = 1) -> Iterator[Deferred
 
         returnValue(self)
 
+    def resizeScreen(self, width: int, height: int) -> Deferred:
+        """Ask the server to change its desktop to `width` x `height`.
+
+        The result fires once the server has said whether it did, and fails
```
sibson: that is our standard pattern, why docstring to describe it?

vncdotool/client.py
```diff
@@ -553,6 +567,89 @@ def mouseDrag(self: TClient, x: int, y: int, step: int = 1) -> Iterator[Deferred
...
+        with :class:`DesktopResizeError` if it refused or never offered the
+        ExtendedDesktopSize pseudo-encoding at all.
+        """
+        log.debug("resize %dx%d", width, height)
+        if not (
```
sibson: not () seems awkward to read here/. Can it be simpler?

vncdotool/client.py
```diff
@@ -553,6 +567,89 @@ def mouseDrag(self: TClient, x: int, y: int, step: int = 1) -> Iterator[Deferred
...
+            0 < width < self.MAX_DESKTOP_SIZE and 0 < height < self.MAX_DESKTOP_SIZE
+        ):
+            raise ValueError((width, height))
+        if self._resize is not None:
```
sibson, line 583: why can't we issue a second resize?

vncdotool/client.py
```diff
@@ -553,6 +567,89 @@ def mouseDrag(self: TClient, x: int, y: int, step: int = 1) -> Iterator[Deferred
...
+        if rfb.Encoding.PSEUDO_EXTENDED_DESKTOP_SIZE in self.negotiated_encodings:
+            self._sendResize()
+        else:
+            self.framebufferUpdateRequest(incremental=False)
+        return d
+
+    def _sendResize(self) -> None:
+        assert self._resize_target is not None
+        width, height = self._resize_target
+        self._resize_sent = True
```
sibson: it's not been sent yet, so whatever we're guarding isn't correct.  Do we really need a sent?

vncdotool/rfb.py
```diff
@@ -536,6 +547,18 @@ def clientFence(self, flags: FenceFlags, payload: bytes = b"") -> None:
         """Request, or respond to, a Fence synchronisation of the data stream."""
         self.transport.write(pack("!BxxxIB", MsgC2S.CLIENT_FENCE, flags, len(payload)) + payload)
 
+    def setDesktopSize(self, width: int, height: int, screens: Sequence[Screen]) -> None:
```
sibson, line 550: what is screens?

vncdotool/rfb.py
```diff
@@ -622,6 +645,25 @@ def updatePointerPos(self, x: int, y: int) -> None:
     def updateDesktopSize(self, width: int, height: int) -> None:
         """New desktop size of width*height."""
 
+    def updateExtendedDesktopSize(
```
sibson, line 648: should this just be collapsed to updateDesktopSize above?

vncdotool/client.py
sibson: Agreed. Bundled `_resize`, `_resize_timeout`, `_resize_target`, `_resize_sent` into one `_PendingResize` object (mirroring the existing `_FullScreenReceived` pattern in this file), so the client only carries one `_resize: _PendingResize | None` attribute. This also let `_resize_sent` go away entirely — see the reply on that thread.

---

vncdotool/client.py
sibson: You're right that setting it at the top of `_sendResize` was misleading — it's true before the method has actually written anything to the wire. Turns out it's not needed as a separate flag at all: `target` only exists to remember what to send, so clearing it once `_sendResize` has committed to sending doubles as the "already sent" signal the guard in `updateExtendedDesktopSize` needs (`self._resize.target is not None` in place of `not self._resize_sent`). Folded into the `_PendingResize` bundling from the other thread — `_resize_sent` is gone.

---
_Generated by [Claude Code](https://claude.ai/code)

---

sibson: (previous reply's footer link got cut off, ignore it)

---

vncdotool/client.py
sibson: Fair — `refreshScreen` right above it has no docstring at all, and none of the others spell out the fire/fail contract in prose. Trimmed to one line; `DesktopResizeError`'s own docstring already covers the refused-vs-silent distinction.

---

vncdotool/client.py
sibson: Rewrote as a plain OR-chain of the four rejection conditions instead of "proceed only if not (both in range)" — reads more directly as a guard clause.

Noticed `updateDesktopSize` (existing code, not touched by this PR) has the exact same `not (...)` shape for the same bound. Left it alone since it's outside this diff, but happy to match it to the new style too if you'd like — just say the word.

---

vncdotool/client.py
sibson: Checked — there isn't an existing generic mechanism to reuse. `self.deferred` (the slot `captureScreen`/`expectScreen`/`stableScreen`/`refreshScreen` all share, resolved via `commitUpdate`) isn't failed on `connectionLost` today either — if the connection drops mid-capture, that Deferred just never fires. `resizeScreen`'s docstring/contract is the first promise in this file that a Deferred won't hang forever, so it needed its own hook.

I don't want to widen this PR into also fixing `self.deferred`'s hang-on-disconnect for the older commands — that's a real gap but a separate, bigger change (four other call sites, and deciding what error they should raise). Want me to file a follow-up issue for it, or leave the resize-only fix as is?

---

vncdotool/client.py
sibson: Nothing stops us technically — it's really a question of what happens to the first caller. Two honest options:

1. **Supersede**: cancel the first `resizeScreen()`'s Deferred with an error ("superseded by a new resize") and start the second. The first caller learns their request didn't win.
2. **Piggyback**: let the second call retarget the same in-flight operation, and both Deferreds resolve together off whatever the server ultimately answers. Simpler for the caller, but two different requested sizes now share one outcome, which could surprise someone who asked for 800x600 and got told "success" when the server actually granted the other caller's 640x480.

There's no test pinning down the current reject-outright behavior, so this is a free choice, not something I'd need to justify changing. Which do you want — or is reject-and-raise actually fine and you were just probing?

---

tests/servers/docker-compose.yml
sibson: Checked whether TigerVNC's Xvnc has an equivalent to `LIBVNCSERVER_EXAMPLE`'s `normalize_size_keys` (up/up/down) — it doesn't. That trick works because libvncserver's example app is an SDL/GLUT window that responds to arrow keys by resizing itself; it's a feature of that specific demo, not something RFB or X11 give you generically. Xvnc's framebuffer size is fixed at startup; nothing in `tests/servers/Dockerfile` or the entrypoint enables `-randr`/`xrandr`, so there's no X11-level lever either.

The only way to change TigerVNC's size from outside is `SetDesktopSize` — the protocol path this PR adds and the test exercises. Using it to reset afterward is the same "cleanup is exactly as broken as the thing under test" tradeoff I raised on the sibling thread about the dedicated server. Given that, a "down down down"-style warm-up isn't available here; it's dedicated server vs. reset-via-the-thing-being-tested vs. some third option I haven't found. Still happy to switch if you'd rather accept one of those tradeoffs — which way do you want to go?

---

vncdotool/rfb.py
sibson: One `Screen` per monitor in the new layout (rfbproto's SCREEN structure — id, x, y, width, height, flags). Clarified the docstring: "laid out as `screens` -- one `Screen` per monitor in the new layout."

---

vncdotool/rfb.py
sibson: Can't, they carry different wire payloads. `DesktopSizeDecoder` (PSEUDO_DESKTOP_SIZE, -223) hands `updateDesktopSize` just `width, height` — that pseudo-encoding's rectangle header *is* the whole payload, RFC 6143 gives it nothing else. `ExtendedDesktopSizeDecoder` (-308) hands `updateExtendedDesktopSize` `reason, result, width, height, screens` — a real message body with a reason/result pair (repurposing the rectangle's x/y per rfbproto) and a screen-layout array.

`resizeScreen` depends specifically on `result` (to tell a granted resize from a refused one) and `reason` (to tell "this is the answer to my own request" from "the server or another client changed something"). Neither exists in the -223 message, so there's no way to fold the two callbacks into one signature without either inventing values for a message that doesn't carry them, or silently dropping the resize outcome.

---

sibson:
`Windows - UltraVNC, TightVNC, TigerVNC` failed again on this head, but a different failure than the earlier flake: Chocolatey couldn't install UltraVNC — `winvnc.exe not found after 3 attempts to install ultravnc` ([run](https://github.com/sibson/vncdotool/actions/runs/35666864246/job/106554451192)) — before any test ran.

Not this PR's: nothing in this diff touches CI config, the Windows server setup scripts, or package installation. A package-manager install flake on the runner.

Same as before, I don't have permission to re-run the job (403). Flagging rather than guessing; will keep watching.

---

Issue #310, opened 2026-03-13
Can't connect to raspios - "unknown security types"


MrButlerTron:
Anyone have suggesstions - trying to use vncdo from my Mac to a Raspberry Pi and I'm getting "unknown security types"

```
INFO:twisted:Starting factory <vncdotool.command.VNCDoCLIFactory object at 0x107342cf0>
INFO:twisted:Using protocol version 3.8
INFO:twisted:Offered <AuthTypes.VENCRYPT: 19>
INFO:twisted:Offered <AuthTypes.RSA_AES256: 129>
INFO:twisted:Offered <AuthTypes.RSA_AES: 5>
INFO:twisted:unknown security types: (19, 129, 5)
INFO:twisted:Stopping factory <vncdotool.command.VNCDoCLIFactory object at 0x107342cf0>
INFO:twisted:Main loop terminated.
```

command:
`vncdo -v -s hosname-u pi -p 'raspberry' capture screenshot.png`

lurch:
The latest version of [Raspberry Pi OS](https://www.raspberrypi.com/software/operating-systems/) is `PRETTY_NAME="Debian GNU/Linux 13 (trixie)"` and the `wayvnc` version is now:
```
$ wayvnc --version
wayvnc: 0.9.1
neatvnc: 0.9.5
aml: 0.3.0
```

But FWIW, I also get the same error when trying to connect to that with `vncdo 1.3.0`:
```
$ vncdo -vv -s my_pi_hostname -u my_pi_username -p my_pi_password type hello
INFO:root:connecting to my_pi_hostname:5900
INFO:twisted:Starting factory <vncdotool.command.VNCDoCLIFactory object at 0x72b4871c5c00>
INFO:twisted:Using protocol version 3.8
INFO:twisted:Offered <AuthTypes.VENCRYPT: 19>
INFO:twisted:Offered <AuthTypes.RSA_AES256: 129>
INFO:twisted:Offered <AuthTypes.RSA_AES: 5>
INFO:twisted:unknown security types: (19, 129, 5)
INFO:twisted:Stopping factory <vncdotool.command.VNCDoCLIFactory object at 0x72b4871c5c00>
INFO:twisted:Main loop terminated.
```

TigerVnc is able to connect to the same wayvnc server just fine.

lurch:
...and in case it's helpful, here's the default `/etc/wayvnc/config` file on Raspberry Pi OS:
```
use_relative_paths=true
address=::
enable_auth=true
[...]

Issue #138, opened 2018-12-29
How to use with Proxmox


Richie765:
I'm using Proxmox 5.2-1 and want to login some VM's automatically over the UNIX socket. I can't get the authentication to work.

```
# vncdotool --version
vncdotool 0.12.1

# vncdotool -p password -v -s /var/run/qemu-server/123.vnc capture /tmp/test.png
INFO:root:connecting to /var/run/qemu-server/123.vnc:5900
INFO:twisted:Starting factory <vncdotool.command.VNCDoCLIFactory instance at 0x7f8cbd29c368>
INFO:twisted:Using protocol version 3.800
INFO:twisted:Cannot connect Authentication failed
CRITICAL:root:Connection to the other side was lost in a non-clean fashion.
INFO:twisted:Stopping factory <vncdotool.command.VNCDoCLIFactory instance at 0x7f8cbd29c368>
INFO:twisted:Main loop terminated.
```
I've tried with or without a password. I've set the password using the monitor "set_password vnc password"

Anyone know how to do it or how to diagnose the problem?

sibson:
Various servers implement their own handshakes, I would guess Proxmox is doing something unexpected and you will need to debug the raw messages from [handleInit()](https://github.com/sibson/vncdotool/blob/master/vncdotool/rfb.py#L128) onward to discover what is being sent in the handshake and what the required responses are.

pmhahn:
Version `0.12.1` is way to old and many things changed since than. Please give the current version from git branch `main` a try:
> pip install git+https://github.com/sibson/vncdotool.git@main

It will also print more debug information. Similar to #245 Proxmay will probably require newer `AuthTypes`, which need to be implemented first.

sibson:
Update: PR #498 ("Implement the VeNCrypt security type (19)") landed 2026-09-07 and explicitly targets the TLS half of this issue, but its own description is direct about the gap: "Proxmox (#138) remains untested" — it was verified against TigerVNC and a headless wayvnc, not a real Proxmox host. So there's real progress (VeNCrypt/TLS auth now exists in `vncdotool/security/vencrypt.py`), but nobody has confirmed it actually satisfies Proxmox's handshake — `specs/server-compatibility-plan.md` still lists that as open, since Proxmox's own console auth (a signed ticket over `wss://`, `PVEAuthCookie`) may not even go through VeNCrypt at the RFB layer at all.

[...]

PR #498, opened 2026-09-07
Implement the VeNCrypt security type (19)

sibson:
Closes the `unknown security types: (19, ...)` half of #310 and the TLS half of #138: vncdotool speaks VeNCrypt 0.2 over Twisted's `startTLS`, behind a new `vncdotool[tls]` extra.

Three things a reviewer would not guess from the diff.

Verification is on by default and cannot be lost silently: an X509 chain is checked against the platform trust store or `--tls-ca-cert`, and the anonymous `TLS*` subtypes, which carry no certificate to check, are refused unless `--tls-insecure-skip-verify` is given.

The bare `Plain` subtype (256) is refused outright rather than gated. It has no TLS under it, so offering it let a server strip the TLS subtypes from its list and harvest the password in cleartext.

Handlers live in `vncdotool/security/`, the package #500 added.

Tested live: two TigerVNC fleet services, plus a headless wayvnc — the server #310 reports, and the only one offering `X509Plain`. Proxmox (#138) remains untested.


tests/functional/utils.py
```diff
@@ -102,6 +102,28 @@ class VNCServer(NamedTuple):
 TIGERVNC = VNCServer("tigervnc", 5931, size=(256, 192))
 TIGERVNC_AUTH = VNCServer("tigervnc-auth", 5932, password="vncdotool")
 X11VNC = VNCServer("x11vnc", 5933)
+# Out of DOCKER_SERVERS: neither can be reached without TLS options the
```
sibson, line 105: narrative?

tests/servers/Dockerfile
```diff
@@ -211,3 +212,21 @@ COPY servers/kasmvnc-entrypoint.sh /entrypoint.sh
 RUN chmod +x /entrypoint.sh
 
 ENTRYPOINT ["/entrypoint.sh"]
+
+
+# wayvnc, the VNC server Raspberry Pi OS ships. It drives wlroots
```
sibson: narrative

vncdotool/security/vencrypt.py
```diff
@@ -0,0 +1,94 @@
...
+        client._vencrypt_subtype = chosen
+        client.transport.write(pack("!I", chosen))
+
+        # The ack is sent for TLS and X509 subtypes only, which is every
+        # subtype this client will choose.
+        (ack,) = unpack("!B", (yield 1))
+        if ack != 1:
+            raise SecurityError(f"server refused {chosen!r}")
+        if client._packet:
+            # The TLS layer takes the socket over from here and has not sent
```
sibson, line 55: states code

Issue #259, opened 2023-04-28
Add support for WebSocket connections


jhachandan1994:
i am trying to connect with selenoid vnc server where path is present along with host port and password how can i get this done server looks something like ws://localhost:4444/vnc/c2ec57a377e94f515b35b2a57caad26e?password=selenoid. i dont see any argument of path in vncdotool i am able to connect to this server using clients like novncserver in javascript but i want to connect in python.

pmhahn:
`vncdotool` currently only supports connecting via plain TCP socket, not via a WebSocket (`ws://`)

mjb2k:
We had the same problem and so we built a dedicated Python based WebSocket VNC client here: https://github.com/Cynnovative/py-wsvnc

kevemueller:
I would very much welcome native support of websockets in vncdotool.
Currently I am using a workaround with [`websocat`](https://github.com/vi/websocat) doing the translation, this is particularly un-elegant when it comes to tearing down the socket.

```sh
websocat -b tcp-listen:127.0.0.1:5900 wss://vncserverexposedoverwebsocket &
pid=$!
vncdo -s 127.0.0.1 -p "secret" --delay=20 --force-caps  myscript.vdo
kill "${pid}"
```
I have tried with `--oneshot` at the `websocat` end as well, works less intuitively.

sibson:
Fixed by #499 (`881ac75`, "Connect to ws:// and wss:// VNC servers (#259)"). `vncdo -s ws://HOST:PORT/PATH` and `wss://` URLs connect directly, with no `websocat` bridge in front. The URL's path and query string are sent as given, and `wss://` verifies the certificate against the system trust store. `vnclog -s ws://…` also works on `main`.

This is on `main` and ships in 2.0.0, which isn't released yet. Until then, install from git to use it. Reopen this if it doesn't work for your server.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #499, opened 2026-09-07
Connect to ws:// and wss:// VNC servers (#259)

sibson:
`vncdo -s ws://host:4444/vnc/<session>?token=...` and `wss://` now work, closing #259.

Tested only against websockify, the transport could not reach Selenoid -- the reporter's own server. It 403s a client sending no `Origin` and 400s one offering two subprotocols, both autobahn defaults; a browser always sends Origin, which is why noVNC worked. The fleet now runs four implementations of the handshake (QEMU, Selenoid, KasmVNC, websockify), which is what surfaced it.

autobahn and `Twisted[tls]` become required, so a plain `pip install vncdotool` reaches a `ws://` server. `parse_server()` returns a `WEBSOCKET` sentinel where the address family goes, the polymorphism a Unix socket path already uses. `wss://` always verifies, with no opt-out flag -- unlike #498.

`tests/servers/PORTS.md` registers host ports; #497/#498/#499 each picked blind. `vnclog` still refuses `ws://`; it dials through `portforward`, see #502.


docs/install.md
```diff
@@ -10,6 +10,18 @@ vncdotool relies on a number of libraries, the two major ones are [Pillow](https
...
 While vncdotool should work with any recent version of these libraries sometimes things break.
 
+## Optional extras
+
+Connecting to a `ws://` or `wss://` server -- noVNC, websockify, Proxmox,
+Selenoid -- needs [autobahn](https://autobahn.readthedocs.io/), which the
+`websocket` extra pulls in:
+
+```
+pip install 'vncdotool[websocket]'
```
sibson: let's make websocket there by default.  We want this to just work not force users to think,

docs/usage.md
```diff
@@ -135,6 +135,42 @@ region that excludes whatever is animating.
...
+
+`-s`/`--server` takes a TCP address as `HOST`, `HOST:DISPLAY` or
+`HOST::PORT`, an IPv6 address in brackets, or the path of a Unix socket:
+
+```
+> vncdo -s 10.11.12.13::5901 capture screen.png
+> vncdo -s /run/user/1000/vnc.sock capture screen.png
+```
+
+It also takes a `ws://` or `wss://` URL, which is how noVNC, websockify,
```
sibson: too verbose for usage, whiich should be short and skimmable

tests/functional/utils.py
```diff
@@ -85,6 +87,9 @@ class VNCServer(NamedTuple):
     # The default suits a container on loopback; an OS-hosted server sharing
     # a busy machine's real desktop can be far slower.
     timeout: float = CONNECT_TIMEOUT
+    # What to pass `vncdo -s`, where host and port do not say it: a ws:// or
```
sibson: pointless

tests/functional/utils.py
```diff
@@ -104,6 +109,70 @@ class VNCServer(NamedTuple):
 
 DOCKER_SERVERS = [TIGERVNC, TIGERVNC_AUTH, X11VNC, LIBVNCSERVER_EXAMPLE]
 
+# websockify serves every path, so the path and query in these URLs name no
```
sibson: I HATE websockify, stop using it as a primary thing

tests/functional/utils.py
```diff
@@ -104,6 +109,70 @@ class VNCServer(NamedTuple):
...
+WEBSOCKIFY = VNCServer(
+    "websockify", 5942, size=(256, 192),
+    address="ws://127.0.0.1:5942/vnc/a-session?password=vncdotool",
+)
+WEBSOCKIFY_TLS = VNCServer(
+    "websockify-tls", 5943, size=(256, 192),
+    address="wss://localhost:5943/vnc/a-session?password=vncdotool",
+)
+
+# QEMU answers 404 to any request-URI but exactly "/", so neither address
```
sibson: belongs in commit not in code

tests/servers/Dockerfile
```diff
@@ -136,3 +136,82 @@ COPY --from=libvncserver-build /tmp/libvncserver/libvncserver.so* /usr/local/lib
...
+
+# 5900 here is only the Xvnc behind the bridge, so the base healthcheck would
+# pass before Selenoid is listening on 4444.
+HEALTHCHECK --interval=2s --timeout=2s --start-period=5s --retries=30 \
+    CMD ss -H -ltn 'sport = :4444' | grep -q . && ss -H -ltn 'sport = :5900' | grep -q .
+
+ENTRYPOINT ["/entrypoint.sh"]
+
+
+# KasmVNC, whose X server serves WebSocket itself; its `rfbport` default of
```
sibson: no

tests/servers/Dockerfile
```diff
@@ -136,3 +136,82 @@ COPY --from=libvncserver-build /tmp/libvncserver/libvncserver.so* /usr/local/lib
...
+        openssl \
+    && rm -rf /var/lib/apt/lists/*
+
+COPY servers/websockify-entrypoint.sh /entrypoint.sh
+RUN chmod +x /entrypoint.sh
+
+ENTRYPOINT ["/entrypoint.sh"]
+
+
+# QEMU's built-in VNC server, which speaks WebSocket itself rather than
```
sibson: strike

vncdotool/command.py
```diff
@@ -354,7 +354,11 @@ def build_tool(options: argparse.Namespace, args: list[str]) -> VNCDoCLIFactory:
     # no outcome decided yet; set before connecting so a synchronous
     # connection failure has somewhere to record itself
     reactor.exit_status = None
-    factory_connect(factory, options.host, options.port, options.address_family)
+    try:
+        factory_connect(factory, options.host, options.port, options.address_family)
+    except websocket.WebSocketUnavailable as exc:
```
sibson: why does this exception bleed into command rather than becoming a connection error like others?

vncdotool/command.py
```diff
@@ -537,6 +553,8 @@ def vnclog() -> None:
     setup_logging(options)
 
     options.address_family, options.host, options.port = parse_server(options.server)
+    if options.address_family is websocket.WEBSOCKET:
+        parser.error("vnclog records a TCP or Unix-socket server; ws:// is vncdo-only")
```
sibson, line 553: why can't we log ws?

vncdotool/websocket.py
```diff
@@ -0,0 +1,179 @@
...
+import socket
+from functools import lru_cache
+from typing import TYPE_CHECKING, Any, Union
+from urllib.parse import urlsplit, urlunsplit
+
+from twisted.internet.endpoints import HostnameEndpoint, wrapClientTLS
+from twisted.internet.error import ConnectionDone, ConnectionLost
+from twisted.python.failure import Failure
+
+if TYPE_CHECKING:
```
sibson: why do we have an if here.  seems like a turd.

docs/install.md
sibson: 🤖 Claude Code: autobahn and Twisted[tls] are now hard dependencies and the websocket extra is gone, so a plain `pip install vncdotool` reaches a ws:// server. The Optional extras section is deleted. f65a637

docs/usage.md
sibson: 🤖 Claude Code: Cut to about a third: the URL example, one line on quoting and the session path, and two lines on SSL_CERT_FILE. The certificate rationale and the extras block are gone. f65a637

tests/functional/utils.py
sibson: 🤖 Claude Code: Deleted. f65a637

tests/functional/utils.py
sibson: 🤖 Claude Code: Deleted; the QEMU 404 rule is in the commit body instead. f65a637

tests/servers/Dockerfile
sibson: 🤖 Claude Code: Deleted. f65a637

tests/servers/Dockerfile
sibson: 🤖 Claude Code: Struck. f65a637

vncdotool/command.py
sibson: 🤖 Claude Code: It does not any more. With autobahn a hard dependency the import cannot fail, so WebSocketUnavailable and its install hint are deleted and the try/except around factory_connect goes with them. A ws:// connection failure is now an ordinary connection error on the same path as any other. f65a637

vncdotool/websocket.py
sibson: 🤖 Claude Code: Gone. The if TYPE_CHECKING block only existed because autobahn and optionsForClientTLS were optional imports; now they are plain module-level imports. f65a637

tests/functional/utils.py
sibson: 🤖 Claude Code: Demoted rather than deleted, for now. QEMU, Selenoid and KasmVNC now lead WEBSOCKET_SERVERS and websockify is last, so it no longer reads as the reference implementation. It is still the capture-comparison baseline, because it is the only one of the six proxying a TigerVNC the TCP tests also reach, which is what makes a byte-identical comparison possible.

Leaving this open: say the word and I will drop the websockify and websockify-tls services entirely. f65a637

vncdotool/command.py
sibson: 🤖 Claude Code: No good reason, just not wired up. vnclog proxies through twisted.protocols.portforward.ProxyFactory, which dials upstream with reactor.connectTCP directly and never goes through factory_connect, so the WebSocket path is unreachable from it. The guard is honest about that rather than letting it fail obscurely.

Making it work means overriding the proxy's outbound connect, roughly 30-40 lines plus a functional test. Happy to do it here, but it is orthogonal to the transport and this PR is already ~1500 lines, so my suggestion is a follow-up issue. Leaving this open for your call.

sibson: 🤖 Claude Code: Filed as #502 with the portforward detail and the shape of the fix. Resolving this thread; the work carries on there.

PR #500, opened 2026-09-07
Move the RFB security types onto pluggable handlers

sibson:
`vncdotool/security/` mirrors `messages/`: one generator handler per RFB
security type, driven by the existing `_pump`. None, VNC Authentication and ARD
Diffie-Hellman move over; `rfb.py` loses 111 lines and the handler state it was
keeping on the client. Design in `specs/security-handler-architecture.md`.

Two things the diff won't tell you. `handle()` returns `bool` where
`MessageHandler` returns `None` — the old failure signal was *not* calling
`_doClientInitialization`, which `None` cannot carry. `_challenge` stays on the
client because `sendPassword` is published and may be called out of band, after
the handler's locals are gone.

RFB 3.3 now dispatches through the registry, so it accepts ARD where it did
not. Unreachable in practice: ARD servers announce 3.889 and negotiate to 3.8.

725 unit tests (+18, no existing test edited), flake8 0, mypy 72 against main's
74. The functional suite gave identical results on both branches. No CHANGELOG
entry — internal refactor, nothing user-visible.

### DIFF
--- a/tests/unit/test_security.py
+++ b/tests/unit/test_security.py
@@ -1,0 +1,211 @@
+import warnings
+from contextlib import contextmanager
+from struct import pack
+from unittest import TestCase, mock
+
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives.asymmetric import dh
+from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
+from cryptography.utils import CryptographyDeprecationWarning
+
+from vncdotool import security
+from vncdotool.const import AuthTypes
+
+SECURITY_OK = pack("!I", 0)
+SECURITY_FAILED = pack("!I", 1)
+SECURITY_TOO_MANY = pack("!I", 2)
+
+# RFC 3526 group 14, a published 2048-bit safe prime.
+MODP_2048 = int(
+    "FFFFFFFFFFFFFFFFC90FDAA22168C234C4C6628B80DC1CD129024E08"
+    "8A67CC74020BBEA63B139B22514A08798E3404DDEF9519B3CD3A431B"
+    "302B0A6DF25F14374FE1356D6D51C245E485B576625E7EC6F44C42E9"
+    "A637ED6B0BFF5CB6F406B7EDEE386BFB5A899FA5AE9F24117C4B1FE6"
+    "49286651ECE45B3DC2007CB8A163BF0598DA48361C55D39A69163FA8"
+    "FD24CF5F83655D23DCA3AD961C62F356208552BB9ED529077096966D"
+    "670C354E4ABC9804F1746C08CA18217C32905E462E36CE3BE39E772C"
+    "180E86039B2783A2EC07A28FB5C55DF06F4C52C9DE2BCBF695581718"
+    "3995497CEA956AE515D2261898FA051015728E5A8AACAA68FFFFFFFF"
+    "FFFFFFFF",
+    16,
+)
+
+
+@contextmanager
+def ffdh():
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", CryptographyDeprecationWarning)
+        yield
+
+
+def drive(handler, client, *blocks):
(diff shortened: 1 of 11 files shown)


vncdotool/security/ard.py
```diff
@@ -0,0 +1,66 @@
...
+
+from ..const import AuthTypes
+from .base import SecurityHandler, security_result
+
+
+class DiffieHellmanHandler(SecurityHandler):
+    SECURITY_TYPE = AuthTypes.DIFFIE_HELLMAN
+
+    def handle(self, client: Any) -> Generator[int, bytes, bool]:
+        # rfbproto, Diffie-Hellman Authentication: a U16 generator and a U16
```
sibson: redundent

vncdotool/security/ard.py
```diff
@@ -0,0 +1,66 @@
...
+
+def _encrypt(
+    client: Any, generator: int, key_len: int, modulus: bytes, server_key: bytes
+) -> bytes:
+    userStruct = f"{client.factory.username:\0<64}{client.factory.password:\0<64}"
+
+    p = int.from_bytes(modulus, "big")
+    sk = int.from_bytes(server_key, "big")
+    with warnings.catch_warnings():
+        # ARD auth is specified over classic finite-field DH; the server
```
sibson: pollution, external concern, unnecessary to understand code

vncdotool/security/ard.py
sibson: 🤖 Claude Code: Removed in 43362e1. The three lines below unpack into `generator`, `key_len`, `modulus` and `server_key`, so the comment named the fields the code already names.

vncdotool/security/ard.py
sibson: 🤖 Claude Code: Removed in 43362e1, issue link included. Both comments came across verbatim from rfb.py rather than being written for this file, which is how they got past a sweep.

vncdotool/security/ard.py
```diff
@@ -0,0 +1,66 @@
...
+    private_key = params.generate_private_key()
+    shared_key = private_key.exchange(server_public)
+
+    h = hashes.Hash(hashes.MD5())
+    h.update(shared_key)
+    key_digest = h.finalize()
+
+    cipher = Cipher(algorithms.AES(key_digest), modes.ECB())
+    encryptor = cipher.encryptor()
+    ciphertext = encryptor.update(userStruct.encode("utf-8"))
```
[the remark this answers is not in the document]
sibson: 🤖 Claude Code: Not acting on this, leaving the thread open for a maintainer to judge.

AES-ECB here is dictated by the Apple Remote Desktop authentication scheme, not chosen: the credential struct is encrypted ECB with an MD5 of the Diffie-Hellman shared secret, and a server accepts nothing else. This code moved unchanged from `RFBClient._encryptArd` in `rfb.py`, so the alert is new to CodeQL only because the file is new.

Changing the mode would break ARD authentication outright. The alert wants dismissing as "used in tests"/"false positive" against the protocol, or the repo lives with it.

tests/unit/test_security.py
```diff
@@ -0,0 +1,211 @@
...
+        with ffdh():
+            shared = server_private.exchange(
+                dh.DHPublicNumbers(
+                    int.from_bytes(client_public, "big"), params
+                ).public_key()
+            )
+        digest = hashes.Hash(hashes.MD5())
+        digest.update(shared)
+        decryptor = Cipher(algorithms.AES(digest.finalize()), modes.ECB()).decryptor()
+        plaintext = decryptor.update(ciphertext) + decryptor.finalize()
```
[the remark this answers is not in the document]
sibson: 🤖 Claude Code: Not acting on this, leaving the thread open for a maintainer to judge.

Same root as the `security/ard.py` alert: this is the test oracle, which runs the *server* half of the ARD exchange and decrypts what the client sent. It has to use the mode the protocol specifies, or it would not verify the client against anything real.

Using a stronger mode here would make the test pass while checking nothing.

PR #506, opened 2026-09-08
Establish libvncserver-example's size instead of assuming it

sibson:
`TestServer_libvncserver_example.test_capture` was flaky, once capturing 640x480 where 800x600 was declared. The demo's framebuffer size is process-global state any client can change: XK_Up/XK_Down call `rfbNewFramebuffer` on the whole `rfbScreen`, and the new size outlives the client that asked for it, so a fresh connection gets whatever the last one left. Confirmed against a throwaway container — two Downs, disconnect, reconnect, still 640x480.

The readiness gate now drives the size it is about to assert (up, up, down clamps to 800x600 from any state) and then checks the capture against the declared size, so a fleet that cannot serve what the tests expect fails at `servers-up`. `test_desktop_resize` restores the same way via `addCleanup`.

Two suites running against the fleet at once can still race; nothing short of a per-run server fixes that.

### DIFF
--- a/tests/functional/test_desktop_resize.py
+++ b/tests/functional/test_desktop_resize.py
@@ -7,7 +7,7 @@ from .utils import (
     LIBVNCSERVER_EXAMPLE,
     distinct_colours,
     has_expected_content,
-    normalize_framebuffer,
+    normalize_size,
     port_open,
     run_vncdo,
     screenshot_dir,
@@ -25,9 +25,7 @@ class TestDesktopResize(TestCase):
                 f"{LIBVNCSERVER_EXAMPLE.name} not reachable on {HOST}:{LIBVNCSERVER_EXAMPLE.port} -- "
                 f"{LIBVNCSERVER_EXAMPLE.how_to_start}"
             )
-        # The size this test leaves behind is the size every later client of
-        # the fleet gets, including another checkout's run.
-        self.addCleanup(normalize_framebuffer, LIBVNCSERVER_EXAMPLE)
+        self.addCleanup(normalize_size, LIBVNCSERVER_EXAMPLE)
 
     def test_mid_session_resize_is_decoded(self) -> None:
         """Each capture's size reflects the server's actual PSEUDO_DESKTOP_SIZE
--- a/tests/functional/test_desktop_resize.py
+++ b/tests/functional/test_desktop_resize.py
@@ -2,7 +2,16 @@ from unittest import TestCase
 
 from PIL import Image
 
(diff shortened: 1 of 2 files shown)


tests/functional/test_desktop_resize.py
```diff
@@ -16,6 +25,9 @@ def setUp(self) -> None:
                 f"{LIBVNCSERVER_EXAMPLE.name} not reachable on {HOST}:{LIBVNCSERVER_EXAMPLE.port} -- "
                 f"{LIBVNCSERVER_EXAMPLE.how_to_start}"
             )
+        # The size this test leaves behind is the size every later client of
```
sibson: explainatory

tests/functional/utils.py
```diff
@@ -92,6 +92,10 @@ class VNCServer(NamedTuple):
     how_to_start: str = "start the servers first with `make servers-up`"
     # Honoured only off CI -- see absent_server_skips().
     skip_when_down: bool = False
+    # Keys that drive a server whose framebuffer size any client can change
```
sibson: huge coment for a bad var name

tests/functional/utils.py
```diff
@@ -122,8 +126,15 @@ class VNCServer(NamedTuple):
 WAYVNC_CA_CERT = (
     Path(__file__).resolve().parents[1] / "servers" / "wayvnc-certs" / "cert.pem"
 )
-# 800x600 is the demo's hard-coded size; it takes no -geometry option.
-LIBVNCSERVER_EXAMPLE = VNCServer("libvncserver-example", 5935, size=(800, 600))
+# 800x600 is the size the demo starts at; it takes no -geometry option. Any
```
sibson: explaining, behaviour of external code is not our job.   normalize_keys = () # force screen to 800x600 would be enough.

tests/functional/test_desktop_resize.py
sibson: 🤖 Claude Code: Dropped the comment; addCleanup(normalize_size, ...) reads on its own. bf82a26

tests/functional/utils.py
sibson: 🤖 Claude Code: Renamed to normalize_size_keys and deleted the comment — the name says what it normalizes. bf82a26

tests/functional/utils.py
sibson: 🤖 Claude Code: Cut to two lines — the start-up size and that any client can change it, which is why the field exists. The arrow-key/clamp mechanics are in 77057a5. bf82a26

PR #507, opened 2026-09-09
Tell the distiller which JPEG quality level the capture asked for

sibson:
`make goldens` writes truncated lossy fixtures and reports success. The distiller replays `s2c.bin` alone, so it never offers the JPEG quality level the capture asked for, and `TightDecoder` refuses the first JPEG rectangle; the abort reaches a mocked factory and vanishes. Driving the fleet at quality 9 and 5 gives two-step fixtures where the committed ones hold eight. The guard that refuses postdates the last capture of those fixtures, which is the only reason they are whole today.

`split` now raises on anything that stopped the parser, so this cannot pass silently again.

Not run by CI: `make goldens` against the fleet before and after, which reproduces the eight committed steps of both lossy fixtures, and `make test-func`, green but for four macOS Screen Sharing cases failing on this host's auth.

### DIFF
--- a/specs/decoder-goldens.md
+++ b/specs/decoder-goldens.md
@@ -154,6 +154,14 @@ Auth stripping is irrelevant here: golden capture targets no-auth servers.
 a `NullTransport`, recording each FramebufferUpdate's raw bytes and the
 encoding actually used. It adds no wire parser and starts no reactor.
 
+The replay client is told the pixel format *and* the JPEG quality level the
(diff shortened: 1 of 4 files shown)


sibson:
🤖 Claude Code: superseded by #510 — the branch was renamed to `fix/distiller-jpeg-quality`, which closed this PR rather than retargeting it. The scene-image change this PR originally carried was dropped; see #510 for why.

PR #509, opened 2026-09-09
Stop every grid case on a fleet another checkout started

sibson:
The compose project name is global while the cert and scene mounts are relative to the compose file, so the first worktree to run `make servers-up` owns the fleet and every other worktree's `make test-func` silently drives that worktree's containers, certificates and scene PNGs. `fleet_mismatch()` already diagnoses this precisely, but only `test_scene_player` asked it, and `make test-func` is a bare `unittest discover` that never runs `wait_for_servers.py`.

Asking from `FleetTestCase.setUpClass` puts the real message on all 119 grid classes; the answer is cached per server so the grid doesn't spend two subprocesses per class re-deriving one tree hash.

Found while investigating 43 failures on a clean `origin/main` that turned out to be 39 stale-fleet plus 4 unrelated. Rebuilding the fleet took those 39 to zero; faking the checkout tag then confirmed all 119 classes error with the cause rather than running.

### DIFF
--- a/tests/functional/utils.py
+++ b/tests/functional/utils.py
@@ -343,16 +343,30 @@ def running_fleet_tag(server: VNCServer) -> Optional[str]:
     return tag or None
 
 
+_FLEET_MISMATCHES: Dict[str, Optional[str]] = {}
+
+
 def fleet_mismatch(server: VNCServer) -> Optional[str]:
-    """Why `server` is not this checkout's fleet, or None if it is or may be."""
+    """Why `server` is not this checkout's fleet, or None if it is or may be.
+
+    Answered once per server and remembered: a grid of a hundred cases asks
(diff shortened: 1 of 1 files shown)


tests/functional/utils.py
```diff
@@ -343,16 +343,30 @@ def running_fleet_tag(server: VNCServer) -> Optional[str]:
...
 
 
+_FLEET_MISMATCHES: Dict[str, Optional[str]] = {}
+
+
 def fleet_mismatch(server: VNCServer) -> Optional[str]:
-    """Why `server` is not this checkout's fleet, or None if it is or may be."""
+    """Why `server` is not this checkout's fleet, or None if it is or may be.
+
+    Answered once per server and remembered: a grid of a hundred cases asks
```
sibson, line 352: describes code

PR #510, opened 2026-09-09
Tell the distiller which JPEG quality level the capture asked for

sibson:
`make goldens` writes truncated lossy fixtures and reports success. The distiller replays `s2c.bin` alone, so it never offers the JPEG quality level the capture asked for, and `TightDecoder` refuses the first JPEG rectangle; the abort reaches a mocked factory and vanishes. Driving the fleet at quality 9 and 5 gives two-step fixtures where the committed ones hold eight. The guard that refuses postdates the last capture of those fixtures, which is the only reason they are whole today.

`split` now raises on anything that stopped the parser, so this cannot pass silently again.

Not run by CI: `make goldens` against the fleet before and after, which reproduces the eight committed steps of both lossy fixtures, and `make test-func`, green but for four macOS Screen Sharing cases failing on this host's auth.

### DIFF
--- a/tests/goldens/distill.py
+++ b/tests/goldens/distill.py
@@ -49,8 +49,6 @@ class _Recorder(client.VNCDoToolClient):
         self.init_end = self.consumed
 
     def vncProtocolError(self, reason: str) -> None:
-        # The error reaches a mocked factory and goes nowhere, so it is kept
-        # here for `split` to raise on.
         self.abort_reason = reason
         super().vncProtocolError(reason)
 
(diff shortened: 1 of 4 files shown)


tests/goldens/distill.py
```diff
@@ -42,11 +42,18 @@ def __init__(self) -> None:
...
         self.init_end: Optional[int] = None
         self.consumed = 0
+        self.abort_reason: Optional[str] = None
 
     def vncConnectionMade(self) -> None:
         super().vncConnectionMade()
         self.init_end = self.consumed
 
+    def vncProtocolError(self, reason: str) -> None:
+        # The error reaches a mocked factory and goes nowhere, so it is kept
```
sibson: describes code

tests/goldens/distill.py
```diff
@@ -82,14 +95,17 @@ def split(self, s2c: bytes) -> Tuple[bytes, List[Step]]:
...
 
 
-def _make_client(pixel_format: str) -> _Recorder:
-    """SetPixelFormat is client-to-server (RFC 6143 section 7.5.1), so the s2c
-    stream being replayed never says the server switched layouts. The recorder
-    has to be told, or it unpacks the bytes as ServerInit announced them and
-    permutes every channel.
+def _make_client(pixel_format: str, jpeg_quality: Optional[int] = None) -> _Recorder:
+    """SetPixelFormat and SetEncodings are client-to-server (RFC 6143 sections
+    7.5.1 and 7.5.2), so the s2c stream being replayed never says the server
```
sibson: narrative

tests/goldens/distill.py
sibson: 🤖 Claude Code: Deleted in 3d352de. The fact it carried — the error reaches a mocked factory and goes nowhere — is already in the commit body that introduced the method.

tests/goldens/distill.py
sibson: 🤖 Claude Code: Reshaped in 3d352de. I kept the facts and fixed the shape rather than deleting: the pre-existing half of this docstring survived the repo-wide sweep in 71d5b9b, and my edit turned it into one chained sentence. Now one fact per clause, with the two failure modes stated apart.

The same pass caught two spec paragraphs you did not flag — Driving was re-teaching what expect-matching.md defines, Distillation was restating its own commit body.

PR #511, opened 2026-09-10
Name protocol values in messages instead of printing enum reprs

sibson:
Protocol errors printed Python enum reprs — `<AuthTypes.VENCRYPT: 19>`, or a quoted `'<AuthTypes.UNKNOWN: 13ff>'` where the enum had no member. All sixteen call sites already used `!r`, so the rendering is fixed on `IntEnumLookup` rather than at each site, and `AuthTypes`/`VeNCryptSubtypes` label themselves with rfbproto's spelling — what a user finds in their server's configuration. An unnamed value prints in the base rfbproto writes that range in: hex for the vendor blocks, decimal below them.

The refusal behind #310 now names what the server offered and what vncdotool supports. Two fixes beyond that: rfbproto allows a plain security type among the VeNCrypt subtypes, which printed as `unknown (2)`; and the connection-refused reason is the server's own text, previously shown as a bytes literal.


vncdotool/const.py
```diff
@@ -1,12 +1,50 @@
 from __future__ import annotations
 
 from enum import IntEnum, IntFlag
+from typing import cast
+
+
+def _named(name: str, value: int) -> str:
+    # rfbproto writes the vendor blocks in hex and everything below them in
+    # decimal, and this number is here to be looked up there.
```
sibson, line 9: here to be looked up there?

vncdotool/vencrypt.py
```diff
@@ -115,14 +115,23 @@ def choose(
     return None
 
 
+def name(subtype: int) -> str:
+    """What to call `subtype` in a message."""
+    # rfbproto: any normal security type may be listed among the subtypes,
+    # which are numbered from 256 up.
```
sibson, line 121: so what?

PR #512, opened 2026-09-10
Reclassify KasmVNC as Broken, and add --dialect to drive it anyway

sibson:
KasmVNC reads eleven bytes for the PointerEvent that RFB 6143 defines as six, so a conforming client's six wedge its reader and the connection dies twenty seconds later. README moves KasmVNC from Supported to Broken; upstream declines to support VNC clients (kasmtech/KasmVNC#119).

`--dialect NAME` mixes a per-server class into whichever client the entry point already uses, so the CLI keeps `VNCDoCLIClient` and the library keeps its `factory_class`. `--dialect kasmvnc` sends the eleven-byte form: the session survives and the pointer moves, checked with `XQueryPointer` inside the container. A press reaches X but has not been seen to reach an application. Misapplying a dialect desyncs a conforming server.

Stacked on #515, which is red on kasmvnc until this lands. The `--dialect` functional case has not run through the harness — the fleet is serving another checkout's images — so it was checked by driving `vncdo` at the fleet directly.

### DIFF
--- a/tests/functional/test_scene_player.py
+++ b/tests/functional/test_scene_player.py
@@ -98,7 +98,7 @@ class ClickSelectsAScene:
 
 
 for _server in SCENE_SERVERS:
-    if not _server.accepts_pointer_events:
+    if _server.skip_pointer_tests:
         continue
     _name = "TestClick_" + _server.name.replace("-", "_")
     globals()[_name] = type(
--- a/tests/functional/test_scene_player.py
+++ b/tests/functional/test_scene_player.py
@@ -98,6 +98,8 @@ class ClickSelectsAScene:
 
 
 for _server in SCENE_SERVERS:
+    if not _server.accepts_pointer_events:
+        continue
     _name = "TestClick_" + _server.name.replace("-", "_")
     globals()[_name] = type(
         _name, (ClickSelectsAScene, FleetTestCase), {"server": _server, "__module__": __name__}
--- a/tests/functional/test_server_compat_docker.py
+++ b/tests/functional/test_server_compat_docker.py
@@ -67,3 +67,16 @@ class TestKasmVNCInput(FleetTestCase):
             f"exited {result.returncode} rather than timing out, stderr:\n"
             f"{result.stderr}",
         )
+
+    def test_the_kasmvnc_dialect_keeps_the_session_answering(self) -> None:
+        result, png = self.capture_after("--dialect", "kasmvnc", "move", "10", "10")
+        self.assertEqual(
+            result.returncode, 0,
+            f"KasmVNC: --dialect kasmvnc exited {result.returncode}, stderr:\n"
+            f"{result.stderr}",
+        )
+        self.assertTrue(
+            png.exists(),
+            "KasmVNC served no framebuffer update after an eleven-byte "
+            "PointerEvent, so --dialect kasmvnc no longer matches its reader",
+        )
--- a/tests/functional/test_server_compat_docker.py
+++ b/tests/functional/test_server_compat_docker.py
@@ -24,18 +24,24 @@ register_server_tests(SUPPORTED_SERVERS, globals(), base=FleetTestCase)
 
 
 class TestKasmVNCInput(FleetTestCase):
+    """Locks the KasmVNC behaviour that keeps it out of SUPPORTED_SERVERS.
+
+    A failure here means the server changed: re-evaluate whether it moves
+    back to README.md's Supported class.
+    """
+
     server = KASMVNC
     # Long enough that a served update arrives, short enough that the wedge
     # surfaces well before KasmVNC's own 20-second idle timeout drops the
     # socket.
-    deadline = "5"
+    timeout = "5"
 
     def capture_after(self, *args: str) -> Tuple[subprocess.CompletedProcess, Path]:
         tmp = tempfile.TemporaryDirectory()
         self.addCleanup(tmp.cleanup)
         png = Path(tmp.name) / "after-input.png"
         result = run_vncdo(
-            self.server, "--timeout", self.deadline,
+            self.server, "--timeout", self.timeout,
             *args, "pause", "0.3", "capture", str(png),
         )
         return result, png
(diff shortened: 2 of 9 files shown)


tests/functional/test_server_compat_docker.py
```diff
@@ -9,8 +9,74 @@
...
+from vncdotool.command import ExitStatus
+
+from .utils import KASMVNC, SUPPORTED_SERVERS, FleetTestCase, register_server_tests, run_vncdo
 
 # Every scenario shells out to the vncdo CLI (see utils.run_vncdo), so
 # no reactor ever starts in this process and no api.shutdown() is needed.
 register_server_tests(SUPPORTED_SERVERS, globals(), base=FleetTestCase)
+
+
+class TestKasmVNCInput(FleetTestCase):
```
sibson, line 26: ''' Test locks broken KasmVNC behaviours, if these fail then re-evaluate if KasmVNC moves to fully supported. ```

tests/functional/test_server_compat_docker.py
```diff
@@ -9,8 +9,74 @@
...
+    server = KASMVNC
+    # Long enough that a served update arrives, short enough that the wedge
```
sibson, line 28: describes a timeout

tests/functional/utils.py
```diff
@@ -84,6 +84,9 @@ class VNCServer(NamedTuple):
     # False means a flat, usually all-black, framebuffer is expected rather
     # than a failure -- see the macOS note in tests/servers/screen-sharing.
     renders_desktop: bool = True
+    # False where a pointer event ends the session instead of being acted on,
```
sibson, line 87: False is over specific to the kasm case, should just be skip_pointer_tests.

vncdotool/client.py
```diff
@@ -654,6 +663,31 @@ def dataReceived(self, data: bytes) -> None:
             super().dataReceived(data)
 
 
+class VMWareClient(VMWareDialect, VNCDoToolClient):
```
sibson, line 666: Why does this need to be a double inheritence rather than just having the Dialect implmenented here on the client?

tests/functional/test_scene_player.py
```diff
@@ -98,6 +98,8 @@ def test_a_click_selects_the_scene_under_it(self) -> None:
 
 
 for _server in SCENE_SERVERS:
```
sibson, line 100: SUPPORTED_SERVERS?

tests/functional/test_server_compat_docker.py
sibson: 🤖 Claude Code: Added in 58e04c6, as a class docstring: "Locks the KasmVNC behaviour that keeps it out of SUPPORTED_SERVERS. A failure here means the server changed: re-evaluate whether it moves back to README.md's Supported class."

tests/functional/test_server_compat_docker.py
sibson: 🤖 Claude Code: Renamed `deadline` to `timeout` in 58e04c6 — it is the value passed to `vncdo --timeout`, so it now says so.

tests/functional/utils.py
sibson: 🤖 Claude Code: Renamed to `skip_pointer_tests` in 58e04c6, with the sense inverted (default `False`, `True` on KasmVNC). It now names the decision the grids make rather than the KasmVNC behaviour behind it, so `test_scene_player` and `test_websocket` read `if _server.skip_pointer_tests` / `if not server.skip_pointer_tests`.

vncdotool/client.py
sibson: 🤖 Claude Code: Because the dialect has to compose with a client class chosen elsewhere, and there is more than one.

The CLI does not use `VNCDoToolClient`: `VNCDoCLIFactory.protocol = VNCDoCLIClient`, which carries the exit-status and error handling. `api.connect` takes an arbitrary `factory_class` too. If a dialect were a concrete client subclass — which is what `VMWareClient` was — then selecting it would mean assigning `factory.protocol = VMWareClient` and throwing `VNCDoCLIClient` away, so `vncdo --dialect vmware` would lose its error reporting and exit statuses.

As a mixin, `apply_dialect` builds `type(name, (dialect, factory.protocol), {})` and the dialect lands in front of whatever class is already there. `class VMWareClient(VMWareDialect, VNCDoToolClient)` exists only to keep that public name working; it is not how the dialect is selected.

Happy to fold it back into a plain subclass if you would rather the CLI client and the dialects be enumerated as a fixed matrix instead.

tests/functional/test_scene_player.py
sibson: 🤖 Claude Code: It has to stay `SCENE_SERVERS` — this grid reads a scene back out of the framebuffer, so it needs a server running `tests/goldens/scene_player.py`.

`SUPPORTED_SERVERS` is wider: it also holds `LIBVNCSERVER_EXAMPLE`, which has no X server, and `QEMU`, which shows firmware. Neither can display a scene, so `click_targets.scene_at()` would have nothing to read back and both cases would fail for a reason unrelated to the button event.

Current membership:

- `SCENE_SERVERS` = tigervnc, x11vnc, wayvnc, selenoid, kasmvnc
- `SUPPORTED_SERVERS` = tigervnc, x11vnc, wayvnc, libvncserver-example, qemu, selenoid

`skip_pointer_tests` then drops kasmvnc, leaving tigervnc, x11vnc, wayvnc and selenoid, which is what registers today.

Leaving this open in case you meant something else by it.

tests/functional/utils.py
sibson: let's not say ends the session, there might be other reasons a server doesn't support pointer.  At which point this comment collapses to just the var name.

vncdotool/client.py
sibson: Should there not be a KasmClient?

sibson: 🤖 Claude Code: There should — added in 32dc842, mirroring the VMWare pair:

```python
class KasmVNCClient(KasmVNCDialect, VNCDoToolClient): ...
class KasmVNCFactory(VNCDoToolFactory):
    protocol = KasmVNCClient
```

You were pointing at a real gap, not just asymmetry. `--dialect kasmvnc` worked from the CLI, but `api.connect(server, password, factory_class=...)` had nothing to pass for it, while vmware had `VMWareFactory`. A library user could reach the vmware behaviour and not the kasmvnc one.

The mixin stays the single definition; the pair is the ready-made composition against the plain client. `TestKasmVNCDialect` now drives `KasmVNCClient` directly instead of building an anonymous subclass.

PR #513, opened 2026-09-10
Boot the fleet's QEMU on OVMF so its screen holds still

sibson:
QEMU had no encoding coverage because SeaBIOS blinks a VGA text cursor and no two captures of an idle machine match. Booting OVMF gives a screen that holds still and a shell `type` can drive, so the decoders now run against QEMU's Hextile, ZRLE and Tight — written from the specification rather than shared with the rest of the fleet, a variation they were otherwise never held against. The oracle is a Raw capture of the same screen, firmware having no committed image to compare against.

Note the size constant moves 720x400 → 1280x800.

`Read rectangle encodings out of the log format #511 left behind` fixes main, red at 0a8dfbb: 32 of 84 cases fail across the whole scene grid. Worth taking separately.

Checked by hand: with QEMU restarted immediately beforehand, the widened readiness gate reports `serving (640, 480), expected (1280, 800)` and waits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,10 @@ pull request: [TigerVNC](https://tigervnc.org),
 [KasmVNC](https://kasmweb.com/kasmvnc), [QEMU](https://qemu.org)'s built-in
 server, [Selenoid](https://aerokube.com/selenoid/).
 
+QEMU is the exception: it boots firmware with no guest, so there is no
+desktop to draw a known scene, and its encodings are checked against its own
+Raw output rather than against a committed image.
+
 **Supported on their own OS** — the same journey, but only on the pull
 requests that touch code able to affect it, since these servers need a
 Windows or macOS runner: UltraVNC on Windows, and Apple Screen Sharing on
--- a/README.md
+++ b/README.md
@@ -68,10 +68,6 @@ pull request: [TigerVNC](https://tigervnc.org),
 [KasmVNC](https://kasmweb.com/kasmvnc), [QEMU](https://qemu.org)'s built-in
 server, [Selenoid](https://aerokube.com/selenoid/).
 
-QEMU is the exception: it boots firmware with no guest, so there is no
-desktop to draw a known scene, and its encodings are checked against its own
-Raw output rather than against a committed image.
-
 **Supported on their own OS** — the same journey, but only on the pull
 requests that touch code able to affect it, since these servers need a
 Windows or macOS runner: UltraVNC on Windows, and Apple Screen Sharing on
--- a/tests/functional/test_encodings.py
+++ b/tests/functional/test_encodings.py
@@ -32,7 +32,7 @@ SCENES = ("0", "s")
 # Flat enough that every server encodes it the way it was asked to.
 HONOURED_SCENE = "0"
 
-RECTANGLE_ENCODING = re.compile(r"Received <Encoding\.([A-Z_]+):")
+RECTANGLE_ENCODING = re.compile(r"Received ([A-Z_]+) \(-?\d+\) rectangle")
 
 # Measured against the fleet by offering one encoding at a time and reading
 # back the encoding of the rectangles that arrived. Every server here
--- a/tests/functional/test_encodings.py
+++ b/tests/functional/test_encodings.py
@@ -64,13 +64,12 @@ def capture(
         return Image.open(path).convert("RGB").copy(), result.stderr
 
 
-# `cls N` repaints OVMF's whole shell framebuffer in colour N.
 QEMU_SCREENS = {
-    "flat": ("cls 4",),
+    "flat": ("cls 4",),  # `cls N` repaints OVMF's whole shell framebuffer in colour N
     "text": ("cls 1", "echo vncdotool encoding probe"),
 }
 
-# Measured against QEMU's firmware screens the way EMITTED was.
+# Offered one at a time; QEMU sent these and answered with Raw for the rest.
 QEMU_EMITTED = {"raw", "hextile", "zrle", "tight"}
 
 # The shell answers a command over several framebuffer updates, so a capture
--- a/tests/functional/test_encodings.py
+++ b/tests/functional/test_encodings.py
@@ -48,13 +48,13 @@ EMITTED: Dict[str, Set[str]] = {
 
 
 def capture(
-    test: TestCase, server: VNCServer, encodings: str, key: str
+    test: TestCase, server: VNCServer, encodings: str, *before: str
 ) -> Tuple[Image.Image, str]:
+    """The screen `before` leaves behind, and the log naming its rectangles."""
     with tempfile.TemporaryDirectory() as tmp:
         path = Path(tmp) / "screen.png"
         result = run_vncdo(
-            server, "-v", "-v", "--encodings", encodings,
-            "key", key, *awaiting(key), "capture", str(path),
+            server, "-v", "-v", "--encodings", encodings, *before, "capture", str(path),
         )
         if result.returncode != 0:
             test.fail(
@@ -77,27 +77,13 @@ QEMU_EMITTED = {"raw", "hextile", "zrle", "tight"}
 QEMU_SETTLE_SECONDS = "1"
 
 
-def capture_qemu(test: TestCase, encoding: str, *before: str) -> Tuple[Image.Image, str]:
-    with tempfile.TemporaryDirectory() as tmp:
-        path = Path(tmp) / "screen.png"
-        result = run_vncdo(
-            QEMU, "-v", "-v", "--encodings", encoding, *before, "capture", str(path)
-        )
-        if result.returncode != 0:
-            test.fail(
-                f"qemu: vncdo --encodings {encoding} failed "
-                f"({result.returncode}): {result.stderr}"
-            )
-        return Image.open(path).convert("RGB").copy(), result.stderr
-
-
 def draw_on_qemu(test: TestCase, screen: str) -> Image.Image:
     """Put one of QEMU_SCREENS up, and return the Raw capture of it."""
     argv: List[str] = []
     for command in QEMU_SCREENS[screen]:
         argv += ["type", command, "key", "enter"]
     argv += ["stable", QEMU_SETTLE_SECONDS]
-    oracle, _ = capture_qemu(test, "raw", *argv)
+    oracle, _ = capture(test, QEMU, "raw", *argv)
     return oracle
 
 
@@ -112,7 +98,9 @@ class RendersTheScene:
     scene: str
 
     def test_renders_the_scene_through_the_encoding_it_asked_for(self) -> None:
-        screen, log = capture(self, self.server, self.encoding, self.scene)
+        screen, log = capture(
+            self, self.server, self.encoding, "key", self.scene, *awaiting(self.scene)
+        )
         oracle = Image.open(SCENES_DIR / f"{self.scene}.png").convert("RGB")
         self.assertEqual(screen.size, oracle.size)
         self.assertEqual(
@@ -145,7 +133,7 @@ class RendersTheSameScreenAsRaw:
 
     def test_renders_the_screen_as_raw_renders_it(self) -> None:
         oracle = draw_on_qemu(self, self.screen)
-        screen, log = capture_qemu(self, self.encoding)
+        screen, log = capture(self, QEMU, self.encoding)
         self.assertEqual(
             screen.tobytes(), oracle.tobytes(),
             f"qemu: {self.encoding} and raw disagree about the {self.screen} screen",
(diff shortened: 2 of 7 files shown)


tests/functional/test_encodings.py
```diff
@@ -63,6 +65,57 @@ def capture(
...
+        result = run_vncdo(QEMU, "-v", "-v", "--encodings", encoding, "capture", str(path))
+        if result.returncode != 0:
+            test.fail(
+                f"qemu: vncdo --encodings {encoding} failed "
+                f"({result.returncode}): {result.stderr}"
+            )
+        return Image.open(path).convert("RGB").copy(), result.stderr
+
+
+def type_commands(test: TestCase, commands: Tuple[str, ...]) -> None:
```
sibson: qemu specfic function with generic name

tests/functional/test_encodings.py
```diff
@@ -63,6 +65,57 @@ def capture(
...
+    argv: List[str] = []
+    for command in commands:
+        argv += ["type", command, "key", "enter"]
+    result = run_vncdo(QEMU, *argv)
+    if result.returncode != 0:
+        test.fail(f"qemu: vncdo {' '.join(argv)} failed ({result.returncode}): {result.stderr}")
+
+
+def settled(test: TestCase) -> Image.Image:
```
sibson: we have a vncdo command for this...

tests/functional/test_encodings.py
```diff
@@ -98,8 +151,42 @@ def test_renders_the_scene_through_the_encoding_it_asked_for(self) -> None:
         )
 
 
+class RendersTheSameScreenAsRaw:
```
sibson: what does this prove and do we need to prove it?

tests/functional/test_encodings.py
```diff
@@ -98,8 +151,42 @@ def test_renders_the_scene_through_the_encoding_it_asked_for(self) -> None:
...
+        arrived = set(RECTANGLE_ENCODING.findall(log))
+        self.assertIn(
+            wanted.name, arrived,
+            f"no {wanted!r} rectangle arrived; qemu answered with {sorted(arrived)}",
+        )
+
+
 def load_tests(loader: unittest.TestLoader, tests: unittest.TestSuite, pattern: object) -> unittest.TestSuite:
     suite = unittest.TestSuite()
+    for encoding in sorted(decoders.ENCODING_NAMES):
```
sibson, line 153: this probably deserves a comment or putting into a function to self comment

tests/functional/utils.py
```diff
@@ -461,9 +462,30 @@ def connect(server: VNCServer, timeout: Optional[float] = None) -> api.ThreadedV
     return client
 
 
-def capture_screenshot(server: VNCServer, path: Path, timeout: Optional[float] = None) -> Path:
+@contextlib.contextmanager
+def probe_context(server: VNCServer) -> Iterator[Optional[Dict[str, str]]]:
+    """What a bare `vncdo capture` of this server needs around it.
+
+    Selenoid's /vnc/ route resolves only while a WebDriver session is open,
```
sibson: why call out selenoid here.

tests/servers/qemu-entrypoint.sh
```diff
@@ -32,10 +32,14 @@ else
     set --
 fi
 
-# No disk: SeaBIOS draws its own boot screen, which is a real framebuffer
-# with stable content and needs no guest image.
+# No disk, so the firmware's own screen is the framebuffer. SeaBIOS blinks a
```
sibson: narration

README.md
```diff
@@ -68,6 +68,10 @@ pull request: [TigerVNC](https://tigervnc.org),
 [KasmVNC](https://kasmweb.com/kasmvnc), [QEMU](https://qemu.org)'s built-in
 server, [Selenoid](https://aerokube.com/selenoid/).
 
+QEMU is the exception: it boots firmware with no guest, so there is no
```
sibson: testing detail leaking into readme. violates readme is minimal content for new users

tests/functional/test_encodings.py
```diff
@@ -63,6 +65,57 @@ def capture(
         return Image.open(path).convert("RGB").copy(), result.stderr
 
 
+# `cls N` repaints OVMF's whole shell framebuffer in colour N.
```
sibson: should be inline rather than above

tests/functional/test_encodings.py
```diff
@@ -63,6 +65,57 @@ def capture(
...
+QEMU_SCREENS = {
+    "flat": ("cls 4",),
+    "text": ("cls 1", "echo vncdotool encoding probe"),
+}
+
+# Measured against QEMU's firmware screens the way EMITTED was.
```
sibson: what is the way EMITTED was

tests/functional/test_encodings.py
sibson: 🤖 Claude Code: Renamed. `capture_current` → `capture_qemu`, and `type_commands` folded into `draw_on_qemu`, which puts a screen up and returns the Raw capture of it. Neither takes a server, so both now say so. 28566f2

tests/functional/test_encodings.py
sibson: 🤖 Claude Code: You are right, and it was worse than duplication — my loop settled by taking repeated captures over separate connections, where `stable` watches one. Gone; the screen now comes back from a single `type ... stable 1 capture` run. 038a73a

tests/functional/test_encodings.py
sibson: 🤖 Claude Code: Extracted to `qemu_cases()`, alongside the existing scene loop. 038a73a

tests/functional/utils.py
sibson: 🤖 Claude Code: Fair. probe_context() no longer names anyone: VNCServer grew a `tls_ca` to trust and a `session` to hold open, and the function reads both off the descriptor. That also got the `is SELENOID` check out of FleetTestCase.setUp. 038a73a

tests/servers/qemu-entrypoint.sh
sibson: 🤖 Claude Code: Cut the first sentence; the flags below it said the same thing. What is left is the two facts the flags cannot: SeaBIOS blinks a cursor, and OVMF PXE-loops without -net none. 038a73a

README.md
sibson: 🤖 Claude Code: Agreed, removed. specs/testing-framework.md already carries it, and the list header already points there. 038a73a

tests/functional/test_encodings.py
sibson: 🤖 Claude Code: Moved inline onto the `flat` entry. 28566f2

tests/functional/test_encodings.py
sibson: 🤖 Claude Code: Bad cross-reference on my part — a reader should not have to go and read another comment. Now says it: "Offered one at a time; QEMU sent these and answered with Raw for the rest." 28566f2

tests/functional/test_encodings.py
sibson: 🤖 Claude Code: Leaving this one open, since it is your call rather than a fix.

What it proves: that QEMU decodes hextile, zrle and tight to the same pixels as its own Raw — plus, for those three, that a rectangle in the asked-for encoding actually arrived. QEMU's Tight is its own implementation, not LibVNCServer's or TigerVNC's, so it is the one encoder in the fleet nothing else covers.

What it does not prove: that Raw is right. The scene servers hold a capture against a committed PNG; QEMU has no desktop, so its oracle is our own Raw decoder. Weaker, and specs/testing-framework.md says so.

So the honest answer to "do we need to prove it" is: it is the only rendering coverage QEMU can have short of the gnu-efi payload in the spec backlog, and if that reads as not worth 14 cases, dropping the class costs nothing else in the PR. Say the word.

sibson: 🤖 Claude Code: You are right, and my previous answer had it backwards — I argued this covers QEMU's encoder, which is not ours to prove.

The reason to keep it is decoder variety: QEMU's Hextile, ZRLE and Tight are written from the specification rather than shared with the rest of the fleet, so the bytes on the wire are a variation the decoders are otherwise never held against. Everything else in the grid is TigerVNC, x11vnc, wayvnc or LibVNCServer-derived.

The QEMU_EMITTED guard serves the same end: without it a Raw answer would pass the comparison while the decoder under test never ran.

Reframed in specs/testing-framework.md at 424899e; no test change.

tests/functional/test_encodings.py
```diff
@@ -98,8 +136,46 @@ def test_renders_the_scene_through_the_encoding_it_asked_for(self) -> None:
...
+                name, (RendersTheSameScreenAsRaw, FleetTestCase),
+                {"encoding": encoding, "screen": screen},
+            )
+            yield case("test_renders_the_screen_as_raw_renders_it")
+
+
 def load_tests(loader: unittest.TestLoader, tests: unittest.TestSuite, pattern: object) -> unittest.TestSuite:
     suite = unittest.TestSuite()
+    suite.addTests(qemu_cases())
     for server in SCENE_SERVERS:
```
sibson, line 165: this should become a function too following the new pattern

tests/functional/utils.py
```diff
@@ -461,9 +485,25 @@ def connect(server: VNCServer, timeout: Optional[float] = None) -> api.ThreadedV
...
+    """What a bare `vncdo capture` of this server needs around it."""
+    env = {"SSL_CERT_FILE": str(server.tls_ca)} if server.tls_ca else None
+    if server.session is None:
+        yield env
+    else:
+        with server.session():
+            yield env
+
+
+def capture_screenshot(
```
sibson, line 499: why did we need a standalone capture_qemu() if this existins?

tests/functional/test_encodings.py
sibson: 🤖 Claude Code: Done — `scene_cases()` alongside `qemu_cases()`, and `load_tests` is now just the two `addTests` calls. ab17103

tests/functional/utils.py
sibson: 🤖 Claude Code: We did not. The one I should have pointed at is this file's own `capture()` a few lines up, not `capture_screenshot()` — `capture_qemu()` was that function with different arguments in front of the capture: same subprocess, same `-v -v` to read rectangle encodings back out of the log, same decoded image and stderr returned. Folded into it; `capture()` now takes the commands to run first, so the scene grid passes its key and expect and the QEMU cases pass their shell commands and stable. ab17103

`capture_screenshot()` in utils.py stays separate on purpose: it takes a path and returns a path, and the readiness gate and the gallery have no use for the log. That is the only reason the grids do not call it.

tests/functional/test_encodings.py
```diff
@@ -63,6 +64,29 @@ def capture(
         return Image.open(path).convert("RGB").copy(), result.stderr
 
 
+QEMU_SCREENS = {
+    "flat": ("cls 4",),  # `cls N` repaints OVMF's whole shell framebuffer in colour N
+    "text": ("cls 1", "echo vncdotool encoding probe"),
+}
+
+# Offered one at a time; QEMU sent these and answered with Raw for the rest.
+QEMU_EMITTED = {"raw", "hextile", "zrle", "tight"}
```
sibson: QEMU_ENCODINGS, emited is very unclear and drop th ecomment

tests/functional/test_encodings.py
```diff
@@ -98,8 +124,45 @@ def test_renders_the_scene_through_the_encoding_it_asked_for(self) -> None:
...
-def load_tests(loader: unittest.TestLoader, tests: unittest.TestSuite, pattern: object) -> unittest.TestSuite:
-    suite = unittest.TestSuite()
+class RendersTheSameScreenAsRaw:
+    """One encoding against QEMU's own Raw, on a screen the shell was told to draw."""
+
+    server = QEMU
+    encoding: str
+    screen: str
+
+    def test_renders_the_screen_as_raw_renders_it(self) -> None:
```
sibson: test_renders_the_screen_as_raw_renders_it -> test_render_matches_raw

tests/functional/test_encodings.py
```diff
@@ -98,8 +124,45 @@ def test_renders_the_scene_through_the_encoding_it_asked_for(self) -> None:
         )
 
 
-def load_tests(loader: unittest.TestLoader, tests: unittest.TestSuite, pattern: object) -> unittest.TestSuite:
-    suite = unittest.TestSuite()
+class RendersTheSameScreenAsRaw:
```
sibson: Not keen on this name.

tests/functional/test_encodings.py
sibson: 🤖 Claude Code: `QEMU_ENCODINGS`, comment dropped. The old name said how the set was arrived at rather than what it holds. 9327d8f

tests/functional/test_encodings.py
sibson: 🤖 Claude Code: Renamed. 9327d8f

tests/functional/test_encodings.py
sibson: 🤖 Claude Code: `RenderMatchesRaw`, to match the method. A failing case now reads `TestMatchesRaw_qemu_hextile_screen_flat`. Say if you would rather it went another way. 9327d8f

PR #508, opened 2026-09-09
Keep the mouse pointer out of captures by default

sibson:
`vncdo` now always offers the Cursor pseudo-encoding and, by default, discards the shape instead of drawing it — a capture no longer contains a pointer whose position no script controls. `--cursor {none,server,local}` chooses; `server` restores the old painted behaviour, `local` draws it client-side. `--localcursor`/`--nocursor` still work as aliases. Design and measurements in `specs/cursor.md`.

Not obvious from the diff: `CursorMode` is a plain `Enum`, not `str`, so `api.connect(cursor=...)` coerces explicitly; `--cursor` defaults to `None` rather than to the `omit` mode, so contradicting it with an alias errors instead of silently losing. UltraVNC is marked `paints_pointer` — offered both Cursor and PointerPos it still composites a 12x19 arrow, so its check runs in reverse.

The Windows job clears the desktop first: the agent's console and the desktop icons sit under the probe box and move a capture for reasons that are not the pointer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #514, opened 2026-09-10
Stop the pointer test blaming the server for an unpainted desktop

sibson:
`test_capture_does_not_depend_on_where_the_pointer_is` reported "the server is painting the pointer into the framebuffer" twice for things that were not a pointer: a capture 99% black because the desktop had not painted yet, and a desktop gradient still fading in. Counting colours misses both — an unpainted desktop has hundreds of colours in it.

Two guards. A capture at least 95% its commonest colour fails saying so and names the file. And a third capture at the same pointer position as the first is a control: if the region moves between those two, the desktop is animating there and no claim about the pointer can be made. A new desktop is unsteady briefly, so the triple is re-taken up to three times.

`stable` is no use here — these desktops never go quiet, and waiting took the Windows job from 36s to 230s.

Stacked on #508, whose test this fixes.


tests/functional/test_cursor.py
```diff
@@ -71,7 +73,16 @@ def at(self, tag: str, position: Tuple[int, int], *flags: str) -> Image.Image:
             f"{result.returncode}, stderr:\n{result.stderr}",
         )
         with Image.open(png) as image:
-            return image.convert("RGB").copy()
+            capture = image.convert("RGB").copy()
+
+        blank = blank_fraction(capture)
+        self.assertLess(
```
sibson, line 79: this belongs in the probe_context or other similar wait until server is ready?

tests/functional/utils.py
```diff
@@ -515,6 +515,25 @@ def has_expected_content(server: VNCServer, colours: Optional[int]) -> bool:
     return colours != 1
 
 
+# A desktop that has not finished painting is almost entirely one colour but
+# not entirely, so counting colours does not catch it. UltraVNC's first
+# capture after connecting came back 99% black with 700 colours in it.
+BLANK_FRACTION = 0.95
```
sibson, line 562: should we just use the stable command in vncdo?

tests/functional/test_cursor.py
sibson: 🤖 Claude Code: Moved into `wait_until_ready` in 5fabbe2, which is the right home — its docstring already says readiness has to be a capture because servers accept connections before they have anything to show. It now also rejects a capture that is almost entirely one colour, which is the case a colour count misses (the 99%-black one had 700 colours in it). That let the per-capture guard and the warm-up capture both go.

tests/functional/utils.py
sibson: 🤖 Claude Code: Yes — done in 5fabbe2. I had tried `stable` and wrongly concluded it was unusable: I gave it the default fuzz, which for a 32bpp format means exact, and a desktop gradient drifting by one or two per channel never settles exactly. That is why it ran to the timeout and took the Windows job from 36s to 230s. `stable 1 64` settles fine.

The control capture and the retry loop are both gone as a result. The same fuzz also replaces exact equality in the comparison, which was needed anyway: two captures of a settled Windows desktop differ by up to 36 per channel where the server re-encoded the region the pointer left, while a composited pointer differs by 209 and 243. Measured against the CI artifacts, 64 passes TightVNC and TigerVNC and still fails UltraVNC, which is genuinely painting.

PR #515, opened 2026-09-10
Verify basic function against every supported server

sibson:
The per-server grid ran against TigerVNC and libvncserver-example alone, so no test had ever sent a pointer event to KasmVNC, Selenoid or wayvnc. It now runs against `SUPPORTED_SERVERS`, one entry per product in README's Supported class, so the set we claim and the set we test are the same list.

Input cases no longer trust the exit status: vncdo returns before the server reacts, so `move 10 10` exits 0 against a server that then drops the session. Each input case uses the connection again afterwards. Clicking is verified rather than survived, the scene player mapping a click to a scene as it already maps a keypress.

KasmVNC fails `test_mousemove`, `test_click` and the click case here. That is the defect the widened grid was meant to find, so **this PR is red on kasmvnc by design**; #512 stacks on this branch and fixes it.

### DIFF
--- a/tests/functional/utils.py
+++ b/tests/functional/utils.py
@@ -191,7 +191,6 @@ QEMU_TLS = VNCServer(
     tls_ca=QEMU_TLS_CA,
 )
 
-
 @contextlib.contextmanager
 def selenoid_session() -> Iterator[None]:
     endpoint = f"http://{HOST}:{SELENOID.port}/wd/hub/session"
@@ -240,6 +239,20 @@ WEBSOCKET_SERVERS = [QEMU, QEMU_TLS, SELENOID, KASMVNC]
 # format grids reach them over ws:// and the handshake is exercised by every
 # case rather than by a test of its own.
 SCENE_SERVERS += [SELENOID, KASMVNC]
+
+# One entry per product in README.md's "Supported" class, which is the claim
+# this grid tests. Wider than SCENE_SERVERS, because a server can answer
+# input without being able to display a scene: libvncserver-example has no X
+# server, and qemu shows firmware.
+SUPPORTED_SERVERS = [
+    TIGERVNC,
+    X11VNC,
+    WAYVNC,
+    LIBVNCSERVER_EXAMPLE,
+    KASMVNC,
+    QEMU,
+    SELENOID,
+]
 
 
 # An event sink rather than a rendering server, so it stays out of the smoke
@@ -713,7 +726,7 @@ def start_replay_server(
     testcase.fail(f"vncdo-replay --server never listened on {port}: {server.communicate()[1]}")
 
 
-class _VNCServerTestMixin:
+class _VNCServerVerificationTestMixin:
     """Shared test body, parameterized per-server by register_server_tests().
 
     Deliberately does NOT subclass TestCase, or `unittest discover` would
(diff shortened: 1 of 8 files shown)


tests/functional/utils.py
```diff
@@ -188,6 +188,20 @@ def vnclog_can_reach(server: VNCServer) -> bool:
 # case rather than by a test of its own.
 SCENE_SERVERS += [SELENOID, KASMVNC]
 
+# One entry per product in README.md's "Supported" class, which is the claim
```
sibson: narrative

tests/functional/utils.py
```diff
@@ -673,7 +687,7 @@ def start_replay_server(
     testcase.fail(f"vncdo-replay --server never listened on {port}: {server.communicate()[1]}")
 
 
-class _VNCServerTestMixin:
+class _VNCServerVerificationTestMixin:
     """Shared test body, parameterized per-server by register_server_tests().
 
     Deliberately does NOT subclass TestCase, or `unittest discover` would
```
sibson, line 729: unneeded

tests/goldens/click_targets.py
```diff
@@ -0,0 +1,26 @@
+"""Imports nothing, so the scene player can run this arithmetic inside the
```
sibson: unhelpful and suggests a bad design choice to duplicate things.

tests/goldens/scene_player.py
```diff
@@ -9,6 +9,9 @@
 from PIL import Image
 from Xlib import X, display
 
+sys.path.insert(0, str(Path(__file__).resolve().parents[2]))
+from tests.goldens import click_targets  # noqa: E402
```
sibson: noqa suggests we are doing something wrong here

tests/goldens/scene_player.py
```diff
@@ -33,10 +36,11 @@ def __init__(self, scene_dir: Path) -> None:
             X.InputOutput, X.CopyFromParent,
             background_pixel=screen.black_pixel,
             override_redirect=True,
-            event_mask=X.ExposureMask,
+            event_mask=X.ExposureMask | X.ButtonPressMask,
         )
-        # Without this, key events never reach the container's `xev -root` sink.
-        screen.root.change_attributes(event_mask=X.KeyPressMask)
+        # Without this, key and button events never reach the container's
```
sibson, line 40: is this discoverable by reading docs?

tests/functional/utils.py
sibson: 🤖 Claude Code: Deleted -- design rationale for why SUPPORTED_SERVERS is wider than SCENE_SERVERS belongs in the commit body, not here. eaa2d84.

tests/functional/utils.py
sibson: 🤖 Claude Code: Reverted -- back to _VNCServerTestMixin. eaa2d84.

tests/goldens/click_targets.py
sibson: 🤖 Claude Code: Rewrote the header -- it should have said the module exists so both sides share one mapping, not that they duplicate it. eaa2d84.

tests/goldens/scene_player.py
sibson: 🤖 Claude Code: Moved the click_targets import into handle_button, where it's actually used, so no top-level noqa is needed. eaa2d84.

tests/goldens/scene_player.py
sibson: 🤖 Claude Code: No -- this isn't in the X11 protocol spec or Xlib docs. It's specific to this container: there's no window manager, so the override-redirect scene_player window never gets input focus, and key/button events land on the root window instead. Verified empirically against this fleet, not cited from a doc. Leaving this comment as-is since it names a real, non-obvious constraint; happy to trim it if you'd rather it read differently.

PR #517, opened 2026-09-11
Boot the native QEMU/KVM test server on OVMF

sibson:
#513 gave the Docker fleet's QEMU `-bios` OVMF and `-net none`. The native server the "Linux - QEMU/KVM" job starts raw on the runner is a different server and kept SeaBIOS, whose blinking cursor and boot-device cycling make two captures of an idle machine differ everywhere — currently failing `test_capture_does_not_depend_on_where_the_pointer_is`.

The firmware path is probed rather than hardcoded: bookworm's `ovmf` ships `OVMF_CODE.fd`, Ubuntu 24.04's ships only `OVMF_CODE_4M.fd`.

I can't run QEMU/KVM on macOS, so the change is unverified outside CI; the "Linux - QEMU/KVM" job is the check. I did drive `install_qemu` against stubbed candidate paths to confirm all three branches (both names, one name, neither) behave.

### DIFF
--- a/tests/servers/qemu-kvm/setup.sh
+++ b/tests/servers/qemu-kvm/setup.sh
@@ -21,6 +21,13 @@ ACCEL="${VNCDOTOOL_QEMU_ACCEL:-kvm}"
 MEMORY="${VNCDOTOOL_QEMU_MEMORY:-256}"
 
 QEMU=qemu-system-x86_64
+# Ubuntu 24.04's ovmf package ships only the 4M name; Debian bookworm, which
+# the Docker fleet builds on, ships only the plain one.
+OVMF_CANDIDATES=(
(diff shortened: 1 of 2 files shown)


tests/servers/qemu-kvm/setup.sh
```diff
@@ -79,8 +112,11 @@ start_qemu() {
     echo "--- starting $QEMU on display :$DISPLAY_NUMBER with -accel $ACCEL"
     mkdir -p "$RUNTIME_DIR"
     rm -f "$PIDFILE"
-    # No -drive and no -kernel: with nothing to boot, the machine sits at the
-    # firmware's "no bootable device" screen, which is a real framebuffer.
+    # No -drive and no -kernel: with nothing to boot, the firmware stops on
```
sibson, line 115: does this need to be so long

PR #520, opened 2026-09-12
Load OVMF on the native QEMU/KVM server via pflash, not -bios

sibson:
Main's `Linux - QEMU/KVM` job has failed on every merge since #517: `-bios OVMF_CODE_4M.fd` dies with `could not load PC BIOS`. Ubuntu noble's `ovmf` package (2024.02-1+) only ships the split 4M firmware image, which `-bios` can't parse — it expects a legacy monolithic image. QEMU only accepts the 4M variant as a pflash drive.

Confirmed root cause from the `qemu-kvm-diagnostics` artifact on the failing run (`qemu.log`: `could not load PC BIOS`), and verified the `-drive if=pflash,format=raw,readonly=on` fix by reproducing the same invocation locally with Homebrew's qemu and edk2 firmware — it boots and the process stays alive.

Also drops the `OVMF_CODE.fd` (2M) candidate path, since CI only runs on `ubuntu-latest` and that path was dead code there.

### DIFF
--- a/tests/servers/qemu-kvm/README.md
+++ b/tests/servers/qemu-kvm/README.md
@@ -29,10 +29,6 @@ Why this server is raw rather than a container, when Tier 1 already has a
   so two captures of an idle machine differ. Without `-net none` OVMF
   retries PXE forever and the screen scrolls instead.
 
-* **OVMF loads as pflash, not `-bios`.** Ubuntu noble's `ovmf` package only
-  ships the split 4M `OVMF_CODE_4M.fd`, which `-bios` can't parse.
-  `setup.sh` loads it with `-drive if=pflash,readonly=on` instead.
-
 * **`-vnc :0` listens on every interface.** QEMU's VNC server has no
   authentication unless it is started with `password=on` *and* a password is
   then set over the monitor, so the listen address is the only thing
--- a/tests/servers/qemu-kvm/README.md
+++ b/tests/servers/qemu-kvm/README.md
@@ -29,6 +29,10 @@ Why this server is raw rather than a container, when Tier 1 already has a
   so two captures of an idle machine differ. Without `-net none` OVMF
   retries PXE forever and the screen scrolls instead.
 
+* **OVMF loads as pflash, not `-bios`.** Ubuntu noble's `ovmf` package only
+  ships the split 4M `OVMF_CODE_4M.fd`, which `-bios` can't parse.
+  `setup.sh` loads it with `-drive if=pflash,readonly=on` instead.
+
 * **`-vnc :0` listens on every interface.** QEMU's VNC server has no
   authentication unless it is started with `password=on` *and* a password is
   then set over the monitor, so the listen address is the only thing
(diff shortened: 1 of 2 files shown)


tests/servers/qemu-kvm/README.md
```diff
@@ -29,6 +29,10 @@ Why this server is raw rather than a container, when Tier 1 already has a
   so two captures of an idle machine differ. Without `-net none` OVMF
   retries PXE forever and the screen scrolls instead.
 
+* **OVMF loads as pflash, not `-bios`.** Ubuntu noble's `ovmf` package only
```
sibson: uneeded, this is documented elsewhere

tests/servers/qemu-kvm/setup.sh
```diff
@@ -21,11 +21,10 @@ ACCEL="${VNCDOTOOL_QEMU_ACCEL:-kvm}"
 MEMORY="${VNCDOTOOL_QEMU_MEMORY:-256}"
 
 QEMU=qemu-system-x86_64
-# ovmf ships OVMF_CODE_4M.fd on Ubuntu 24.04 and OVMF_CODE.fd on Debian
-# bookworm.
+# Ubuntu noble's ovmf package (2024.02-1 onward) dropped the legacy 2M
```
sibson: we don't need to describe history.

tests/servers/qemu-kvm/setup.sh
```diff
@@ -21,11 +21,10 @@ ACCEL="${VNCDOTOOL_QEMU_ACCEL:-kvm}"
...
+# OVMF_CODE.fd; CI runs on ubuntu-latest, so only the 4M path matters here.
 OVMF_CANDIDATES=(
```
sibson: we don't need Candidates if there is only one.  Let's remove all this selcetion code.

tests/servers/qemu-kvm/setup.sh
```diff
@@ -121,13 +120,15 @@ start_qemu() {
     #
     # The 127.0.0.1: prefix is load-bearing. A bare -vnc :0 listens on every
     # interface, and this server has no authentication at all.
+    # -bios rejects the split 4M OVMF image; it wants pflash, and a
```
sibson: narrative

tests/servers/qemu-kvm/README.md
sibson: 🤖 Claude Code: Removed — dropped in f553c7c.

tests/servers/qemu-kvm/setup.sh
sibson: 🤖 Claude Code: Removed the history comment (and the OVMF_CANDIDATES machinery it was attached to) in f553c7c.

tests/servers/qemu-kvm/setup.sh
sibson: 🤖 Claude Code: Removed find_ovmf/OVMF_CANDIDATES; BIOS is now a hardcoded path in f553c7c.

tests/servers/qemu-kvm/setup.sh
sibson: 🤖 Claude Code: Removed in f553c7c.

PR #521, opened 2026-09-13
Wait for a non-incremental refresh to cover the framebuffer

sibson:
`capture` has saved part-black screenshots since 1.3.0: `commitUpdate` fires on the first framebuffer update that painted anything, but RFC 6143 §7.5.3 lets a server answer a non-incremental request across several messages. A non-incremental refresh now tracks coverage in a 1-bit Pillow mask, asks once more when the screen is short, then saves with a warning naming the unpainted pixels rather than erroring or waiting on a deadline.

Coverage is counted in `updateRectangle`/`copyRectangle`, not from `rectanglePos` — that list carries the Cursor pseudo-rectangle, whose x/y are a hotspot rather than a region of the framebuffer.

Measured against the docker fleet before writing any of it: tigervnc, x11vnc, libvncserver, kasmvnc and qemu each answer with one update at 100% coverage, in raw, tight, hextile and zrle alike. No fleet server splits an answer, so the path this fixes has unit tests alone; UltraVNC, which does split, is Windows-only. Unblocks #508.


vncdotool/client.py
```diff
@@ -141,6 +141,46 @@ def _settle(self) -> None:
...
+
+    def __init__(self, width: int, height: int) -> None:
+        self.asked = 0
+        self.restart(width, height)
+
+    @property
+    def complete(self) -> bool:
+        return self.unpainted is None or self.unpainted.getbbox() is None
+
+    def painted(self, x: int, y: int, width: int, height: int) -> None:
```
sibson: Would this read better as paintRect(),

vncdotool/client.py
```diff
@@ -141,6 +141,46 @@ def _settle(self) -> None:
...
+        if self.unpainted is None or width <= 0 or height <= 0:
+            return
+        # Image.paste clips a box that runs off the mask, so a rectangle
+        # reaching past the framebuffer marks only what is on it.
+        self.unpainted.paste(0, (x, y, x + width, y + height))
+
+    def restart(self, width: int, height: int) -> None:
```
sibson: let's put this above painted, so it's in logical call order

vncdotool/client.py
```diff
@@ -141,6 +141,46 @@ def _settle(self) -> None:
...
+    def __init__(self, width: int, height: int) -> None:
+        self.asked = 0
+        self.restart(width, height)
+
+    @property
+    def complete(self) -> bool:
+        return self.unpainted is None or self.unpainted.getbbox() is None
+
+    def painted(self, x: int, y: int, width: int, height: int) -> None:
+        if self.unpainted is None or width <= 0 or height <= 0:
```
sibson: where does unpainted get initialized?

vncdotool/client.py
sibson: Do we need a restart or should the client just create a new class instance?

vncdotool/client.py
```diff
@@ -141,6 +141,46 @@ def _settle(self) -> None:
...
+
+    def restart(self, width: int, height: int) -> None:
+        self.size = (width, height)
+        self.unpainted: Image.Image | None = (
+            Image.new("1", (width, height), 1) if width > 0 and height > 0 else None
+        )
+
+    def shortfall(self) -> str:
+        width, height = self.size
+        unpainted = sum(self.unpainted.histogram()[1:]) if self.unpainted else 0
```
sibson: ternary is hard to read here

vncdotool/client.py
```diff
@@ -141,6 +141,46 @@ def _settle(self) -> None:
...
+        # reaching past the framebuffer marks only what is on it.
+        self.unpainted.paste(0, (x, y, x + width, y + height))
+
+    def restart(self, width: int, height: int) -> None:
+        self.size = (width, height)
+        self.unpainted: Image.Image | None = (
+            Image.new("1", (width, height), 1) if width > 0 and height > 0 else None
+        )
+
+    def shortfall(self) -> str:
```
sibson: it's weird this is a string results.  should this just be __str__()

vncdotool/client.py
```diff
@@ -141,6 +141,46 @@ def _settle(self) -> None:
...
+        width, height = self.size
+        unpainted = sum(self.unpainted.histogram()[1:]) if self.unpainted else 0
+        return (
+            f"the server left {unpainted} of the {width}x{height} framebuffer "
+            f"unpainted after {self.asked + 1} full-screen update requests; "
```
sibson: asked only exists for this message? that seems like extra unneeded state.

vncdotool/client.py
```diff
@@ -571,15 +617,27 @@ def copyRectangle(
             return
         region = self.screen.crop((srcx, srcy, srcx + width, srcy + height))
         self.screen.paste(region, (x, y))
+        self._painted(x, y, width, height)
         self.drawCursor()
 
+    def _painted(self, x: int, y: int, width: int, height: int) -> None:
```
sibson: almost shadows another method, confusing

vncdotool/client.py
sibson: 🤖 Claude Code: Renamed to paintRect() in c788967.

vncdotool/client.py
sibson: 🤖 Claude Code: restart() is gone in c788967, so there is nothing left to order — the mask is built in __init__ and a desktop resize gets a new instance.

vncdotool/client.py
sibson: 🤖 Claude Code: It was in restart(), which __init__ called — which is exactly why it was hard to find. In c788967 it is assigned in __init__ and nowhere else.

vncdotool/client.py
sibson: 🤖 Claude Code: A new instance. updateDesktopSize now does `self._refresh = _FullRefresh(width, height)`, and restart() is deleted. Side effect worth noting: the re-request allowance resets with the resize, which seems right for a screen that just changed size.

vncdotool/client.py
sibson: 🤖 Claude Code: Both ternaries are gone. A zero-sized Pillow image behaves as an already-finished mask — getbbox() returns None and paste is a no-op — so the mask never needs to be Optional, which removed the `if self.unpainted else 0` here and the `if width > 0 and height > 0` one in the constructor, along with the None check at the top of every method. c788967.

vncdotool/client.py
sibson: 🤖 Claude Code: Agreed — it is __str__ now, and commitUpdate logs the object rather than calling a method for its text. c788967.

vncdotool/client.py
sibson: 🤖 Claude Code: Not only for the message — commitUpdate counts it against REFRESH_REREQUESTS to decide whether to ask again or give up, so the state has to live somewhere, and it has to reset per refresh, which is what this object is. I renamed it `rerequests` in c788967 so the field says what it counts. Leaving this open in case you want it moved onto the client instead.

vncdotool/client.py
sibson: 🤖 Claude Code: Renamed to _markPainted() in c788967 — it read as a predicate and sat next to _FullRefresh.painted(), which is now paintRect(), so neither name is close to the other any more.

sibson:
🤖 Claude Code: Timed the +83 `_markPainted` calls the bot flagged. `make bench` on `tigervnc-raw-bgrx8888`, 300 runs, three repetitions each: this branch 801.2 / 799.9 / 813.1 us median, `origin/main` 812.0 / 802.8 / 805.5 us. The extra call per rectangle is one `Image.paste` of a solid colour into a 1-bit mask, and it does not show above run-to-run noise. Not recorded to `bench.jsonl` — same machine, no real delta to pin.

vncdotool/client.py
```diff
@@ -141,6 +141,38 @@ def _settle(self) -> None:
...
+    FramebufferUpdate messages as it likes, so one message arriving does not
+    mean the area has been sent.
+    """
+
+    def __init__(self, width: int, height: int) -> None:
+        self.rerequests = 0
+        self.unpainted = Image.new("1", (width, height), 1)
+
+    def paintRect(self, x: int, y: int, width: int, height: int) -> None:
+        # Image.paste clips a box that runs off the mask, so a rectangle
```
sibson: narrative

vncdotool/client.py
```diff
@@ -141,6 +141,38 @@ def _settle(self) -> None:
         self.result.callback(self.client)
 
 
+class _FullRefresh:
```
sibson: this is not a good name

vncdotool/client.py
```diff
@@ -562,6 +599,7 @@ def updateRectangle(
         else:
             self.screen.paste(update, (x, y))
 
+        self._markPainted(x, y, width, height)
```
sibson: method call feels like overhead here

vncdotool/client.py
```diff
@@ -571,15 +609,27 @@ def copyRectangle(
...
+            self._refresh.paintRect(x, y, width, height)
+
     def commitUpdate(self, rectangles: list[tuple[int, int, int, int]] | None = None) -> None:
         if self.deferred:
             if not rectangles or self.screen is None:
                 # No rectangle in this update painted self.screen; wait for
                 # one that does before completing the refresh.
                 self.framebufferUpdateRequest()
                 return
+            if self._refresh is not None and not self._refresh.complete:
```
sibson: would it be easier to always have a refresh object?  _refresh name also doesn't seem quite right

vncdotool/client.py
```diff
@@ -155,6 +187,9 @@ class VNCDoToolClient(rfb.RFBClient):
     _raw_mode_format: rfb.PixelFormat | None = None
     _raw_mode = ""
     deferred: Deferred | None = None
+    _refresh: _FullRefresh | None = None
+
+    REFRESH_REREQUESTS = 1
```
sibson: Needs a better name for readability

vncdotool/client.py
sibson: 🤖 Claude Code: Cut to the one fact from outside the file — that Image.paste clips — in d2cd4a9. What that means for a rectangle reaching past the framebuffer is visible in the call on the next line.

vncdotool/client.py
sibson: 🤖 Claude Code: It named the caller's operation rather than its own contents. _ScreenCoverage in d2cd4a9.

vncdotool/client.py
sibson: 🤖 Claude Code: MAX_REFRESH_RETRIES now, with the field counting against it renamed `retries`. d2cd4a9.

vncdotool/client.py
sibson: 🤖 Claude Code: Renamed to self._coverage in d2cd4a9. The always-have-one half I have not done, and I think it collides with something — leaving this open for your call.

An always-present object has to hold a Pillow image from the moment the client exists, and client.py is written to import and connect without PIL (the comment above the import, and `key`/`type`/`press`/`move` are meant to work there). `Image.new()` in connectionMade, or a module-level shared instance as the class default, both raise ImportError on that path. Creating it only in _requestRefresh is what keeps PIL confined to the commands that already need it, and that is what makes it Optional.

The alternative I can see: a zero-sized coverage for the incremental case, since a 0x0 mask is already complete — `getbbox()` is None and paste is a no-op. That makes it always-present *within* a refresh, but the class attribute would still start as None for a client that has not refreshed yet. Happy to do that if it reads better to you.

vncdotool/client.py
sibson: 🤖 Claude Code: Measured rather than guessed — this is the call the decode-counts bot flagged as +83. `make bench` on tigervnc-raw-bgrx8888, 300 runs, three repetitions: this branch 801.2 / 799.9 / 813.1 us median, origin/main 812.0 / 802.8 / 805.5 us. It does not show above run-to-run noise.

Leaving it open because the shape is tied to the thread above: _markPainted exists to keep the `is not None` check in one place instead of repeating it in updateRectangle and copyRectangle. If the coverage object becomes always-present, this collapses to a direct `self._coverage.paintRect(...)` at both call sites and the forwarder goes away. Say the word on that one and this follows.

vncdotool/client.py
sibson: 🤖 Claude Code: You are right and my objection was wrong on the facts — Pillow is a hard dependency in pyproject.toml, not an extra, so the _RuntimeImportError shim guards a path no supported install reaches. Done in 5f84ae6.

The client builds one in __init__ and replaces rather than clears it. A coverage promised no area — an incremental refresh, or none in flight — is `_ScreenCoverage.satisfied()`, a zero-sized mask that getbbox() already reports as covered and that paste cannot dirty. `pending` tells updateDesktopSize whether a full-screen refresh is riding on the current mask, which is what decides whether a resize re-arms it.

vncdotool/client.py
sibson: 🤖 Claude Code: Gone in 5f84ae6 — with coverage always present, updateRectangle and copyRectangle call self._coverage.paintRect() straight and the forwarder is deleted.

vncdotool/client.py
```diff
@@ -141,6 +141,49 @@ def _settle(self) -> None:
         self.result.callback(self.client)
 
 
+class _ScreenCoverage:
+    """What a non-incremental refresh is still waiting to be sent.
```
sibson: awkward.  rename class _FullScreenRecieved?  removes need for most of the comment?

sibson: 🤖 Claude Code: Taken, as `_FullScreenReceived` (spelling — i before e), reached as `self._fullscreen`. In f4faa20. You were right that it absorbs the comment: the docstring keeps only the fact from outside the file, that RFC 6143 §7.5.3 lets one non-incremental answer span several FramebufferUpdate messages. What follows from that is what the code does.

Unrelated: the functional job went red on 5f84ae6. It was a flake — 14 failures, all of them x11vnc timing out, which test_pixel_format multiplies into one case per format and scene. Re-ran the same SHA and it is green, no code change.

vncdotool/client.py
```diff
@@ -276,9 +320,16 @@ def captureRegion(
 
     def refreshScreen(self, incremental: bool = False) -> Deferred:
         d = self.deferred = Deferred()
-        self.framebufferUpdateRequest(incremental=incremental)
+        self._requestRefresh(incremental)
         return d
 
+    def _requestRefresh(self, incremental: bool) -> None:
+        if incremental:
```
sibson, line 339: this seems opposite

vncdotool/client.py
```diff
@@ -580,6 +631,13 @@ def commitUpdate(self, rectangles: list[tuple[int, int, int, int]] | None = None
                 # one that does before completing the refresh.
                 self.framebufferUpdateRequest()
                 return
+            if not self._fullscreen.complete:
+                if self._fullscreen.retries < self.MAX_REFRESH_RETRIES:
+                    self._fullscreen.retries += 1
```
sibson: incrementing outside the class breaks encapsulation.

sibson: 🤖 Claude Code: Agreed. `retry()` now answers whether the server is worth asking again and counts the attempt itself, so the field is never touched from outside, and MAX_RETRIES moved off the client onto the class that reads it. commitUpdate is just `if self._fullscreen.retry():`. 351e0ed.

This also answers the older thread about that field being unneeded state — it is now entirely internal.

vncdotool/client.py
sibson: 🤖 Claude Code: Settled by 351e0ed via your encapsulation comment: the counter is now private to _FullScreenReceived, incremented only by retry(), with MAX_RETRIES alongside it. Nothing outside the class reads or writes it.

vncdotool/client.py
sibson: 🤖 Claude Code: I think what read backwards was the constructor: `_FullScreenReceived(w, h)` says the full screen was received at the exact moment nothing has been. Named constructors in 351e0ed, so each branch says which way round it is:

    if incremental:
        self._fullscreen = _FullScreenReceived.satisfied()
    else:
        self._fullscreen = _FullScreenReceived.awaiting(self.width, self.height)

The logic itself was the right way round — incremental is owed no area, non-incremental is owed the screen. Leaving this open in case you meant something else by "opposite"; say which and I will take it.

Issue #206, opened 2021-12-22
ineffective nocursor argument


Willard-Robbins:
vncdotool version 1.0.0
python  version 3.10

The VNC server is an old binary. I believe it came from Real VNC. 

I'm passing the output screenshot of vncdotool to OpenCV to confirm on-screen UI elements. I would like the screenshot to exclude the cursor so that it does not interfere with template matching. Despite using the nocursor argument, my screenshots still contain the cursor... Am I missing something? 

```
    vnc_client = api.connect(host, password=None)
    vnc_client.timeout = 10
    vnc_client.nocursor = True  # Appears to be ineffective. Want to remove the cursor from screenshots to prevent occlusion.
    vnc_client.captureScreen('screenshot.png')  # Resulting screenshot contains the cursor.
```

pmhahn:
`nocursor` onyl works when client-side rendering is negotiated with the server: In that case the cursor shape and position are sent to the client, which is then supposed to merge the cursor shape into the screen before displaying it to the user.
Without the extension the server will handle that and will instead sent the screen with the cursor already merged in, which the client cannot undo.
Please check if your server does the merging — in which case this issue would be invalid — or the client is supposed to do it. Sadly there is no easy way in `vncdotool` to see, if `PSEUDO_CURSOR = -239` is negotiated: If you're experienced with `pdb` you can add a `break _handleDecodePsuedoCursor` or you might want to use `wireshark` to see, if you an _updateRectangle_ with that encoding is received by the client — or use another VNC client, which is more verbose.

pevogam:
I just want to add that before closing this issue it might be worth it to add some of @pmhahn's explanation in the documentation. At first I looked at the `nocursor` and `pseudocursor` options in https://vncdotool.readthedocs.io/en/latest/modules.html?highlight=cursor#vncdotool.client.VNCDoToolClient.cursor (since my initial guess for `nocursor` from just looking at the source code had no effect as above) but there are no docstrings explaining the meaning of either of these, even a single short sentence might be quite helpful. Then also adding the above detail could be very helpful so that other users don't have to look up the issues on Github to figure this out.

PR #522, opened 2026-09-14
Stop compositing the mouse cursor into the framebuffer

sibson:
`drawCursor()` pasted the `--cursor local` pointer into `self.screen` and
nothing restored what the previous one covered, so an incremental capture or a
live `expect` could see one arrow per position the pointer had been. The
framebuffer now holds the server's pixels alone and `_visibleScreen()`
composites the pointer onto a crop or a copy when an image is asked for.

The judgement call a reviewer would want: `expect` and `stable` composite too,
rather than comparing raw server pixels. A reference image written by
`vncdo --cursor local capture ref.png` contains the pointer, so an `expect`
against it has to be comparing against something that does. The commit body
carries the rest.

The functional fleet did not run here — Docker was unresponsive on this
machine. `tests/functional/test_cursor.py` covers this path through the CLI,
so CI is the check.


tests/unit/test_client.py
```diff
@@ -124,7 +124,7 @@ def test_updateCursor_hides_pointer_on_zero_size(self):
 
         self.assertIsNone(cli.cursor)
         self.assertIsNone(cli.cmask)
-        cli.drawCursor()
+        cli._visibleScreen()
```
sibson: visibleScreen is a bad name

vncdotool/client.py
```diff
@@ -365,16 +361,37 @@ def _requireOnScreen(self, box: tuple[int, int, int, int]) -> None:
         if box[0] < 0 or box[1] < 0 or box[2] > width or box[3] > height:
             raise RegionError(f"region {box} is not inside the {width}x{height} screen")
 
+    def _visibleScreen(
+        self, box: tuple[int, int, int, int] | None = None
+    ) -> Image.Image:
+        """The framebuffer as a capture or a comparison sees it."""
+        assert self.screen is not None
```
sibson: this is pointless

vncdotool/client.py
```diff
@@ -365,16 +361,37 @@ def _requireOnScreen(self, box: tuple[int, int, int, int]) -> None:
...
+        # Fresh every call: a caller may hold the result across the updates
+        # that paste into self.screen.
+        visible = self.screen.crop(box) if box else self.screen.copy()
+
+        if self.factory.cursor is not CursorMode.LOCAL or not self.cursor:
+            return visible
+
+        at_x, at_y = self.cursor_pos or (self.x, self.y)
```
sibson: both cursor_pos and self.x/self.y is a smell that we have two values for the same thing

vncdotool/client.py
```diff
@@ -365,16 +361,37 @@ def _requireOnScreen(self, box: tuple[int, int, int, int]) -> None:
...
+
     def _captureSave(
         self: TClient, data: object, fp: TFile, *args: int, format: str | None = None
     ) -> TClient:
         log.debug("captureSave %s", fp)
-        assert self.screen is not None
         if args:
             self._requireOnScreen(args)  # type: ignore[arg-type]
-            capture = self.screen.crop(args)  # type: ignore[arg-type]
+            capture = self._visibleScreen(args)  # type: ignore[arg-type]
```
sibson: not acceptable to have ignore

tests/unit/test_client.py
sibson: 🤖 Claude Code: Renamed to `_snapshot()` in db53703. It names the property the callers actually rely on — a fresh image, not a live view of `self.screen`, which is what `_StableWatch` needs to hold one across updates.

vncdotool/client.py
sibson: 🤖 Claude Code: Agreed — it was a no-op under `-O` and a messageless AssertionError otherwise, so it bought nothing the next line would not have raised anyway. db53703 replaces it with a `NoFramebufferError`.

Nothing live reaches it: `_expectCompare` and `_StableWatch` both guard on `self.screen`, and `_captureSave` runs as a `commitUpdate` callback, which returns early without one. The fourth caller, `VNCLoggingClient.commitUpdate`, sits behind a `capture_file` that nothing in the tree ever assigns — separate finding, not this diff. Happy to just delete the check if you would rather have nothing there; it is the one line that keeps mypy off `Image | None` without the ignore your other comment rules out.

vncdotool/client.py
sibson: 🤖 Claude Code: Right, and the `or` was the tell. Fixed in db53703: `cursor_pos` is now where the pointer is, full stop — `mouseMove` sets it beside `x`/`y`, `updatePointerPos` overwrites it, and it starts at `(x, y)` because `connectionMade` sends a pointer event there. `_snapshot` reads it with no fallback.

`self.x`/`self.y` stay, as a genuinely different quantity: where the script last aimed, and so where a later `click` goes. They diverge exactly when the desktop warps the pointer out from under the script, which is the case PointerPos exists for. Written up in `specs/cursor.md`.

vncdotool/client.py
sibson: 🤖 Claude Code: Both gone in db53703. The cause was `*args: int` on `_capture`/`_captureSave` — no signature can narrow that to the four-tuple `_requireOnScreen` and `_snapshot` want, so the ignore was standing in for a type the shape could not express. They take a box tuple instead. That also removes the `_requireOnScreen` ignore, which predates this branch.

vncdotool/client.py
sibson: 🤖 Claude Code: You are right, and my reply above argued the wrong way — I kept the split and justified it. 7a5cc9b removes it.

`cursor_pos` is gone. `self.x`/`self.y` is the pointer position, full stop: `mouseMove` sets it, `updatePointerPos` sets it, `_snapshot` draws there, `mouseDown`/`mouseUp` click there. A pointer the desktop warps takes the next click with it.

`test_a_click_still_goes_where_the_script_put_it` was asserting the behaviour you just rejected; it is now `test_a_click_goes_where_the_pointer_is`, and it checks the actual `pointerEvent` rather than the attribute. The reasoning is in the commit body and in a new `specs/cursor.md` section, including the cost — on a shared desktop a `click` after a `move` can land where something else left the pointer. Same race a human shares a mouse with, and visible in a capture, which the old behaviour was not.

vncdotool/client.py
sibson: 🤖 Claude Code: Deleted in dde7c17 — nothing there now. `self.screen` being None reaches PIL as an AttributeError, which every live caller already guarantees cannot happen. Cost is two advisory `union-attr` errors from mypy on that line (76 -> 78); no ignore.

Also renamed `_snapshot` to `_render` in the same commit, verb-first like `_requestRefresh` and `_requireOnScreen`, naming the compositing rather than the result.

specs/cursor.md
```diff
@@ -136,6 +136,48 @@ comment above TigerVNC's admits as much: "Unfortunately we can't know for
...
 enabled, neither of its two `drawCursor` paths matches.
 
+## The pointer is an overlay, not framebuffer content
+
+`self.screen` holds the server's pixels and nothing else. Under `--cursor
+local` the shape is kept as client state — `cursor`, `cmask`, `cfocus` — and
+`_render()` composites it onto a copy at the moment an image is asked for.
+`updateCursor` and `updatePointerPos` record; they do not draw.
+
+It used to paste into `self.screen` and nothing restored what the previous
```
sibson: history

specs/cursor.md
```diff
@@ -136,6 +136,48 @@ comment above TigerVNC's admits as much: "Unfortunately we can't know for
...
+default mode, which is the reason the default is `none`.
+
+## One pointer, one position
+
+`self.x`/`self.y` is where the pointer is, and it is the only answer to that
+question. `mouseMove` sets it, `updatePointerPos` sets it, whichever moved
+the pointer last. `_render` draws there and `mouseDown`/`mouseUp` click
+there.
+
+An earlier revision tracked the server's reported position separately, so
```
sibson: history

vncdotool/client.py
```diff
@@ -106,11 +106,7 @@ def start(self) -> Deferred:
         return self.result
 
     def _frame(self) -> Image.Image:
```
sibson: do we still need this function?

vncdotool/client.py
```diff
@@ -365,17 +361,35 @@ def _requireOnScreen(self, box: tuple[int, int, int, int]) -> None:
         if box[0] < 0 or box[1] < 0 or box[2] > width or box[3] > height:
             raise RegionError(f"region {box} is not inside the {width}x{height} screen")
 
+    def _render(self, box: tuple[int, int, int, int] | None = None) -> Image.Image:
+        """The framebuffer as a capture or a comparison sees it."""
+        # Fresh every call: a caller may hold the result across the updates
```
sibson: decribes code

specs/cursor.md
sibson: 🤖 Claude Code: Deleted in 689eb8c. The bug the paragraph narrated is in the commit body where it belongs; the section still says what the model is without recounting how it got there.

specs/cursor.md
sibson: 🤖 Claude Code: Rewritten in 689eb8c. Dropped the "an earlier revision..." framing and kept the reason as a rule: tracking the reported position separately would let a click land where the script last aimed, which is invisible — the capture beside it shows the pointer somewhere else. Tell me if you want that gone too and only the rule at the top of the section left standing.

vncdotool/client.py
sibson: 🤖 Claude Code: No — gone in 689eb8c. It was a one-line delegate once the cursor composite moved into `_render`; both callers now say `self.client._render(self.box)` directly.

vncdotool/client.py
sibson: 🤖 Claude Code: Deleted in 689eb8c. `crop`/`copy` already say they return a new image.

specs/cursor.md
sibson: 🤖 Claude Code: Both paragraphs gone in f55fee5. The section is the rule and its one consequence: a `click` after a `move` lands wherever something else left the pointer, and a script that wants it elsewhere moves there first. Nothing left arguing against the alternative.

vncdotool/client.py
```diff
@@ -96,7 +96,7 @@ def __init__(
 
     def start(self) -> Deferred:
         if self.client.screen is not None:
-            self.baseline = self._frame()
+            self.baseline = self.client._render(self.box)
```
sibson: calling _method breaks encapsulation/contract

sibson: 🤖 Claude Code: Fixed in 746057a, and it was worse than the one line — `_StableWatch` reached into the client three ways: `_render` for a frame, plus `client.deferred = d` beside `client._requestRefresh(...)` to ask for an update. The last two predate this PR.

`refreshScreen()` turned out to be the public spelling of that pair already — it assigns `self.deferred` and calls `_requestRefresh` — so `_request` is now one line against it, with the callback on the Deferred it returns.

`_render` becomes `render()`. A caller wanting the screen as a PIL Image is the #206 case (template matching in OpenCV), which `captureScreen` serves only through a PNG on disk. Its docstring states what it does not do: nothing is asked of the server, so `refreshScreen()` comes first when the last update is not recent enough.

`_StableWatch` now touches `client.screen`, `client.render()` and `client.refreshScreen()` and nothing else.

tests/unit/test_client.py
```diff
@@ -124,7 +124,7 @@ def test_updateCursor_hides_pointer_on_zero_size(self):
 
         self.assertIsNone(cli.cursor)
         self.assertIsNone(cli.cmask)
-        cli.drawCursor()
+        cli.render()
```
sibson: renderScreen fits existing naming conventions better?

tests/unit/test_decoder_pointer_pos.py
```diff
@@ -61,7 +65,7 @@ def test_a_move_by_the_script_supersedes_the_server(self) -> None:
 
         self.cli.mouseMove(10, 20)
 
-        self.assertIsNone(self.cli.cursor_pos)
+        self.assertEqual((self.cli.x, self.cli.y), (10, 20))
```
sibson: cli abbrv is confusing because we also have a cli compienent

tests/unit/test_client.py
sibson: 🤖 Claude Code: Agreed — `renderScreen()` in 4633eb8, beside `captureScreen`, `refreshScreen`, `expectScreen` and `stableScreen`.

One asymmetry worth your call: those four all pair with a `...Region` variant taking `x, y, w, h`, and none of the `...Screen` spellings takes a box. `renderScreen(box=None)` does. If you want the convention held exactly I would add `renderRegion(x, y, w, h)` and drop the parameter — say the word and I will; I left it as one method rather than widen the public surface unasked.

tests/unit/test_decoder_pointer_pos.py
sibson: 🤖 Claude Code: Renamed to `self.client` throughout `test_decoder_pointer_pos.py` in 4633eb8.

The abbreviation is in twelve other test modules and in `tests/unit/utils.py` (`make_client()` returns `cli`), around 700 uses. I left those alone rather than put a repo-wide mechanical rename in this diff — happy to do it as its own change if you want it swept.

tests/unit/test_client.py
sibson: 🤖 Claude Code: Split in e182d30. `renderScreen()` takes nothing, `renderRegion(x, y, w, h)` takes the four, matching `capture`/`expect`/`stable`.

`_StableWatch` held a box only to render with, so it now takes the bound method instead — `renderScreen`, or `renderRegion` under a `partial` — and no longer knows what a box is. A private `_render(box)` sits behind the pair for `_captureSave` and `_expectCompare`, which already hold a crop box in PIL's (left, upper, right, lower) form rather than x, y, w, h; converting back to w/h at those two call sites just to go through the public method would be noise.

Three tests added for the new surface: the region crops, an off-screen region raises `RegionError`, and `renderScreen()` is not `self.screen`.

PR #524, opened 2026-09-14
Cut the repeated work out of the fleet job

sibson:
Three cuts to the fleet job, measured off the CI logs rather than guessed: the
functional step is 313s and the smoke step 37s on a recent green run.

The QEMU cases each rebuilt the same Raw oracle — typing the screen, waiting a
second, capturing — so two screens cost fourteen drawings. They now share one
per screen, which is why the generation loops swapped order. The KasmVNC
behaviour lock moves out of the smoke module, since it waits out five-second
timeouts and the smoke step is paid before everything else. And the suite names
its modules so the smoke module is not discovered and run a second time.

Expect roughly 60s off ~350s.

I could not run `make test-func` — no Docker on this machine — so CI is the
first real check. Collection counts match: 242 before, 32 + 210 after.


tests/functional/test_encodings.py
```diff
@@ -76,14 +76,23 @@ def capture(
 QEMU_SETTLE_SECONDS = "1"
 
 
+# One drawing serves every encoding compared against a screen: a QEMU screen
+# stays up until something draws over it. Only the newest is held, because
+# each case captures the live screen again in its own encoding.
```
sibson: long winded

sibson: 🤖 Claude Code: Deleted it rather than trimmed it, in ee55a7c.

A shorter version would have kept the defect: the fact it carried is that a framebuffer holds what was drawn on it until something draws over it, which is a general property rather than something QEMU springs on the reader — CLAUDE.md rules that out even where the code hits one case of it. Why the cache is single-entry is already in the commit body that introduced it.

PR #527, opened 2026-09-14
Ignore a PointerPos that predates our own move

sibson:
x11vnc polls the X pointer rather than reading back the event it just acted on, so between acting on a `move` and its next poll it still holds the old position and reports that as a third-party warp. `--cursor local` drew the arrow where the pointer no longer was, and since #522 the next `click` went there too. A PointerPos arriving within `POINTER_POS_SETTLE` (1.0s) of this client's own `mouseMove` is now discarded.

Worth knowing as a reviewer: `tests/functional/test_cursor.py` was green before this change and is green after. Whether x11vnc's poll lags at all depends on the desktop it is serving, so the unit suite is the regression instrument; the end-to-end evidence is a wire capture, in the commit body.


vncdotool/client.py
```diff
@@ -206,6 +207,13 @@ class VNCDoToolClient(rfb.RFBClient):
     SPECIAL_KEYS_US = '~!@#$%^&*()_+{}|:"<>?'
     MAX_DESKTOP_SIZE = 0x10000
 
+    # x11vnc polls the X pointer rather than reading back the event it just
```
sibson, line 210: overly specific, should be in comment, narrative

sibson: this describes a behaviour and is disconnected from the code

sibson: 🤖 Claude Code: Dropped the x11vnc narration and the `cursor_position()` / `x11vnc/src/cursor.c` citation — a signpost to where the fact lives rather than the fact, and `specs/cursor.md` already carries the mechanism in full. b81ff7c.

sibson: 🤖 Claude Code: Rewritten to say what the constant is rather than narrate a server behaviour, so it attaches to the number it sits above:

```
# A server can report a pointer position from before the move it has
# already acted on. How long one is given to catch up before its reports
# are believed again.
POINTER_POS_SETTLE = 1.0
```

b81ff7c.

PR #528, opened 2026-09-14
Give an unpainted framebuffer an unlimited retry budget

sibson:
`test_capture` against TightVNC failed in roughly half the Windows OS-server runs with a flat capture: vncdotool's own never-painted framebuffer, saved black with exit 0.

`commitUpdate` reads "did anything paint" off `rectangles`, and `CursorDecoder` reports a change even where it draws nothing. TightVNC leads a fresh connection with pseudo-rectangles before any content tile, so the cursor rectangle spent the single retry and the refresh completed on an unpainted screen. The budget was the wrong part rather than the guard, so `retry()` now says yes for as long as nothing has reached the framebuffer.

The Windows servers run in CI only, so the evidence is the failing run's artifacts: `tightvnc.png` is 786432 pixels of black, taskbar included, while `ultravnc.png` captured a second later in the same step has 110812 colours.


vncdotool/client.py
```diff
@@ -143,6 +143,7 @@ class _FullScreenReceived:
 
     def __init__(self, width: int, height: int) -> None:
         self.retries = 0
+        self.painted = False
```
sibson, line 146: painted and unpainted read as inverse of one another.  need to differenciate better

vncdotool/client.py
```diff
@@ -157,12 +158,17 @@ def satisfied(cls) -> "_FullScreenReceived":
 
     def retry(self) -> bool:
         """Whether the server is worth asking again, counting this attempt."""
+        if self.untouched:
+            # A Cursor pseudo-rectangle carries a shape and a hotspot, not a
+            # screen region, so an update can arrive having painted nothing.
```
sibson: comment describes a specific case, when others could also be true.  strke

vncdotool/client.py
sibson: 🤖 Claude Code: Renamed the mask to `remaining` — it is the area the server still owes, which is a different thing from the `painted` flag rather than its inverse. `__str__` still reports the count as "unpainted" in its prose. 9c29bbab

vncdotool/client.py
sibson: 🤖 Claude Code: Struck. DesktopSize, PointerPos and an empty update all arrive having painted nothing, so naming Cursor narrowed a condition that covers all of them. The `untouched` property went with it — one use, and the inlined `self.pending and not self.painted` reads without help. 9c29bbab

PR #529, opened 2026-09-14
fix(servers): stop the scene player losing its key selection to xev

sibson:
Fixes the CI flake where 14 x11vnc cases time out at 15s: every case that sends a key to switch scenes, none that stays on the startup scene.

The player asked for `KeyPressMask | ButtonPressMask` on the root window in one request. X grants `ButtonPressMask` to at most one client per window and the entrypoint's `xev -root` asks for it too, so the loser of that startup race got `BadAccess` on the whole request, `KeyPressMask` included. Clicks are unaffected: the player's own window already selects `ButtonPressMask`, and a button event stops there rather than reaching the root.

Verified with `xev` forced to start first in a throwaway container. The rebuilt fleet now has `xev` holding root `ButtonPress`, so the losing order is the normal one and the fix is exercised rather than bypassed.


tests/goldens/scene_player.py
```diff
@@ -37,9 +37,21 @@ def __init__(self, scene_dir: Path) -> None:
...
-        # Without this, key and button events never reach the container's
-        # `xev -root` sink.
-        screen.root.change_attributes(event_mask=X.KeyPressMask | X.ButtonPressMask)
+        # Keys are selected on the root window, not on this one, or they stop
+        # here instead of propagating to the container's `xev -root` sink.
+        # ButtonPressMask is left out: X grants it to at most one client per
+        # window, xev holds it, and naming it would fail the whole request.
+        #
+        # python-xlib prints, rather than raises, an error no handler was
+        # given for, and calls the handler it was given with (error, request).
```
sibson: This is meaningless drivel

tests/servers/x11vnc-entrypoint.sh
```diff
@@ -45,4 +43,9 @@ for _ in $(seq 1 30); do
     sleep 0.5
 done
 
+# Behind that wait: python-xlib raises rather than retrying a refused
```
sibson: meaningless

tests/goldens/scene_player.py
sibsonk: 🤖 Claude Code: Deleted in bd48b30. It said python-xlib prints an error rather than raising it -- design rationale, which CLAUDE.md routes to the commit body, and b23038a already carries it. The clause after it just restated the lambda on the next line.

tests/servers/x11vnc-entrypoint.sh
sibsonk: 🤖 Claude Code: Deleted in bd48b30. Same call -- it explained why the player start sits behind the xdpyinfo gate, which is commit-body material and b23038a carries it. One consequence worth knowing: nothing now stops a later edit moving that line back above the gate, where python-xlib will not retry a refused connection and nothing restarts the player.

tests/goldens/scene_player.py
sibson: 🤖 Claude Code: Deleted in bd48b30. It said python-xlib prints an error rather than raising it -- design rationale, which CLAUDE.md routes to the commit body, and b23038a already carries it. The clause after it just restated the lambda on the next line.

tests/servers/x11vnc-entrypoint.sh
sibson: 🤖 Claude Code: Deleted in bd48b30. Same call -- it explained why the player start sits behind the xdpyinfo gate, which is commit-body material and b23038a carries it. One consequence worth knowing: nothing now stops a later edit moving that line back above the gate, where python-xlib will not retry a refused connection and nothing restarts the player.

tests/goldens/scene_player.py
```diff
@@ -37,9 +37,15 @@ def __init__(self, scene_dir: Path) -> None:
             override_redirect=True,
             event_mask=X.ExposureMask | X.ButtonPressMask,
         )
-        # Without this, key and button events never reach the container's
-        # `xev -root` sink.
-        screen.root.change_attributes(event_mask=X.KeyPressMask | X.ButtonPressMask)
+        # Without this the player never sees a key and the scene never changes.
```
sibson: # register player to receive key presse

sibson: 🤖 Claude Code: Taken in 61b1b43, as `# Register the player to receive key presses.` -- your text read `key presse`, so I took it as "presses", and capitalised to match the other comments in the file. Say the word if you wanted it verbatim.

PR #540, opened 2026-09-18
Let vnclog's outbound leg dial a ws:// or wss:// target

sibson:
Part of #138: `vnclog -s ws://...`/`wss://...` (including `--capture-raw`) now works instead of being refused outright, so a server only reachable over a WebSocket — Proxmox's console being the motivating case — can be captured for a bug report per `docs/capture.md`.

`portforward.ProxyServer`'s outbound half is a bare `reactor.connectTCP(host, port)`; `VNCLoggingServerProxy` now runs its own version that dials through `websocket.connect()` instead when the target is a WebSocket URL — the same tunnel `client.factory_connect()` already uses for `vncdo`. The inbound listener is untouched; a VNC client still points at `vnclog` over plain TCP either way.

Not run: the fleet's Selenoid/KasmVNC `ws://` services would exercise this end to end, but Docker isn't available in this environment, so this lands with unit coverage of the dispatch logic only (mocked reactor and `websocket.connect`, no real socket). Flagging that explicitly rather than claiming more than I verified.

https://claude.ai/code/session_0141SmXgHJ8eyDEqHQB9K9Jp

---
_Generated by [Claude Code](https://claude.ai/code/session_0141SmXgHJ8eyDEqHQB9K9Jp)_


sibson:
Added `TestVNCLOGCaptureWebSocket` in `tests/functional/test_proxy.py`, driving `vnclog --capture-raw` against the fleet's `qemu` service over its `ws://` address and checking the archive holds real RFB bytes and a decoded rectangle — the same shape as the existing `TestVNCLOGCapture`/`TestVNCLOGCaptureVNCAuth`. This is the actual end-to-end proof the unit tests alone don't give.

I still can't run it myself: Docker isn't available in this sandbox, and starting a local `dockerd` was (correctly) blocked as a container-escape pattern when I tried. CI runs the real fleet, so that's what will actually prove this — watching it now.

---

sibson:
CI status: everything is green except **"Test server functional tests"**, which failed. Every other job (lint, type check, unit tests on 3.10–3.13, CodeQL, coverage, the OS-hosted native-server jobs) passed.

I can't yet tell whether the failure is in my new `TestVNCLOGCaptureWebSocket` or elsewhere — the job's log is dominated by the fleet's own interleaved server debug output (tens of thousands of lines from TigerVNC/KasmVNC/libvncserver), and every log window I can fetch through the GitHub API tooling available to me ends up covering that debug spam rather than the actual test summary, no matter the tail size requested. I also don't have Docker in this environment to reproduce locally (confirmed by trying), so I can't get at this any other way from here.

I've asked for help reproducing this against a real Docker fleet elsewhere. Will push a fix or report back with the actual failure once I have it — not closing this out silently.

---

vncdotool/loggingproxy.py
```diff
@@ -494,6 +506,10 @@ class VNCLoggingServerFactory(portforward.ProxyFactory):
     protocol = VNCLoggingServerProxy
     shared = True
 
+    # host doubles as the full ws(s):// URL when address_family is
```
sibson: explainy

vncdotool/websocket.py
```diff
@@ -105,6 +105,13 @@ def loseConnection(self) -> None:
             # the peer's echo of it as an invalid close code 1005.
             self.sendClose(code=self.CLOSE_STATUS_CODE_NORMAL)
 
+        # portforward.ProxyClient registers this object itself as a producer
```
sibson: narrative?

vncdotool/loggingproxy.py
sibson: 🤖 Claude Code: Deleted — restates what _connectOutbound already shows a few lines up (574b6da...4e0ac33).

vncdotool/websocket.py
sibson: 🤖 Claude Code: Trimmed — dropped the how-I-found-it narrative, kept the one fact that isn't visible elsewhere: what portforward.ProxyClient does with a registered producer (574b6da...4e0ac33).

